PR #141, opened 2012-08-30
Added new syntax for comments

binarycleric:
Comments can be inserted into liquid variables using the following syntax (in addition to the old style using the `comment` block).

``` liquid
{* this is a comment *}

{*
  this is a multi-line comment
*}

{* this works too
*}
```

**Update:** I noticed that multi-line comments would probably be expected using the new format as well. Those should be working now.

### DIFF
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -26,8 +26,8 @@ module Liquid
   VariableAttributeSeparator  = '.'
   TagStart                    = /\{\%/
   TagEnd                      = /\%\}/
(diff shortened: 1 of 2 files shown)


SirRawlins:
Know this is a really old topic but would anyone be interested in having this looked at again? For me a short-hand syntax for comments would be really great, as the existing {% comment %} syntax is too verbose to be user reliably, I find myself falling back to HTML comments  but these get rendered out which isn't great.

PR #259, opened 2013-09-13
Add url encode/decode filters

jeromecornet:
@fw42 @Sirupsen 

url encode / decode filters for liquid (useful to add parameters to dynamic urls)

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -38,6 +38,14 @@ module Liquid
 
     alias_method :h, :escape
 
+    def encode(input)
+      CGI.escape(input) rescue input
+    end
+
+    def decode(input)
+      CGI.unescape(input) rescue input
+    end
+
     # Truncate a string down to x characters
     def truncate(input, length = 50, truncate_string = "...")
       if input.nil? then return end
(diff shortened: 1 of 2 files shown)


j0rd:
does url_param_escape escape '?' <- question marks

I'm trying to append some query params to a pinterest share:

```
href="//pinterest.com/pin/create/button/?url={{ url_to_share | append: '?utm_campaign=Social&utm_medium=share&utm_source=pinterest' | url_param_escape }}
```

But I end up with this as the URL

```
href="//pinterest.com/pin/create/button/?url=http://www.harempants.com/products/harem-pants-women-tribal-prints-blue?utm_campaign=Social%26utm_medium=share%26utm_source=pinterest
```

Notice the two question marks in the URL

fw42:
```url_param_escape``` is not part of the Liquid core project, so this is off-topic here. In the future, please post Shopify-specific questions in the Shopify Forums or contact the design support team.

In Shopify, that filter uses the Ruby ```URI.escape()``` method under the hood, so you should check out the documentation for that to see what exactly it's doing.

For your particular use-case, it looks to me as if the following would make more sense:

```
href="//pinterest.com/pin/create/button/?url={{ url_to_share | url_param_escape | append: '&utm_campaign=Social&utm_medium=share&utm_source=pinterest' }}
```

(note that I changed one ```?``` to ```&```)

PR #336, opened 2014-04-08
implementation/tests of the template inheritance mechanism 

did:
hi @fw42 & @tylerball,
Based on the conversation here (https://github.com/Shopify/liquid/pull/264), this is my implementation of template inheritance, similar to what offers Django (https://docs.djangoproject.com/en/dev/topics/templates/). 
Code mainly extracted from the LocomotiveCMS fork of Liquid. 
Feel free to comment and ask me questions.
Thanks!

### DIFF
--- a/lib/liquid/tags/extends.rb
+++ b/lib/liquid/tags/extends.rb
@@ -1,6 +1,6 @@
 module Liquid
 
-  # Extends allows designer to use template inheritance
+  # Extends allows designer to use template inheritance.
   #
   #   {% extends home %}
   #   {% block content }Hello world{% endblock %}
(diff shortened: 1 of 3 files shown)


did:
btw @fw42 & @tylerball, it works now with liquid 3.0.

jeroenvisser101:
@did I've tried to port changes from 7e4eacd to the gem, but it doesn't seem to work. Instead, the tests throw `Liquid::SyntaxError: Liquid syntax error: 'extends' tag was never closed`, I assume it expects a `{% endextends %}` (which shouldn't be required). Any idea how to proceed?

did:
hey @jeroenvisser101, I think they changed a couple of stuff in their API. 
The liquid version we're using for locomotivecms (http://www.locomotivecms.com) can be found here: https://github.com/locomotivecms/liquid. The last commit from the original master branch was from Jan 24, 2015. 
Hope it helps...

jeroenvisser101:
@did thanks! I'll have a look at that for sure. My best outcome would be to get it working with the 'original' version of liquid, and to get it working with the master branch (there's some things that changed there, options doesn't seem to be writable anymore, but I'll have to investigate more)

I'll let you know when I've got something solid. I don't think it'll be able to be merged into locomotivecms's fork, since it might not work with the liquid fork, but let's have a look when/if I get this to work.

devdreamsolution (approved):

PR #376, opened 2014-07-02
Avoid holding on to stale tag classes

davidcornu:
Liquid's `Template.register_tag` keeps a reference to the class used for each tag in a hash. While this is fine for production use, Rails redefines constants when reloading classes which causes any reloaded tag classes to break.

Ping @byroot

### DIFF
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -38,6 +38,9 @@ module Liquid
   PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/om
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/om
   VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
+
+  singleton_class.send(:attr_accessor, :cache_classes)
+  self.cache_classes = true
 end
 
 require "liquid/version"
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -21,6 +21,36 @@ module Liquid
     attr_accessor :root, :resource_limits
     @@file_system = BlankFileSystem.new
 
+    class TagRegistry
+      def initialize
+        @tags  = {}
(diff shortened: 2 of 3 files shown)


jeroenvisser101:
@davidcornu the same happens with Filters, is this something we can do the same for?

davidcornu:
@jeroenvisser101 I haven't done too much digging into how those are loaded. IIRC they're included in anonymous modules and referenced from there.

jeroenvisser101:
@davidcornu does seem like it yeah, I'm not sure how you'd make something that removes the old and then adds the new back

davidcornu:
Did some digging.

Possible causes:

1. Both `@@global_strainer` and `@@strainer_class_cache` in `Liquid::Strainer` hold references to anonymous sub and subsubclasses of `Liquid::Strainer` that `include` every filter module, which means any reloaded modules are completely ignored

  https://github.com/Shopify/liquid/blob/86944fe7b77710b299a5d03fb6ef816f6c09feb4/lib/liquid/strainer.rb#L10-L18
2. `Liquid::Context` holds on to an array of filter modules

  https://github.com/Shopify/liquid/blob/86944fe7b77710b299a5d03fb6ef816f6c09feb4/lib/liquid/context.rb#L43-L55

Thoughts:

- `Liquid::Context` is rebuilt every time the template is rendered so that shouldn't be an issue
- `@@strainer_class_cache`
  - Subclasses `@@global_strainer` which isn't included in the cache key
  - Should be busted when constants are redefined by Rails' reloading (the same constant would point to a new object which would mean a different key in the hash)
  - `Hash` holds a reference to its keys, which means reloaded filter modules are kept in memory (this could be solved with a weak reference)
  - Callers to `Liquid::Strainer.create` would have to make sure they don't hold references
- `@@global_strainer`
  - Will never be reloaded and holds references to every filter module
  - Could rely on something like `TagRegistry` to resolve the constants for filter modules and check against a cache of some sort

@byroot and @jeroenvisser101 - Am I making any sense?

Issue #427, opened 2014-08-22
Can't Concatenate / Join Arrays


parkr:
Hey,

I was just reviewing the Jekyll issues and saw [this comment](https://github.com/jekyll/jekyll/issues/2793#issue-40882124) by @sylvaindurand:

> I couldn't find a way to perform a loop on both elements, by joining site.posts and site.pages (or site.posts and site.documents), or finding another variable. Is this possible ? It would be great and I don't think it would be too specific.

I agree that we should have a way to join arrays. I tried doing this with `+` but no dice. I tried doing this with `| plus: ` but also, no dice.

Do you think this is a worthy feature for the standard filters or for the `+` operand? Seems pretty straight forward and necessary to me.

rohitpaulk:
+1 on this, #429 (improving `append` to handle arrays) looks like a good solution.

thiagodebastos:
+1 to this! I am looking for a way to display all collections in one masonry grid

fw42:
@pushrax I think https://github.com/Shopify/liquid/pull/540 addressed this issue, right? Can we close?

PR #429, opened 2014-08-24
add concat filter

dvbeato:
adding concat filter for arrays. it is very usefully and fix the issue #427

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -177,6 +177,10 @@ module Liquid
       input.to_s + string.to_s
     end
 
+    def concat(input, array)
+      InputIterator.new(input).concat(array)
+    end
+
     # prepend a string to another
     def prepend(input, string)
       string.to_s + input.to_s
@@ -342,6 +346,10 @@ module Liquid
 
       def join(glue)
         to_a.join(glue)
+      end
+
+      def concat(args)
+        to_a.concat Array(args)
       end
 
       def reverse
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -349,7 +349,7 @@ module Liquid
       end
 
       def concat(args)
-        to_a.concat Array(args)
+        to_a.concat args
       end
 
       def reverse
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -358,6 +358,14 @@ class StandardFiltersTest < Minitest::Test
     assert_template_result('bcd',"{{ a | append: b}}",assigns)
   end
 
+  def test_concat
+    assert_equal([1,2,3,4 ], @filters.concat([1,2], [3,4] ))
+    assert_equal([1,2,'a' ], @filters.concat([1,2], ['a'] ))
+    assert_equal([1,2, 10 ], @filters.concat([1,2], [ 10] ))
+    assert_equal(['a', 10 ], @filters.concat('a',  10  ))
+    assert_equal(['a','bc'], @filters.concat('a', 'bc' ))
+  end
+
   def test_prepend
     assigns = {'a' => 'bc', 'b' => 'a' }
     assert_template_result('abc',"{{ a | prepend: 'a'}}",assigns)
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -362,8 +362,6 @@ class StandardFiltersTest < Minitest::Test
     assert_equal([1,2,3,4 ], @filters.concat([1,2], [3,4] ))
     assert_equal([1,2,'a' ], @filters.concat([1,2], ['a'] ))
     assert_equal([1,2, 10 ], @filters.concat([1,2], [ 10] ))
-    assert_equal(['a', 10 ], @filters.concat('a',  10  ))
-    assert_equal(['a','bc'], @filters.concat('a', 'bc' ))
   end
 
   def test_prepend
--- a/example/server/example_servlet.rb
+++ b/example/server/example_servlet.rb
@@ -23,7 +23,7 @@ class Servlet < LiquidServlet
   end
 
   def products
-    { 'products' => products_list, 'description' => description, 'section' => 'Snowboards', 'cool_products' => true}
+    { 'products' => products_list, 'more_products' => more_products_list, 'description' => description, 'section' => 'Snowboards', 'cool_products' => true}
   end
 
   private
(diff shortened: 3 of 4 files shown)


fw42:
@diogobeato: Any success with JRuby? Can you rebase this PR?

diogobeato:
I'm having troubles with git, sorry! I'll fix it

diogobeato:
@fw42 can you revise?

test/integration/standard_filter_test.rb
```diff
@@ -353,9 +353,21 @@ def test_floor
...
+    assert_template_result('bcd'      ,"{{ a | append: 'd'}}" ,assigns)
+    assert_template_result('bcd'      ,"{{ a | append:  b }}" ,assigns)
+    assert_template_result('123410'   ,"{{ c | append:  d }}" ,assigns)
+    assert_template_result('bc1234'   ,"{{ a | append:  c }}" ,assigns)
+    assert_template_result('12341234' ,"{{ c | append:  c }}" ,assigns)
+
+    assert_equal([1,2,3,4],  @filters.append([1,2], [3,4] ))
+    assert_equal([1,2,'a'],  @filters.append([1,2],  'a'  ))
+    assert_equal([1,2, 10],  @filters.append([1,2],  10   ))
+    assert_equal(['a', 1,2], @filters.append('a',  [1,2]  ))
```
fw42: I still have this concern

fw42:
@arthurnn can you take another look here as well please?

test/integration/standard_filter_test.rb
fw42: hmm.. I guess it kinda makes sense. Alternative would be to ```a.to_s + b.to_s``` here or just return nil if the two inputs are of different type. hm.

fw42:
I think I'm ok with this. If @arthurnn or @dylanahsmith are cool with it too, I will merge it.

diogobeato:
@fw42 what do you think about use https://gitter.im/ on liquid repository?

test/integration/standard_filter_test.rb
parkr: It'd be nice to have a use-case. In Jekyll, i'd be happy to have arrays of multiple types. Example is creating an array of Document, Post, Page, or StaticFile instances. If we are restricted to same-type, it could get frustrating. So having `append` always return an array bo matter what would be ideal.

diogobeato: if I make append always return array, it will change completely their behavior, that was to append strings.  maybe be better have a different filter to append arrays, like I made before.

arthurnn: IMO, this should behave like plan-ruby would, so nobody gets unexpected behaviour:
```ruby
irb(main):004:0> '1' << '2'
=> "12"
irb(main):005:0> '1' << ['2']
TypeError: no implicit conversion of Array into String
	from (irb):5
	from /usr/bin/irb:12:in `<main>'
irb(main):006:0> ['1'] << ['2']
=> ["1", ["2"]]
irb(main):007:0> ['1'] << '2'
=> ["1", "2"]
```

Those 4 cases should be the tests in here.

fw42: Hm yeah agree, the fact that both ```append([1,2], 3)``` and ```append([1,2], [3])``` return the same thing seems like too much magic. Lets do what Ruby does. If you actually want a way to turn ```[1,2]``` and ```[3]``` into ```[1,2,3]```, then we should probably just have another filter (```concat```?).

diogobeato: I like concat returning a new array, like plain-ruby does

diogobeato:
change to use concat filter, can you revise pls? @fw42 @parkr @arthurnn

test/integration/standard_filter_test.rb
```diff
@@ -358,6 +358,14 @@ def test_append
     assert_template_result('bcd',"{{ a | append: b}}",assigns)
   end
 
+  def test_concat
+    assert_equal([1,2,3,4 ], @filters.concat([1,2], [3,4] ))
+    assert_equal([1,2,'a' ], @filters.concat([1,2], ['a'] ))
+    assert_equal([1,2, 10 ], @filters.concat([1,2], [ 10] ))
+    assert_equal(['a', 10 ], @filters.concat('a',  10  ))
```
fw42: Hmm. Still not sure if I feel like this is too much magic or not. I can already see people "misusing" this feature to "create" Array objects. Not sure if that's something that we want. Imho ```concat``` should only work if both inputs are an array, since that is what Ruby does. If you want to append an element to an existing array, you should use the ```append``` filter.

lib/liquid/standardfilters.rb
```diff
@@ -177,6 +177,10 @@ def append(input, string)
       input.to_s + string.to_s
     end
 
+    def concat(input, array)
+      InputIterator.new(input).concat(array)
```
fw42: I guess this part is okay

lib/liquid/standardfilters.rb
```diff
@@ -344,6 +348,10 @@ def join(glue)
         to_a.join(glue)
       end
 
+      def concat(args)
+        to_a.concat Array(args)
```
fw42: but this I would prefer to just be ```to_a.concat(args)```. @arthurnn, thoughts?

arthurnn: agree with @fw42

test/integration/standard_filter_test.rb
arthurnn: agree with @fw42

test/integration/standard_filter_test.rb
```diff
@@ -358,6 +358,12 @@ def test_append
     assert_template_result('bcd',"{{ a | append: b}}",assigns)
   end
 
+  def test_concat
+    assert_equal([1,2,3,4 ], @filters.concat([1,2], [3,4] ))
+    assert_equal([1,2,'a' ], @filters.concat([1,2], ['a'] ))
+    assert_equal([1,2, 10 ], @filters.concat([1,2], [ 10] ))
+  end
+
```
diogobeato: @parkr @fw42 what do you think now? will works only with arrays, and rise exceptions with other types.

lib/liquid/standardfilters.rb
fw42: why do we still need InputIterator? Why not just ```input.concat(array)```?

diogobeato: because if use only input.concat(array) the code will be able to concat strings like assert_equal('1,210', @filters.concat('1,2', '10' ))

diogobeato: with InputIterator ensures that works only with arrays

pushrax:
Merged via #540

PR #435, opened 2014-08-28
Add Parser Comparison Test

trishume:
This adds a somewhat useful test that checks if the strict parser and the lax parser produce the same tag tree output. On its own this only useful to make sure the strict and lax parsers do the same thing, but the real fanciness comes from the implications of the fact that this can be done.

I did this as a proof of concept for how one could test that Shopify/liquid-c does the same thing as the existing lax parser. If one was to take this code and run it on a huge dataset of Shopify shop templates then one could be reasonably confident that the two parsers behave the same.

Previous attempts to compare parsers have tried to render templates with both parsers but this is difficult to do since rendering only really works in a production environment. However, this method only needs the template data, and the parsing code for any custom tags.

This method could solve the transition problem of how to be confident that a new parser works.

@fw42 @camilo @tonyzou

### DIFF
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -130,7 +130,7 @@ module Liquid
         @variable_name = $1
         @collection_name = $2
         @name = "#{$1}-#{$2}"
-        @reversed = $3
+        @reversed = $3 || false
         @attributes = {}
         markup.scan(TagAttributes) do |key, value|
           @attributes[key] = value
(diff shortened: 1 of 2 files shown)


fw42:
@trishume do you still want to ship this?

trishume:
I don't think so, as one can see from the failing tests this isn't really portable across ruby versions and all the other methods I remember trying (like YAML) weren't either. I mainly created this PR as a proof of concept and reference for how liquid_c could be tested for compatibility.

There is already a decent amount of tests for compatibility of the strict parser in liquid and I don't think these tests add anything more with the relatively small dataset they are run on, they only really make sense if run on huge sets like Shopify's template database.

PR #441, opened 2014-09-11
Removed context from read_template_file, fixed tests to match new arity

sunblaze:
We want to remove context from read_template_file so we can move parsing of partials to the time of parsing the template. This is to allow for better caching of parsed templates in the future.

New method signature for implementing your own liquid filesytem:
```ruby
def read_template_file(template_path)
```

### DIFF
--- a/History.md
+++ b/History.md
@@ -6,6 +6,7 @@
 * Block parsing moved to BlockBody class (#458) [Dylan Thacker-Smith, dylanahsmith]
 * Add concat filter to concatenate arrays (#429) [Diogo Beato, dvbeato]
 * Ruby 1.9 support dropped (#491) [Justin Li, pushrax]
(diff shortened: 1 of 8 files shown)


fw42:
@sunblaze can you rebase and ship this?

sunblaze:
Rebased. It's been a while so I'm not sure what needs to be added in the History.md for these changes.

pushrax:
Test failure fixed in master.

fw42:
@sunblaze can this be shipped? can you own it?

sunblaze:
Yep, I'll own it!

sunblaze:
Updated History.md and removed a couple more instances of the old method signature being used.

fw42:
:ship: :it:

pushrax:
:+1:

chamnap:
@sunblaze, just wonder a little bit. Previously, I depend on this context variable to query current rails controller name, then I can build a template path variable which prefixed `home` or `dashboard`. Without this context variable, what should I do?

PR #442, opened 2014-09-12
Benchmark marshal load and render against parse and render.

dylanahsmith:
cc @camilo, @sunblaze, @grollest 

I modified the benchmarks to show how much time we could spend by saving a Marshal.dump of the parsed liquid template by comparing a Marshal.load against Template.parse.  I found the results from benchmark/ips harder to compare with each other than just using benchmark from the standard library.

Here are the results:

```
~/src/liquid(benchmark-load-render)$ bundle exec rake benchmark:run
/Users/dylansmith/.rbenv/versions/2.1.2/bin/ruby ./performance/benchmark.rb lax
Rehearsal ----------------------------------------------------------
parse:                   3.770000   0.050000   3.820000 (  3.819407)
marshal load:            1.320000   0.010000   1.330000 (  1.326707)
render:                  4.400000   0.040000   4.440000 (  4.447144)
marshal load & render:   5.780000   0.020000   5.800000 (  5.806654)
parse & render:          8.130000   0.030000   8.160000 (  8.154378)
------------------------------------------------ total: 23.550000sec

parse:                   3.600000   0.020000   3.620000 (  3.619495)
marshal load:            1.320000   0.010000   1.330000 (  1.331599)
render:                  4.260000   0.020000   4.280000 (  4.279742)
marshal load & render:   5.720000   0.020000   5.740000 (  5.748194)
parse & render:          8.320000   0.050000   8.370000 (  8.368795)
```

Note that the Marshal.dump of the parsed template was twice as big as the source in the benchmark themes.

### DIFF
--- a/performance/benchmark.rb
+++ b/performance/benchmark.rb
@@ -1,17 +1,14 @@
-require 'benchmark/ips'
+require 'benchmark'
 require File.dirname(__FILE__) + '/theme_runner'
 
 Liquid::Template.error_mode = ARGV.first.to_sym if ARGV.first
 profiler = ThemeRunner.new
 
-Benchmark.ips do |x|
-  x.time = 60
-  x.warmup = 5
-
-  puts
-  puts "Running benchmark for #{x.time} seconds (with #{x.warmup} seconds warmup)."
-  puts
-
-  x.report("parse:") { profiler.compile }
-  x.report("parse & run:") { profiler.run }
+N = 100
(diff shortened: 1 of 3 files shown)


fw42:
@dylanahsmith status?

pushrax:
I don't think this needs to happen with Liquid C - the parsing step is orders of magnitude faster than rendering now.

PR #452, opened 2014-10-01
replace the regex sequence w with [[:word:]] to be more flexible

craig-day:
Make liquid more flexible in internationalized situations.

Before:
Markup was only rendered properly if the markup contained latin characters. 
i.e.  {{test}}

After:
Markup can contain non-latin characters, for example Japanese. 
i.e. {{テスト}}

@boourns, @fw42, @camilo, @dylanahsmith, or @arthurnn can you code review?

### DIFF
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -26,14 +26,14 @@ module Liquid
   VariableAttributeSeparator  = '.'.freeze
   TagStart                    = /\{\%/
   TagEnd                      = /\%\}/
-  VariableSignature           = /\(?[\w\-\.\[\]]\)?/
-  VariableSegment             = /[\w\-]/
+  VariableSignature           = /\(?[[[:word:]]\-\.\[\]]\)?/
+  VariableSegment             = /[[[:word:]]\-]/
   VariableStart               = /\{\{/
   VariableEnd                 = /\}\}/
   VariableIncompleteEnd       = /\}\}?/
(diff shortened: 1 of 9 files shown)


betelgeuse:
@camilo @boourns @jasonhl How about this being available through a configuration option? My native language has characters outside ASCII and so far we have been using a fork because this is useful.

jasonhl:
@betelgeuse: If this can be done without any performance penalty, I'd support a configuration option.

boourns:
@jasonhl how could it be configurable as more of liquid core moves to C for performance?

jasonhl:
I got the impression from @pushrax that this wasn't insurmountable in Liquid-C, it just might take a bit more work.

Can't we just fallback to Ruby (i.e. bail out of Liquid-C) when this configuration switch is on, like we do when there are template issues?

pushrax:
Yeah, it's doable in C. I would prefer if we didn't have more syntax configuration; if we support this it should just be supported everywhere. Otherwise it will get even more confusing.

camilo:
I don't think the complexity added is justified by the benefits, as long as we can deal properly with utf-8 content I fail to see the need to have utf-8 identifiers, if it came for absolutely free then that'd be fine.

boourns:
:+1: @camilo

betelgeuse:
@camilo In Finnish it's weird for end users when you need to tell them to write {{maara}} instead of the proper {{määrä}}.

PR #473, opened 2014-10-23
Nested square brackets are parsed incorrectly

pushrax:
Another bug in the lax parser found while asserting correctness of Liquid-C.

The cause is that the definition of the `VariableParser` regex will match `a[b[c[d]]]` as `a[b[c[d]`, thinking the closing bracket for the innermost nesting is actually closing the first bracket. This makes the rest of the `VariableLookup` code throw up and produce an incorrect result.

I'm not sure how this could be parsed cleanly with a just a regex. If the input is guaranteed to have no extraneous closing brackets it may be possible to just grab them all, but doing so seems sketchy to me.


pushrax:
I don't think this will ever get fixed in the lax parser unless there is a major rewrite effort.

PR #499, opened 2014-12-16
Fix case sensitivity regression in date standard filter

kreynolds:
The to_date refactor on Apr 30 removed downcase from the obj switch, this patch adds back a case-insensitive regexp to check for 'now' and 'today' as well as a test for it.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -307,7 +307,7 @@ module Liquid
       return obj if obj.respond_to?(:strftime)
 
       case obj
-      when 'now'.freeze, 'today'.freeze
+      when /\A(?:now|today)\z/i
         Time.now
       when /\A\d+\z/, Integer
         Time.at(obj.to_i)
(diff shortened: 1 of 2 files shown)


pushrax:
Could this affect the operation of the later `Time.parse(obj)`? Would it be enough to do:

```ruby
when 'now'.freeze, 'today'.freeze, 'Now'.freeze, 'Today'.freeze
```

?

kreynolds:
It would be enough to fix this specific regression but not particularly robustly (handling other variations, typos), though maybe thats undesirable here? I'm good either way just let me know what to do to get merged.

pushrax:
I'm just worried that the downcasing would potentially break the parsing of certain time formats. That may be totally incorrect though.

IMO it doesn't make sense to support any capitalization – why should `tODAy` be valid? I wouldn't expect that behaviour as a user of the language.

kreynolds:
I looked through date_parse.c and while there are some case sensitive regexes, they are used for speed and not case sensitivity, so it won't break any time formats (and I doubt it ever will, but its not important to me).

A user could also use `TODAY` or `NOW` or make a typo or any number of other things .. I see no real upside for fixing this as narrowly as suggested but like I said, either one fixes _*MY*_ issue so just let me know which way you want to go with it and I'll amend accordingly.

parkr:
What's holding this guy up?

arthurnn:
needs a changelog!

PR #512, opened 2015-01-23
Fix Tobi last name on gemspec

arthurnn:
am I wrong?
Maybe it was wrong in the first place because rubygems would not accept non-ascii chars?

review @tobi


fw42:
ü is (8-bit) ascii

tobi:
there used to be a bug with umlauts in Rubygems when I started this gem. ue is the "official" way to write a ü if you can't actually use an umlaut. Thanks!

PR #519, opened 2015-02-03
Allow filters to redefine Object methods to make them invokable.

### LINKED ISSUE #518: Allowing blacklisted Object methods as filter names
I'm updating to Liquid 3 from an older version in one of my projects, and some of my filters (e.g., `to_json`) don't work. Simply changing the filter names (e.g., by adding a `2` to them, like `to_json2`) makes them work again.

From what I can tell, [this commit](https://github.com/Shopify/liquid/commit/38b4543bf184cb2fcae743528d6a9b39c20bbd32) from two years ago is where these filters would have stopped working, caused by the new method name blacklisting logic.

Because I'm using Liquid 3 in a Rails app, and because Rails' ActiveSupport extends `Object` with several methods (`to_json` among them), these added methods are not allowed to be used as Liquid filter names.

For example, this code works when ActiveSupport's core extensions are not loaded. (Note that for simplicity the `to_json` filter outputs a static JSON string regardless of the input.)

```ruby
require 'liquid'

# to_json is not defined on Object
Object.instance_methods & [:to_json] #=> []

module MyFilters; def to_json(obj); '{"sample":"output"}'; end; end
tmpl = Liquid::Template.parse("{{ obj | to_json }}")
puts tmpl.render({ "obj" => { 'a' => 'b' } }, filters: MyFilters)
#=> {"sample":"output"}
```

Here's the same code, but this time after ActiveSupport's core extensions are loaded.

```ruby
require 'active_support'
require 'active_support/core_ext'
require 'liquid'

# to_json is defined on Object now:
Object.instance_methods & [:to_json] #=> [:to_json]

module MyFilters; def to_json(obj); '{"sample":"output"}'; end; end
tmpl = Liquid::Template.parse("{{ obj | to_json }}")
puts tmpl.render({ "obj" => { 'a' => 'b' } }, filters: MyFilters)
#=> {"a"=>"b"}
```

In this second example, the custom `to_json` filter is not added to the `Liquid::Strainer` due to the `@@method_blacklist`, which is initialized with `Strainer.instance_methods`, which includes `to_json` when ActiveSupport's core extensions are loaded. As such, Liquid simply outputs the `{"a"=>"b"}` hash as a string rather than the `{"sample":"output"}` string we expected.

I was able to work around this with a *very* ugly hack in a Rails initializer:

```ruby
Liquid::Strainer.class_variable_set(:@@method_blacklist,
  Liquid::Strainer.class_variable_get(:@@method_blacklist) -
  CustomLiquidFilters.instance_methods.map(&:to_s))
```

This essentially allows all methods defined by a `CustomLiquidFilters` module to bypass the blacklist. (This does not include `Object` methods; in the context of the example code above, `MyFilters.ancestors #=> [MyFilters]`, and `MyFilters.instance_methods #=> [:to_json]`.)

Is there a better way to whitelist method names for custom filters? I realize I could rename the filter to just `json`, but there are other filters that are being blacklisted by default. Plus—and this is my main reason for opening this issue rather than just relying on a hack—custom filters that work today could be inadvertently blacklisted by future versions of Rails, ActiveSupport, or any other gem that adds methods to `Object`.

This was not easy to track down ("why does this filter work and the one *right next to it* not work?"), and it could be a source of frustration when new methods are added to `Object`, causing some filters to no longer be registered, when nothing about Liquid was changed. For instance, running `bundle update rails` and then having tests for custom filters fail even though those filters are (seemingly) in no way related to the Rails update (just like the example `to_json` filter above that simply returns a static string) would be confusing.

I tried making `Liquid::Strainer` inherit from `BasicObject` (added in Ruby 1.9.1), but this means it loses useful things like `respond_to?`. Before going deeper, I wanted to see if anyone on the Liquid team had any insight or opinion on this, and whether a "fix" would even be considered. Perhaps something as simple as defining `Liquid::Strainer#whitelist_filters(filter_names_or_module)` would do the trick.

dylanahsmith:
@arthurnn & @fw42 for review
cc @nickpearson
Fixes #518

## Problem

Strainer has been using Strainer.instance_methods as a blacklist of methods that couldn't be called.  This means that object like `.to_json` which rails declares on Object can't be used, which is a good default.  However, this blacklist also prevented a filter from having a custom `to_json` that is intended to be exposed.

## Solution

An empty module doesn't have any instance methods predefined on it, unlike a class, so we don't need to filter these methods with a blacklist.  However, I still needed to refactor the Strainer class so that there was a each set of filters had their own whitelist of methods that can be called.

I also simplified the code so that there is only one place where filters are added, which also means that `Liquid::Context#add_filter` now uses the strainer class cache.

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -44,21 +44,8 @@ module Liquid
     # for that
     def add_filters(filters)
       filters = [filters].flatten.compact
-      filters.each do |f|
-        raise ArgumentError, "Expected module but got: #{f.class}" unless f.is_a?(Module)
-        Strainer.add_known_filter(f)
-      end
-
-      # If strainer is already setup then there's no choice but to use a runtime
-      # extend call. If strainer is not yet created, we can utilize strainers
-      # cached class based API, which avoids busting the method cache.
-      if @strainer
-        filters.each do |f|
-          strainer.extend(f)
-        end
-      else
-        @filters.concat filters
-      end
+      @filters += filters
+      @strainer = nil
     end
 
     # are there any not handled interrupts?
--- a/lib/liquid/strainer.rb
+++ b/lib/liquid/strainer.rb
@@ -8,12 +8,13 @@ module Liquid
   # The Strainer only allows method calls defined in filters given to it via Strainer.global_filter,
   # Context#add_filters or Template.register_filter
   class Strainer #:nodoc:
-    @@filters = []
-    @@known_filters = Set.new
-    @@known_methods = Set.new
+    @@global_strainer = Class.new(Strainer) do
+      @filter_methods = Set.new
+    end
     @@strainer_class_cache = Hash.new do |hash, filters|
-      hash[filters] = Class.new(Strainer) do
-        filters.each { |f| include f }
+      hash[filters] = Class.new(@@global_strainer) do
+        @filter_methods = @@global_strainer.filter_methods.dup
+        filters.each { |f| add_filter(f) }
       end
     end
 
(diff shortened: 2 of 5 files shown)


nickpearson:
This looks great. I tested standalone and inside my app (to give it the full Rails treatment), and it fixes the problem in both places. Thanks! I think this behavior makes the most sense—if a filter is explicitly defined, allow it to be called.

:+1:

fw42:
Cool, looks good to me :+1:

dylanahsmith:
Cherry-picked the merge commit to the 3-0-stable branch.

fw42:
sweet. can you release a new version?

fw42:
3.0.2 I guess

PR #520, opened 2015-02-04
Add forloop.parentloop as a reference to the parent loop

### LINKED ISSUE #500: Accessing outer for loop
I can't seem to find a way to access the forloop object for the outermost for loop in a template that looks like this. 

```
{% for outerItem in outerItems %}
    {% for item in items%}
        <div>{{ forloop.counter }}.&nbsp;{{ item }}</div>
    {% endfor %}
{% endfor %}
```
`forloop.counter` returns the innermost for loop's counter in the above example

pushrax:
Fixes #500 in a somewhat nicer way.

### DIFF
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -42,6 +42,7 @@ module Liquid
   #                   where 0 is the last item.
   # forloop.first:: Returns true if the item is the first item.
   # forloop.last:: Returns true if the item is the last item.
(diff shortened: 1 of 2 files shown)


fw42:
LGTM. @dylanahsmith thoughts?

dylanahsmith:
:+1: seems like an elegant solution for the problem.

arthurnn:
This was a bug fix right?
probably good idea to add a Changelog entry for it.

PR #540, opened 2015-03-25
Add array concat filter

pushrax:
Squashes #429 and adds another test. Everyone's previous comments are addressed, so I'm going to go ahead and merge this after CI.

@fw42 @arthurnn @dvbeato FYI


arthurnn:
we need a changelog entry for this!

PR #545, opened 2015-04-02
Explode invokable_methods method on Liquid::Drop

tjoyal:
Exploding the method in order to make it "re-usable".

I was not sure if I wanted it public or private. In my use case I wish to have access to the information in  `invokable_methods` in order to build some tests and maintenance task for docs so I do not care so much if I have to make a `send` call to get it.

On the other hand if I need it from the outside it would make sense to be a public method, but then a public method just for some test and maintenance task feels like it would not be appropriate.

Thoughts?


fw42:
fine with me

pushrax:
:+1: and I am fine with this being public or private

Issue #547, opened 2015-04-17
Liquid::Variable.name is returning a Liquid::VariableLookup in version 3.0.0 and above


willlahr:
This code snippet behaves differently between 2.6.2 and versions 3.0.0 and above:

    puts Liquid::Template.parse("{{test_var}}").root.nodelist[0].name 

In 2.6.2 it prints out the variable name, which is what I was expecting:
    test_var 

In 3.0.0 and 3.0.1 it prints out the following:
    #<Liquid::VariableLookup:0x007f9f0419e830>

Is this a bug?  Or have I misunderstood the API?

pushrax:
Expression parsing (i.e. determining what to do with `test_var`) used to happen in the rendering step, now it happens in the parsing step. There is no bug, just a breaking change in the API.

willlahr:
Thanks for responding.  This change is creating a problem for me because I was using a list of variables to trigger re-rendering of templates,  I've explained it  a bit more here:  
http://stackoverflow.com/questions/29696259/how-to-get-list-of-variables-referenced-in-a-liquid-template
Is there a way to do this with >3.0.0?   VariableLookup doesn't seem to have any methods to get the variable name, even though it is being held in the object.

pushrax:
There isn't a correct way as we hadn't considered that use-case, but I think it would be reasonable to expose the name/lookups/command flags as read-only methods.

@Shopify/liquid what do you think?

fw42:
I'm ok with exposing this information :+1:

willlahr:
What's the timescale for this merge going into a release on RubyGems?  I'm wondering if to pull from github for production or just wait until rubygems.org has been updated.

fw42:
Just released [v3.0.2](https://github.com/Shopify/liquid/releases/tag/v3.0.2) to [Rubygems](https://rubygems.org/gems/liquid/versions/3.0.2) (see https://github.com/Shopify/liquid/pull/553).

PR #551, opened 2015-04-20
Expose name, lookups, and command flags from VariableLookup

pushrax:
Solution for #547

@willlahr will this work for you?

### DIFF
--- a/test/unit/variable_unit_test.rb
+++ b/test/unit/variable_unit_test.rb
@@ -152,11 +152,5 @@ class VariableUnitTest < Minitest::Test
     lookup = VariableLookup.new('a.b.c')
     assert_equal 'a', lookup.name
     assert_equal ['b', 'c'], lookup.lookups
-    assert_equal 0, lookup.command_flags
-
-    lookup = VariableLookup.new('a.first.size')
-    assert_equal 'a', lookup.name
-    assert_equal ['first', 'size'], lookup.lookups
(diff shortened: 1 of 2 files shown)


test/unit/variable_unit_test.rb
```diff
@@ -147,4 +147,16 @@ def test_output_raw_source_of_variable
...
+  def test_variable_lookup_interface
+    lookup = VariableLookup.new('a.b.c')
+    assert_equal 'a', lookup.name
+    assert_equal ['b', 'c'], lookup.lookups
+    assert_equal 0, lookup.command_flags
+
+    lookup = VariableLookup.new('a.first.size')
+    assert_equal 'a', lookup.name
+    assert_equal ['first', 'size'], lookup.lookups
+    assert_equal 0b11, lookup.command_flags
```
fw42: excuse my ignorance, but what is this?

pushrax: Doesn't really need to be in this test, but it's a bitfield to indicate which lookups are to be treated as method calls, i.e. size, first, and last (`VariableLookup::COMMAND_METHODS`). In fact, this ivar probably doesn't need to be exposed at all, it's just used internally in `VariableLookup#evaluate`.

willlahr:
Yes this is exactly what I need, thanks.   I guess some kind of 'what does this template depend on' feature might be worth thinking about but I'm not sure that anyone else has my use case.

PR #554, opened 2015-05-04
Resolves #529. Resolves #404. Added natural sorting filter and tests.

### LINKED ISSUE #404: sort only sorts in binary
jekyll/jekyll#2135

Not sure if there is any easy solution, but this should probably at least be documented.

### LINKED ISSUE #529: Add case-insensitive sort
Currently ["the order of the sorted array is case-sensitive"](http://docs.shopify.com/themes/liquid-documentation/filters/array-filters#sort). It would be useful to also have a case-insensitive sort.

Use case I'm aiming: listing categories or tags on a Jekyll site in the proper order. Example without proper order: http://blog.lanyonm.org/tags.html
To get around this lack, some downcase all tags: http://www.jokecamp.com/tag/ http://www.jokecamp.com/blog/listing-jekyll-posts-by-tag/

arthanzel:


pushrax:
This filter only makes sense for things that have a `downcase`, which means they probably have a `casecmp`, which I would much prefer using (for performance reasons). Having it require string-like input is fine with me, we just need to document it.

arthanzel:
I updated the PR to use `casecmp`, at the expense of being unable to filter non-strings. Tests with numbers were removed.

lib/liquid/standardfilters.rb
```diff
@@ -122,6 +122,20 @@ def sort(input, property = nil)
...
+    # provide optional property with which to sort an array of hashes or drops
+    def sort_natural(input, property = nil)
+      ary = InputIterator.new(input)
+
+      if property.nil?
+        ary.sort {|a,b| a.casecmp b }
+      elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
+        ary.sort {|a,b| a[property].casecmp b[property] }
+      elsif ary.first.respond_to?(property)
+        ary.sort {|a,b| a.send(property).casecmp b.send(property) }
```
fw42: can you use parentheses here (and above)? `a.casecmp(b)`

fw42:
I'm cool with this.. I'm not sure about the same "sort_natural"... But I like having a filter like this in general.

@arthurnn thoughts?

pushrax:
Actually this isn't quite a natural sort as is, it would probably be good to sort embedded numbers in order, like `a2` coming before `a10`. Not sure we need it though.

arthanzel:
I corrected the parens (oops).

The `sort_natural` came from a bug report (which I can't find anymore) where they mentioned that case-insensitive sorting is more "natural" for humans. The filter was formerly called `sort_insensitive` but I found that it has too many syllables. Changing it may be a good idea to future-proof if somebody ever introduces a "true" natural sort, but I can't think of a good name for it. Maybe `sort_strings`?

pushrax:
I am fine with `sort_natural` or `natural_sort` actually – if we decide to go further we can just improve the same filter. It would be a breaking change but really minor and just affecting output for humans.

arthanzel:
All of the changes discussed above are now in the PR. Do you plan on doing a merge soon?

pushrax:
@fw42 are you ok with the name? @arthurnn want to take a look?

fw42:
Arthur is on vacation I think. Maybe @dylanahsmith can take a look.

test/integration/filter_test.rb
```diff
@@ -74,11 +74,35 @@ def test_sort
...
     assert_equal [1,2,3,4], Variable.new("numbers | sort").render(@context)
     assert_equal ['alphabetic', 'as', 'expected'], Variable.new("words | sort").render(@context)
     assert_equal [3], Variable.new("value | sort").render(@context)
     assert_equal ['are', 'flower'], Variable.new("arrays | sort").render(@context)
+    assert_equal ['Expected', 'case', 'sensitive'], Variable.new("case_sensitive | sort").render(@context)
+  end
+
+  def test_sort_natural
+    @context['words'] = ['case', 'Assert', 'Insensitive']
+    # This specific syntax forces hashes to have string keys. Colons won't work.
```
fw42: I dont think we need this comment

test/integration/filter_test.rb
```diff
@@ -136,3 +160,12 @@ def test_local_filter_with_deprecated_syntax
     assert_equal " 1000$ CAD ", Template.parse("{{1000 | money}}").render!(nil, [CanadianMoneyFilter])
   end
 end # FiltersTest
+
+# Simple object that gmay be passed into a filter.
+# Note to test subjects: do not smuggle test objects out of the testing area.
```
fw42: or those two comment lines

arthanzel:
No room for amusing references in the comments?

PR updated with the offending lines removed.

fw42:
fun is strictly forbidden at shopify :rage:

arthanzel:
> fun is strictly forbidden at shopify :rage:

Funny, I stopped by the Ottawa office last week and I got the opposite impression :grin:

pushrax:
@arthanzel flo's sarcasm is next level ;p

arthanzel:
Well, I'm glad that at least you appreciate the reference

pushrax:
@dylanahsmith ping

dylanahsmith:
:+1:

pushrax:
@fw42 I can't really think of a better name than `sort_natural`, any ideas?

fw42:
meh me neither, :ship: :it:

Issue #461, opened 2014-10-17
Liquid variable with include tag and using for/with not working correctly


nulian:
If you do {% include liquid_var_name %} it loads the template correctly.

But if you do {% include liquid_var_name for arrays %}  it will safe the arrays in the wrong variable name instead of the template it will save it in the iquid_var_nam variable.

PR #556, opened 2015-05-11
make sure to get the correct context_variable_name

nulian:
If the include variable name is a variable get the true template name as context variable name instead of the variable with missing first and last letter.

See issue #461 for more information about this fix.

### DIFF
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -26,7 +26,6 @@ module Liquid
         variable_name = $3
 
         @variable_name = Expression.parse(variable_name || template_name[1..-2])
-        @context_variable_name = template_name[1..-2].split('/'.freeze).last
         @template_name = Expression.parse(template_name)
         @attributes    = {}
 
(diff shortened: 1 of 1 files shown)


fw42:
tests are failing

lib/liquid/tags/include.rb
```diff
@@ -45,7 +44,13 @@ def parse(tokens)
     def render(context)
       partial = load_cached_partial(context)
       variable = context.evaluate(@variable_name)
-
+      
+      if @template_name =~ QuotedString
+          context_variable_name = @template_name[1..-2].split('/'.freeze).last
+      else
+          context_variable_name = context[@template_name]
+      end
```
fw42: This should stay in `initialize`, otherwise you will repeat this parsing work on every `render` call (although it's result will never change).

fw42:
Can you please add a new test that captures this problem? I'm not sure I understand the bug that is being fixed here.

lib/liquid/tags/include.rb
```diff
@@ -45,7 +44,13 @@ def parse(tokens)
     def render(context)
       partial = load_cached_partial(context)
       variable = context.evaluate(@variable_name)
-
+      
+      if @template_name =~ QuotedString
+          context_variable_name = @template_name[1..-2].split('/'.freeze).last
+      else
+          context_variable_name = context[@template_name]
```
pushrax: The issue with this is that it doesn't allow the template name to come from a variable.

pushrax:
Fixed in #559

PR #558, opened 2015-05-11
Allow for a block to be passed on liquid_methods that can transform…

rovermicrover:
…the allowed methods return value. Useful for sanitization and dynamic content.

Added comment explains it best.

You may also pass a block that can be used to transform the method's returned value.

```ruby
class SomeClass
  liquid_methods :an_allowed_method do |object, method, value|
    if value.is_a?(String)
      value += ' that has been transformed'
    else
      value
    end
  end

  def an_allowed_method
    'this comes from an allowed method'
  end

end
```
usage:
```ruby
@something = SomeClass.new
```
template:
```ruby
{{something.an_allowed_method}}
```
output:
```ruby
'this comes from an allowed method that has been transformed'
```
@boourns, @fw42, @camilo, @dylanahsmith, or @arthurnn

### DIFF
--- a/lib/liquid/module_ex.rb
+++ b/lib/liquid/module_ex.rb
@@ -41,22 +41,52 @@
 # You can also chain associations, by adding the liquid_method call in the
 # association models.
 #
+# You may also pass a block that can be used to transform the method's returned value.
+#
+#   class SomeClass
+#     liquid_methods :an_allowed_method do |object, method, value|
+#       if value.is_a?(String)
+#         value += ' that has been transformed'
+#       else
+#         value
+#       end
+#     end
+#
+#     def an_allowed_method
+#       'this comes from an allowed method'
+#     end
+#
+#   end
+#
+# usage:
+#   @something = SomeClass.new
+#
+# template:
+#   {{something.an_allowed_method}}
+#
+# output:
+#   'this comes from an allowed method that has been transformed'
 class Module
-  def liquid_methods(*allowed_methods)
+  def liquid_methods(*allowed_methods, &block)
     drop_class = eval "class #{self.to_s}::LiquidDropClass < Liquid::Drop; self; end"
-    
+
     define_method :to_liquid do
       drop_class.new(self)
     end
-    
+
     drop_class.class_eval do
       def initialize(object)
         @object = object
       end
-      
+
       allowed_methods.each do |sym|
         define_method sym do
-          @object.send sym
+          _value = @object.send(sym)
+          if block
+            block.call(@object, sym, _value)
+          else
+            _value
+          end
         end
       end
     end
--- a/test/unit/module_ex_unit_test.rb
+++ b/test/unit/module_ex_unit_test.rb
@@ -36,6 +36,39 @@ class TestClassC::LiquidDropClass
   end
 end
 
+class TestClassD
+  liquid_methods :allowedD, :chainedE do |object, method, value|
+    if value.is_a?(String)
+      value += "andtransformedD"
+    else
+      value
+    end
+  end
+
+  def allowedD
+    'allowedD'
+  end
+  def restrictedD
+    'restrictedD'
+  end
+  def chainedE
+    TestClassE.new
+  end
+end
+
+class TestClassE
+  liquid_methods :allowedE do |object, method, value|
+    if value.is_a?(String)
+      value += "andtransformedE"
+    else
+      value
+    end
+  end
+  def allowedE
+    'allowedE'
+  end
+end
+
 class ModuleExUnitTest < Minitest::Test
   include Liquid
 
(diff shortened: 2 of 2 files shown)


pushrax:
To me it feels like this doesn't need to be Liquid's concern - it could be implemented in the caller just as easily. Are there any generally useful concrete cases where this would come in handy?

rovermicrover:
For legal reasons we have to keep raw untrusted user input, but we need to sanitize it, by removing all possible html tags, before being used in the liquid templates in order to avoid XSS issues. We also don't want to modify the built in rails attribute read method, in order to avoid causing weird compatibility issues down the line. This leaves us with three solutions, 1 "safe" columns,  2 "safe" methods, or 3 an intermediary object between the model and the liquid drop.

---------

1 - We could have a "safe" column for each liquid permissible attribute, where we stored via callback the sanitized input. This though would create a lot of columns which could cause issues with keeping up the code later down the line if a major refactor was required. We could also try some type of JSON or hstore column, but once again this would make things a lot more difficult down the line. All values are also only used in a liquid template twice, meaning we wouldn't gain very much of a performance increase from saving the sanitized input versus doing it on the fly.

---------

2 - We could also, without a callback, create "safe" methods for each attribute, but this and the above would force us to have to change the methods currently being called in the liquid templates. This could also lead to end user confusion, {{actor.safe_first_name}} is a lot less clear than {{actor.first_name}}.

---------

3 - We could also add in an intermediary object between the model and the liquid drop by overriding to_liquid, but that seems like a lot more complexity than should be needed, and once again introduces compatibility issues and increases the brittleness of the code.

---------

I understand that this could be poorly used, or misused. For example someone putting logic in the proc that would be better suited for a callback, or just generally expensive logic.

But as our case shows sometimes an attribute as it stands just needs to behavior differently in a liquid template than it would normally, and this seems to be the simplest and least obtrusive way to do so.

nickpearson:
What about using procs in the `to_liquid` hash?

```ruby
def to_liquid
  %w( name title ).inject({}) do |hash, col|
    hash[col] = proc{ |context| Sanitizer.sanitize(col) }
    hash
  end
end
```

Each proc will only be called if it is referenced in the Liquid template. Here's a quick example using `upcase` instead of sanitization:

```ruby
class Person
  def name;  puts 'called name';  'Jon'; end
  def title; puts 'called title'; 'Dev'; end
  def to_liquid
    %w( name title ).inject({}) do |hash, col|
      hash[col] = proc{ |context| send(col).upcase }
      hash
    end
  end
end

Liquid::Template.parse('Hi, {{ person.name }}').render('person' => Person.new)
# => called name
# => "Hi, John"
```

Hope this helps.

rovermicrover:
Makes sense but wouldn't that result in to_liquid returning a hash of procs instead of LiquidDrop object? It also has the same effect as using an intermediary object like I mentioned before I believe.

nickpearson:
Yes, there will be procs in the hash returned by `to_liquid`, but the values don't all have to be procs. I generally use procs for operations that are expensive and may not be called on every render, but they seem suited to your use case since several properties can share the same sanitization behavior on-demand.

And yes, you would use this in place of a `Drop`. Defining `to_liquid` is similar to defining `liquid_methods`, but you have a lot more control over what is actually handed to the Liquid template during rendering.

boourns:
I don't think this should live in liquid core, sorry.

fw42:
Agree. I actually feel like we should delete "lib/liquid/module_ex.rb" since it's kinda hacky and I don't really want to encourage it's use.

nickpearson:
Agreed. I've never used `liquid_methods` because I want everything exposed to Liquid templates to be clearly defined in a single place (in `to_liquid`). I imagine `liquid_methods` makes it easy to accidentally expose something sensitive to Liquid since the declaration can be far away (in terms of line count) from the method definition.

rovermicrover:
I thought liquid_methods was the standard way to handle this, but if it's on its own way out then I will close this request. So if this is getting dropped then the solution moving forward is to explicitly implement your own class that inherits from Liquid::Drop for each class and your own to_liquid method like this? I assume this is standard from the current "liquid_methods".

```ruby
class Cat

  attr_reader :name

  def to_liquid
    Cat::Drop.new(self)
  end

end

class Cat::LiquidDropClass < Liquid::Drop

  attr_reader :object

  def initialize(object)
    @object = object
  end

  # Delegations to parent object's methods 
  # that would have before passed to liquid_methods
  def name
    self.object.name.whatever.methods.to.transform.you.want
  end

end
```

nickpearson:
If you use `to_liquid` and don't care to encapsulate/separate other Liquid-specific logic, then you don't need a `Drop` at all:

```ruby
class Cat

  attr_reader :name

  def to_liquid
    { 'name' => name.whatever.methods.to.transform.you.want }
  end

end
```

PR #559, opened 2015-05-11
Lazily evaluate template name for context variable injection

### LINKED ISSUE #461: Liquid variable with include tag and using for/with not working correctly

pushrax:
Fixes #461, as an alternative to #556. The added tests explain what this allows.

### DIFF
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -45,8 +45,13 @@ module Liquid
       partial = load_cached_partial(context)
 
       template_name = context.evaluate(@template_name_expr)
-      variable = @variable_name_expr ? context.evaluate(@variable_name_expr) : context[template_name]
       context_variable_name = template_name.split('/'.freeze).last
+
+      variable = if @variable_name_expr
+        context.evaluate(@variable_name_expr)
+      else
+        context.find_variable(template_name)
+      end
 
       context.stack do
         @attributes.each do |key, value|
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -45,7 +45,7 @@ module Liquid
       partial = load_cached_partial(context)
 
       template_name = context.evaluate(@template_name_expr)
-      variable = context.evaluate(@variable_name_expr || Expression.parse(template_name))
+      variable = @variable_name_expr ? context.evaluate(@variable_name_expr) : context[template_name]
       context_variable_name = template_name.split('/'.freeze).last
 
       context.stack do
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -25,10 +25,9 @@ module Liquid
         template_name = $1
         variable_name = $3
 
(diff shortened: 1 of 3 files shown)


lib/liquid/tags/include.rb
```diff
@@ -44,7 +43,10 @@ def parse(tokens)
 
     def render(context)
       partial = load_cached_partial(context)
-      variable = context.evaluate(@variable_name)
+
+      template_name = context.evaluate(@template_name_expr)
+      variable = context.evaluate(@variable_name_expr || Expression.parse(template_name))
```
fw42: Hm. As I said on the other PR, do we really have to do this here? Calling `parse` in the `render` step feels kinda wrong.

pushrax: Unfortunately yes, since `template_name` comes from the context :(

pushrax: Wait, I was wrong, it can be improved. Pushed a fix.

lib/liquid/tags/include.rb
```diff
@@ -44,7 +43,10 @@ def parse(tokens)
 
     def render(context)
       partial = load_cached_partial(context)
-      variable = context.evaluate(@variable_name)
+
+      template_name = context.evaluate(@template_name_expr)
+      variable = @variable_name_expr ? context.evaluate(@variable_name_expr) : context[template_name]
```
fw42: well that's kinda cheating ;-) `Context[]` just calls `Expression#parse` now :-)

pushrax: Right... the only solution then is to create a VariableLookup manually or do what I had before..

pushrax:
@fw42 how about this? Means that you can't do `{{ include "foo.bar" }}` and have it actually look up `foo.bar` but that should be fine?

fw42:
@dylanahsmith mind taking a look as well?

lib/liquid/tags/include.rb
```diff
@@ -44,7 +43,15 @@ def parse(tokens)
 
     def render(context)
       partial = load_cached_partial(context)
-      variable = context.evaluate(@variable_name)
+
+      template_name = context.evaluate(@template_name_expr)
+      context_variable_name = template_name.split('/'.freeze).last
```
fw42: where did the `[1..-2]` go?

fw42:
Code looks good to me I guess

lib/liquid/tags/include.rb
pushrax: `Expression.parse` takes care of turning quoted strings into strings (if it was unquoted it'd return a `VariableLookup` and `context.evaluate` would look up the variable on the previous line).

dylanahsmith:
LGTM

Issue #560, opened 2015-05-12
Unified syntax for tag arguments [RFC]


pushrax:
Tags are currently implemented as an arbitrary syntax defined in their `#initialize` or `#parse` method. Some of them use a combination of Liquid's regexes to do parsing, but not in a consistent way. This makes updating the language difficult – in particular, phasing out the lax parser.

Filters, on the other hand, are simply passed an array (positional arguments) and hash (keyword arguments), and have no say in how they were parsed. This decoupling is a huge advantage, allowing fixes and improvements to affect all filters without breaking them.

So, the idea is automatically parse tags in a certain format, for example:

```
arg = id colon expr | expr
tag = id (arg (comma arg)*)?
```

Then, the tag can access the resulting arguments (i.e. values returned from `Expression.parse`).

Since we do still want certain tags to have custom syntax, we can't 100% separate parsing from the tag definition. Some kind of way to extend the default will be necessary. However, since most tags don't need this (maybe none outside of the core tags?), it still allows most of the parsing logic to be decoupled.

Tags that need partially custom syntax could still take advantage of the argument parsing, e.g. the for tag:

```
for foo in bar reversed, limit: 4, offset: 5
```

Does this make sense? @Shopify/liquid

nickpearson:
:+1: This will make implementing and maintaining custom tags a lot easier, and will make it easier for new template developers. It would also make it easier for tag developers to keep things consistent and not have to deal with parsing arguments, leaving that to Liquid.

Regarding the need for core tags to have custom parsing, like the `reversed` keyword in the `for` loop you showed, what if HTML-style boolean/no-value keywords were supported? (Like `nowrap` in `<b nowrap class="x">`.)

Valueless keywords could be set to `true`. So in your example, the resulting arguments hash of the `for` tag would be:

```ruby
{ "reversed" => true, "limit" => 4, "offset" => 5 }
```

I'm going off of memory here for the core tags, so sorry if this idea doesn't apply to some others.

pushrax:
I think having the positional arguments be treated as expressions would be the more common case (like in filters). Then many tags like `{% section 'foo' %}` are trivial to implement, and also automatically work with variables or more complicated expressions.

fw42:
seems reasonable to me

parkr:
[...]

PR #561, opened 2015-05-12
Tag formatter/syntax fixer (WIP)

pushrax:
#### Problem

People have templates with broken syntax that is accepted by the lax parser but really makes no sense. We can't switch entirely to the strict parser because of this.

#### Potential Solution

Add a method to convert a Liquid pseudo-AST back into Liquid source. Then, to fix broken templates, this can be run on the lax-parsed AST and written back to the template source.

#### Notes

This has the side effect of reformatting everything, which is arguably good or bad.

Tags that don't implement `format` will use the default, which is to just spit the markup back out. This is one of the motivations for #560 (unified syntax for tag arguments).

I'm not sure how to format `for` and `tablerow`, they accept arguments with commas and without. This is another reason for #560, there is no standard for how we treat arguments.

@fw42 @dylanahsmith @trishume thoughts?

### DIFF
--- a/lib/liquid/expression.rb
+++ b/lib/liquid/expression.rb
@@ -8,6 +8,8 @@ module Liquid
       'empty'.freeze => :empty?
     }
 
+    INVERTED_LITERALS = LITERALS.invert
+
     def self.parse(markup)
       if LITERALS.key?(markup)
         LITERALS[markup]
@@ -29,5 +31,18 @@ module Liquid
       end
     end
 
+    def self.format(value)
+      if INVERTED_LITERALS.key?(value)
+        INVERTED_LITERALS[value].dup
+      elsif value.is_a?(VariableLookup) || value.is_a?(RangeLookup)
+        value.format
+      elsif value.is_a?(String)
+        "\"#{value}\""
+      elsif value.is_a?(Range)
+        "(#{value.to_s})"
+      else
+        value.to_s
+      end
+    end
   end
 end
--- a/lib/liquid/block.rb
+++ b/lib/liquid/block.rb
@@ -57,6 +57,10 @@ module Liquid
       @block_delimiter ||= "end#{block_name}"
     end
 
+    def format
+      "{% #{block_name} %}#{@body.format}{% #{block_delimiter} %}"
+    end
+
     protected
 
     def parse_body(body, tokens)
--- a/lib/liquid/block.rb
+++ b/lib/liquid/block.rb
@@ -58,7 +58,7 @@ module Liquid
     end
 
     def format
-      "{% #{block_name} %}#{@body.format}{% #{block_delimiter} %}"
+      "{% #{raw} %}#{@body.format}{% #{block_delimiter} %}"
     end
 
     protected
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -99,6 +99,18 @@ module Liquid
       output.join
     end
 
+    def format
+      @nodelist.map do |node|
+        if node.is_a?(Variable)
+          "{{ #{node.format} }}"
+        elsif node.is_a?(Tag)
+          node.format
+        else
+          node
+        end
+      end.join
+    end
+
     private
 
     def render_token(token, context)
--- a/lib/liquid/condition.rb
+++ b/lib/liquid/condition.rb
@@ -33,6 +33,12 @@ module Liquid
       @right = right
       @child_relation  = nil
       @child_condition = nil
+
+      if right && right.is_a?(VariableLookup) && right.name.nil?
+        # Strip blank variable lookups resulting from cases like:
+        # {% if foo && bar %}
+        @right = nil
+      end
     end
 
     def evaluate(context = Context.new)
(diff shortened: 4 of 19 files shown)


trishume:
:+1: Awesome work!

I've thought this was a good idea for a while so obviously no objections from me:
https://vault.shopify.com/Liquid-Parsing-Quirks#_examples-of-parsing-weirdness_the-problem-why-this-is-bad_past-efforts_issues-with-fixing-the-problem_ideas-for-fixing-the-problem

Only thing is we have to be *super* careful about expressions because they don't have a strict parser and they are parsed at runtime rather than parse time so they don't have a pseudo-ast to format properly. The solution you have currently might work but I bet there are a lot of edge cases that can slip through. Example that might slip through this (haven't checked or looked at the code in a while) is `something[[0]]`.

pushrax:
@trishume actually expressions are parsed at parse-time since https://github.com/Shopify/liquid/pull/391 and https://github.com/Shopify/liquid/pull/443 :D

trishume:
Great! I've been out of the liquid loop for a while, good to know that problem has been fixed.

Speaking of me being out of the loop you should update the Liquid Parsing Quirks vault page with new info on the work and stats you got while doing the liquid-c work:
https://vault.shopify.com/Liquid-Parsing-Quirks

I know that you did some parse tree comparisons on large samples of templates a while ago. The information I put on there about liquid-c is from back when we were attempting to make it 100% compatible with the lax parser :neutral_face:

fw42:
Hm I kinda feel like it would be cleaner if all those `format` methods would not be on the tags themselves. They are cluttering the code a bit, imho.

fw42:
I think I would prefer this to be implemented in a more "unintrusive" way, as a separate class for example.

pushrax:
@fw42 I agree about it being separate. However, since parsing is implemented in the tags, and formatting is tightly coupled to the parse output, I think it is correct given the current architecture. Otherwise we need to either expose a bunch of ivars or use `instance_variable_get`. Ideally we would have a real AST and the parser would be separated completely, but :poop:

> you should update the Liquid Parsing Quirks vault page

Definitely, good idea!

lib/liquid/expression.rb
```diff
@@ -8,6 +8,8 @@ class Expression
       'empty'.freeze => :empty?
     }
 
+    INVERTED_LITERALS = LITERALS.invert
```
dylanahsmith: How do we want `nil` inverted, since there are 4 literals that map to it: `nil => nil, 'nil'.freeze => nil, 'null'.freeze => nil, ''.freeze => nil`

dylanahsmith:
```
irb> Liquid::Expression::LITERALS.invert[nil]
=> ""
```

lib/liquid/expression.rb
```diff
@@ -29,5 +31,18 @@ def self.parse(markup)
       end
     end
 
+    def self.format(value)
+      if INVERTED_LITERALS.key?(value)
+        INVERTED_LITERALS[value].dup
```
dylanahsmith: Why do you need to `.dup` the frozen string?

lib/liquid/expression.rb
```diff
@@ -29,5 +31,18 @@ def self.parse(markup)
...
+      elsif value.is_a?(VariableLookup) || value.is_a?(RangeLookup)
+        value.format
+      elsif value.is_a?(String)
+        "\"#{value}\""
```
dylanahsmith: This should probably be formatted with single quotes if the string contains double quotes.

dylanahsmith:
So essentially this would be like gofmt auto formatting their liquid to have standard spacing in tags.  It is an opinionated approach, but I think it would be fine.

> Any reason why this is a Bad Idea?

The original intent of the code could be lost by fixing up the syntax to match what it actually does, although it might also make bugs more apparent, by making it more clear what it will actually do, so I think that would be fine.

We do need to be very careful about bugs, since this wouldn't be the type of thing that could easily be rolled back.

trishume:
@dylanahsmith this is about more than standard spacing like `gofmt` because of the fact that liquid's parser uses `scan` which ignores things that don't match. So when a tag like `{% if true && false %}` is parsed everything after the `&&` will be ignored and not be put in the pseudo-ast. Then when it is formatted the output will be something like `{% if true %}` which does the exact same thing except doesn't include the invalid syntax.

dylanahsmith:
ya, I understand. That is what I meant by losing the original intent of the code.

lib/liquid/expression.rb
pushrax: Having `""` as the inverse of `nil` makes sense to me - would be nice to preserve what they used originally though. Will think about that a bit.

Fortunately hash enumeration order is defined as the insertion order in 1.9+, so the inverse mapping is deterministic at least.

lib/liquid/expression.rb
pushrax: In my head the contract of `format` was to return a modifiable string, so that other places can `<<` to the result without worrying about modifying internal state.

lib/liquid/expression.rb
pushrax: Agree. Similarly to `nil`, it'd be nice to preserve what they used in this case though.

devdreamsolution (approved):

PR #562, opened 2015-05-12
Use custom stack for forloop references

pushrax:
There is some strange code in Shopify that handles references to undefined drop methods, which causes an exception to be thrown on `context['forloop']`.

This method is faster and more accurate for finding the parent for loop (prevents overwriting, for example).

### DIFF
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -99,7 +99,7 @@ module Liquid
       # Store our progress through the collection for the continue flag
       context.registers[:for][@name] = from + segment.length
 
-      parent_loop = context.find_variable('forloop'.freeze)
+      parent_loop = context.find_own_variable('forloop'.freeze)
 
       context.stack do
         segment.each_with_index do |item, index|
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -69,7 +69,8 @@ module Liquid
     end
 
     def render(context)
-      context.registers[:for] ||= Hash.new(0)
+      for_offsets = context.registers[:for] ||= Hash.new(0)
+      for_stack = context.registers[:for_stack] ||= []
 
       collection = context.evaluate(@collection_name)
       collection = collection.to_a if collection.is_a?(Range)
@@ -78,7 +79,7 @@ module Liquid
       return render_else(context) unless iterable?(collection)
 
       from = if @from == :continue
-        context.registers[:for][@name].to_i
+        for_offsets[@name].to_i
       else
         context.evaluate(@from).to_i
       end
@@ -97,14 +98,15 @@ module Liquid
       length = segment.length
 
       # Store our progress through the collection for the continue flag
-      context.registers[:for][@name] = from + segment.length
+      for_offsets[@name] = from + segment.length
 
-      parent_loop = context.find_own_variable('forloop'.freeze)
+      parent_loop = for_stack.last
+      for_stack.push(nil)
 
       context.stack do
         segment.each_with_index do |item, index|
           context[@variable_name] = item
-          context['forloop'.freeze] = {
+          loop_vars = {
             'name'.freeze       => @name,
             'length'.freeze     => length,
             'index'.freeze      => index + 1,
@@ -116,6 +118,9 @@ module Liquid
             'parentloop'.freeze => parent_loop
           }
 
+          context['forloop'.freeze] = loop_vars
+          for_stack[-1] = loop_vars
+
           result << @for_block.render(context)
 
           # Handle any interrupts if they exist.
@@ -126,6 +131,8 @@ module Liquid
           end
         end
       end
+
+      for_stack.pop
       result
     end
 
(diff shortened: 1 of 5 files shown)


parkr:
:clap:

fw42:
> This method is faster

proof?

fw42:
The code looks a bit messy to me.

Can you elaborate why this is better and what problem is solves (and how)?

pushrax:
> proof?

Well, it's clearly faster to read off the tail of an array than to do a variable lookup which involves traversing many hashes and calling a bunch of methods. The tradeoff is creating O(1) arrays per Liquid render, where the contents are pointers to existing values.

> Can you elaborate why this is better and what problem it solves (and how)?

The solves the problem of drops intercepting the `forloop` lookup and doing strange things. Since this separate stack is private to the `For` class, other things can't tamper with it.

> The code looks a bit messy to me.

Any suggestions?

fw42:
> The solves the problem of drops intercepting the forloop lookup and doing strange things

can you explain what you mean by that? or better, add a regression test?

pushrax:
Regression test that I don't really want to include in Liquid:

```ruby
  class Foo < Liquid::Drop
    def before_method(method)
      raise Liquid::StandardError, "method missing"
    end

    def bar
      'bar'
    end
  end

  def test_foo
    # ok:
    assert_equal 'bar', Liquid::Template.parse('{{ bar }}').render!(Foo.new)

    # ok:
    assert_raises(StandardError) { Liquid::Template.parse('{{ baz }}').render!(Foo.new) }

    # throws:
    assert_equal 'barbar', Liquid::Template.parse('{% for i in (1..2) %}{{ bar }}{% endfor %}').render!(Foo.new)
  end
```

pushrax:
@dylanahsmith could you give this a review as well?

lib/liquid/tags/for.rb
```diff
@@ -97,14 +98,15 @@ def render(context)
       length = segment.length
 
       # Store our progress through the collection for the continue flag
-      context.registers[:for][@name] = from + segment.length
+      for_offsets[@name] = from + segment.length
 
-      parent_loop = context['forloop'.freeze]
```
dylanahsmith: so the problem was that this could raise if context was a drop with a before_method defined?

lib/liquid/tags/for.rb
```diff
@@ -126,6 +131,8 @@ def render(context)
           end
         end
       end
+
+      for_stack.pop
```
dylanahsmith: do we need an `ensure` to make sure this isn't missed?

lib/liquid/tags/for.rb
pushrax: Yep, essentially. There might be other weird things you could do with `before_method` as well.

lib/liquid/tags/for.rb
pushrax: Good point :+1:

test/integration/error_handling_test.rb
```diff
@@ -1,24 +1,5 @@
 require 'test_helper'
 
-class ErrorDrop < Liquid::Drop
```
fw42: was this not used?

fw42: oh sorry, you moved it because you want to use it somewhere else too. got it.

pushrax: I moved it to `test_helper.rb` since I needed it in another suite.

fw42:
> the issue is that sometimes the context has a drop as an environment

Sorry if I'm being slow here, but I don't understand what this means and why/when/where that would happen. Can you explain?

lib/liquid/tags/for.rb
```diff
@@ -97,14 +98,15 @@ def render(context)
       length = segment.length
 
       # Store our progress through the collection for the continue flag
-      context.registers[:for][@name] = from + segment.length
+      for_offsets[@name] = from + segment.length
 
-      parent_loop = context['forloop'.freeze]
+      parent_loop = for_stack.last
+      for_stack.push(nil)
```
fw42: Why are you pushing nil here? why not push loop_vars below?

fw42: ah because you only want to push it once, not once per iteration

fw42:
Ok I think I get it now... :+1:

dylanahsmith:
:+1:

pushrax:
@fw42 we allow [passing in a drop](https://github.com/Shopify/liquid/blob/1d3c0b3dab6a3c568cc88eb0439e6997a846d09b/lib/liquid/template.rb#L175-L177) as a sort of "root context" called an [environment](https://github.com/Shopify/liquid/blob/1d3c0b3dab6a3c568cc88eb0439e6997a846d09b/lib/liquid/context.rb#L18-L19). `find_variable` [looks through the context's environments](https://github.com/Shopify/liquid/blob/1d3c0b3dab6a3c568cc88eb0439e6997a846d09b/lib/liquid/context.rb#L162-L163) if no scopes contained the input key.

PR #563, opened 2015-05-13
Prefer Class.new() where possible

fw42:
If [Eric Wong](https://github.com/defunkt/unicorn/commit/d8109e5e531ceca5a4d1c0a4f9e4d625f74e336b) does it, it must be good!

@byroot

### DIFF
--- a/lib/liquid/errors.rb
+++ b/lib/liquid/errors.rb
@@ -49,12 +49,12 @@ module Liquid
     end
   end
 
-  class ArgumentError < Error; end
-  class ContextError < Error; end
-  class FileSystemError < Error; end
-  class StandardError < Error; end
-  class SyntaxError < Error; end
-  class StackLevelError < Error; end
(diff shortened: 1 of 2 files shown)


byroot:
Lol that's foolproof argument I guess.

Code LGTM and I prefer that style anyway. But I would love to hear the explanation behind the 200bytes saving.

Sirupsen:
:fr: @byroot

camilo:
meh LGTM as it was before, if the savings are real :+1:  if not meh

pushrax:
the savings don't matter at all really.. but :+1: for style

camilo:
why is this better style? IMO is just diff

byroot:
Because `; end` is super ugly.

camilo:
I think is pretty... lol

parkr:
:clap:

PR #564, opened 2015-05-13
Rubocop

fw42:
Adds a basic (more or less arbitrary, not too strict, but somewhat reasonable) Rubocop config (that I took from one of my personal projects). This enforces consistent Ruby style (that I believe is mostly identical to the Shopify Ruby style guide).

There is still some stuff in the todo file that I was too lazy to fix, but this PR already cleans up a lot (for the better, imho).

Let the bike-shedding begin! :-) I don't really care too much about the specific rules (with a few exceptions), I mostly care that **some** rule exists.

@pushrax @byroot thoughts on this?


example/server/liquid_servlet.rb
```diff
@@ -10,20 +9,20 @@ def do_POST(req, res)
...
   private
 
-  def handle(type, req, res)
+  def handle(_type, req, res)
     @request = req
     @response = res
 
     @request.path_info =~ /(\w+)\z/
-    @action = $1 || 'index'
+    @action = Regexp.last_match(1) || 'index'
```
byroot: I tend to disable that cop, because honestly this is just as bad if not worse. At least people are familiar with `$1` this not so much.

byroot:
LGTM beside the regexp thing.

Out of the curiosity what ruby versions Liquid target now?

byroot:
Nvm, 2+ I see.

performance/shopify/liquid.rb
```diff
@@ -1,4 +1,4 @@
-$:.unshift File.dirname(__FILE__) + '/../../lib'
+$LOAD_PATH.unshift File.dirname(__FILE__) + '/../../lib'
```
byroot: If we're 2+ we could also relplace all these `File.dirname(__FILE__)` by `__dir__`. I doubt rubocop have something for that.

fw42:
Fixed the regexp thing

byroot:
LGTM.

fw42:
hm now I broke the tests somehow

fw42:
ah I know... $0 and Regexp.last_match(0) is not the same. gr.

lib/liquid/block.rb
```diff
@@ -35,17 +35,17 @@ def warnings
       all_warnings
     end
 
-    def unknown_tag(tag, params, tokens)
+    def unknown_tag(tag, _params, _tokens)
```
pushrax: If no one really wants this rule I'd prefer not to have it.

test/unit/variable_unit_test.rb
```diff
@@ -114,43 +114,43 @@ def test_dashes
...
-    var = Variable.new(%! name_of_variable | upcase !)
+    var = Variable.new(%( name_of_variable | upcase ))
     assert_equal " name_of_variable | upcase ", var.raw
   end
 
   def test_variable_lookup_interface
     lookup = VariableLookup.new('a.b.c')
     assert_equal 'a', lookup.name
-    assert_equal ['b', 'c'], lookup.lookups
+    assert_equal %w(b c), lookup.lookups
```
pushrax: If no one really wants this rule I'd prefer not to have it.

pushrax:
:cool:

arthurnn:
:heart: rubocop

test/unit/variable_unit_test.rb
fw42: yeah removing this one

lib/liquid/block.rb
fw42: I kinda like it, I think it makes the code easier to understand. Will leave it for now.

PR #565, opened 2015-05-13
Modernize code base with __dir__ and require_relative

byroot:
@fw42 as proposed here: https://github.com/Shopify/liquid/pull/564#discussion_r30266546

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,3 @@
+machine:
+  ruby:
+    version: ruby-2.0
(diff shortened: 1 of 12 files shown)


fw42:
tests are failing

byroot:
On circle only because circle use 1.9 by default.

byroot:
I added a circle.yml and that fixed it. Not sure if it's a problem to have one.

fw42:
:+1: when :green_apple:

PR #566, opened 2015-05-13
Remove unnecessary stack pushes

pushrax:
Either the tests should fail here, or these really aren't necessary (they don't seem necessary). If this is to keep tag-created variables scoped, then there should be a test for that.

@fw42 @dylanahsmith

### DIFF
--- a/lib/liquid/tags/case.rb
+++ b/lib/liquid/tags/case.rb
@@ -37,20 +37,18 @@ module Liquid
     end
 
     def render(context)
-      context.stack do
-        execute_else_block = true
+      execute_else_block = true
 
-        output = ''
-        @blocks.each do |block|
-          if block.else?
-            return block.attachment.render(context) if execute_else_block
-          elsif block.evaluate(context)
-            execute_else_block = false
-            output << block.attachment.render(context)
-          end
+      output = ''
+      @blocks.each do |block|
+        if block.else?
+          return block.attachment.render(context) if execute_else_block
+        elsif block.evaluate(context)
+          execute_else_block = false
+          output << block.attachment.render(context)
         end
-        output
       end
+      output
     end
 
     private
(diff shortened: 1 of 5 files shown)


fw42:
FWIW I just ran this branch on the docs.shopify.com Jekyll site and the generated output is identical

fw42:
This seems (!) safe to me, since those tags don't assign any variables. But then I don't expect this to be a big performance win since we already lazily create new stack frames (only if you actually try to assign something). @boourns might have more insights.

pushrax:
Yeah, this shouldn't have a perf impact. I just thought it was strange that these calls exist, and if there is no actual reason to have them it would be good to reduce the confusion.

dylanahsmith:
It looked like those context blocks would be to pop off context for anything assigned within the block, but the assign and capture tag assign to the outermost context rather than the current context, so this might not cause problems for standard liquid tags.

However, custom tags could assign to the current context and have different output.  Here is an example:

```ruby
require 'liquid'

class CustomAssign < Liquid::Tag
    super
    markup =~ Liquid::Assign::Syntax
    @to = $1
    @from = Liquid::Variable.new($2, options)
  end

  def render(context)
    val = @from.render(context)
    context[@to] = val
    ''
  end
end

Liquid::Template.register_tag('custom_assign'.freeze, CustomAssign)

template = "{% custom_assign v = 'a' %}{% if true %}{% custom_assign v = 'b' %}{% endif %}{{ v }}"
puts Liquid::Template.parse(template).render!
```

The output is `a` on master and `b` on this branch, since the if block isn't given its own scope.

pushrax:
@dylanahsmith I agree, but is it in our contract to support that? Do we document it anywhere?

dylanahsmith:
I was just trying to assess the impact.  It looks like it might not be a problem for Shopify or most users of liquid, but I wanted to make it clear where it could break something.

fw42:
Hm. Since this is neither a performance win nor a bug fix, I would say we should close this PR and not risk breaking backwards compatibility.

PR #570, opened 2015-05-14
Fix strict conditions

### LINKED ISSUE #569: Strict parser builds conditionals in the opposite order
```ruby
Liquid::Template.parse("{% if true or true and false %} TRUE {% else %} FALSE {% endif %}").render
=> " TRUE "
Liquid::Template.parse("{% if true or true and false %} TRUE {% else %} FALSE {% endif %}", error_mode: :strict).render
=> " FALSE "
```

conversely,

```ruby
Liquid::Template.parse("{% if false and false or true %} TRUE {% else %} FALSE {% endif %}").render
=> " FALSE "
Liquid::Template.parse("{% if false and false or true %} TRUE {% else %} FALSE {% endif %}", error_mode: :strict).render
=> " TRUE "
```

This is because the lax parser [pulls tokens off the end of the list](https://github.com/Shopify/liquid/blob/3372ca8136a13af28c80b1cbe4e11be82ffbed5e/lib/liquid/tags/if.rb#L71).

I noticed this while testing out the formatter – if the AST was parsed with the strict parser it would spit out the conditions backwards.

Since `and` doesn't have precedence in Liquid, there is no obvious answer to which one is correct. Note: in general both `and` and `or` are commutative when alone, so giving one precedence over the other (as commonly done in programming languages) makes the whole expression commutative, which is why usually this ambiguity is never seen.

@Shopify/liquid @trishume

pushrax:
Starting a branch to fix #569, currently just a regression test until we decide what to do about it.

### DIFF
--- a/lib/liquid/tags/if.rb
+++ b/lib/liquid/tags/if.rb
@@ -84,14 +84,21 @@ module Liquid
     def strict_parse(markup)
       p = Parser.new(markup)
 
-      condition = parse_comparison(p)
+      conditions = [parse_comparison(p)]
 
       while op = (p.id?('and'.freeze) || p.id?('or'.freeze))
-        new_cond = parse_comparison(p)
+        conditions << op << parse_comparison(p)
+      end
+      p.consume(:end_of_string)
+
+      condition = conditions.pop
+
+      until conditions.empty?
+        op = conditions.pop
+        new_cond = conditions.pop
         new_cond.send(op, condition)
         condition = new_cond
       end
-      p.consume(:end_of_string)
 
       condition
     end
--- a/lib/liquid/tags/if.rb
+++ b/lib/liquid/tags/if.rb
@@ -83,23 +83,16 @@ module Liquid
 
     def strict_parse(markup)
       p = Parser.new(markup)
-
-      conditions = [parse_comparison(p)]
-
-      while op = (p.id?('and'.freeze) || p.id?('or'.freeze))
-        conditions << op << parse_comparison(p)
-      end
+      condition = parse_binary_comparison(p)
       p.consume(:end_of_string)
+      condition
+    end
 
-      condition = conditions.pop
-
-      until conditions.empty?
-        op = conditions.pop
-        new_cond = conditions.pop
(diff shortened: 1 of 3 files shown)


pushrax:
Turns out that this probably does affect existing templates. Liquid-C isn't implemented for `If` so the lax parser is still used currently. So, to avoid having to deal with even more broken templates, I think it makes sense to keep the lax behaviour in the strict parser (i.e. this PR).

@trishume @fw42 @dylanahsmith review?

pushrax:
We should also backport this assuming it gets merged.

lib/liquid/tags/if.rb
```diff
@@ -84,14 +84,21 @@ def lax_parse(markup)
     def strict_parse(markup)
       p = Parser.new(markup)
 
-      condition = parse_comparison(p)
+      conditions = [parse_comparison(p)]
```
pushrax: Not ideal to have to build this stack, but without recursion (implicit stack) I dunno how to implement this fix.

trishume:
This makes me :cry: because I think the lax parser behaviour is more wrong, but rule number one is don't break existing templates so oh well.

It makes sense that the way to fix the problem is a bit ugly, because you need to make a LL parser behave like an LR parser.

dylanahsmith:
I think this could be implemented cleaner with a recursive function:

```ruby
    def strict_parse(markup)
      p = Parser.new(markup)
      condition = parse_boolean_comparison(p)
      p.consume(:end_of_string)
      condition
    end

    def parse_boolean_comparison(p)
      condition = parse_comparison(p)
      if op = (p.id?('and'.freeze) || p.id?('or'.freeze))
        condition.send(op, parse_boolean_comparison(p))
      end
      condition
    end
```

pushrax:
I agree actually and in this case the perf should be marginally better I'm guessing? (no extra allocation for the array itself).

pushrax:
@dylanahsmith :ok: :ship::dash::grey_question:

dylanahsmith:
:shipit:

PR #572, opened 2015-05-15
Fixed for Ruby 1.9.3

ryanfox1985:
Hi the code fails in Ruby 1.9.3, I did a refactor :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ pkg
 .rvmrc
 .ruby-version
 Gemfile.lock
+.idea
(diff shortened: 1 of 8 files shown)


fw42:
Ruby 1.9.3 support will be dropped in the next major release of Liquid.

parkr:
I thought 3.x did that?

fw42:
Liquid 3.x dropped support for Ruby 1.8.x. Liquid 4.x will drop support for Ruby 1.9.x.

ryanfox1985:
Well, I don't understand why do you drop the support to 1.9.3.
Regards!

fw42:
Because Ruby 1.9 has been deprecated by the Ruby core team in February 2015 and we would like to be able to use features of Ruby that are exclusive to Ruby 2.x in the future (like keyword arguments for example).

PR #573, opened 2015-05-22
Make liquid error rendering optional.

dylanahsmith:
@shopify/liquid for review

## Problem

Errors that occur at render time make sense to show to theme authors for development, but it is probably not desirable to show them to the page visitor.

## Solution

If an error occurs when rendering a tag, then allow an empty string to be rendered instead of an error message.  After rendering the errors are still available on the template for error notifications and logging.

### DIFF
--- a/test/integration/error_handling_test.rb
+++ b/test/integration/error_handling_test.rb
@@ -185,4 +185,11 @@ class ErrorHandlingTest < Minitest::Test
       template.render('errors' => ErrorDrop.new)
     end
   end
+
+  def test_disabling_error_rendering
+    template = Liquid::Template.parse('This is an argument error: {{ errors.argument_error }}')
+    template.render_errors = false
+    assert_equal 'This is an argument error: ', template.render('errors' => ErrorDrop.new)
+    assert_equal [ArgumentError], template.errors.map(&:class)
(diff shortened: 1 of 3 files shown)


pushrax:
Commit message should maybe be "Make error rendering optional" haha, otherwise :+1:

test/integration/error_handling_test.rb
```diff
@@ -185,4 +185,10 @@ def test_exceptions_propagate
       template.render('errors' => ErrorDrop.new)
     end
   end
+
+  def test_disabling_error_rendering
+    template = Liquid::Template.parse('This is an argument error: {{ errors.argument_error }}')
+    template.render_errors = false
+    assert_equal 'This is an argument error: ', template.render('errors' => ErrorDrop.new)
```
fw42: also assert that the error gets stored in `errors` anyway?

fw42:
:+1:

test/integration/error_handling_test.rb
dylanahsmith: Added `assert_equal [ArgumentError], template.errors.map(&:class)`

pushrax:
:ship:

PR #574, opened 2015-05-25
Include template name with line numbers in render errors.

dylanahsmith:
@Shopify/liquid for review

## Problem

Currently the error line numbers are ambiguous if templates are included, since there is no indication of what template they occur on.

## Solution

Add a `template_name` attribute to the context which is set and reset by the include tag's render method, then set the `template_name` on the Liquid::Error when handling the error.

Note that the main template doesn't have a name, but that name could be given by setting it on the context before rendering the template.

Also note that the name given to include the template might not reflect where the template is actually stored, since that would require more integration with the file system.  E.g. in Shopify the snippets are stored under a snippets directory, so `{% include 'product' %}` would actually use `snippets/product.liquid`.  This could be implemented later, if desired, by adding a `respond_to?` check for an optional method to get the name to display for a given template name that was passed to the include tag.

### DIFF
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -41,9 +41,9 @@ module Liquid
     end
 
     def render(context)
-      partial = load_cached_partial(context)
-
       template_name = context.evaluate(@template_name_expr)
+      partial = load_cached_partial(template_name, context)
+
       context_variable_name = template_name.split('/'.freeze).last
 
       variable = if @variable_name_expr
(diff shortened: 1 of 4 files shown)


pushrax:
LGTM. It might be a good idea to make `Context#template_name` into a stack, so you can see the full hierarchy, not sure how useful that would be though.

lib/liquid/tags/include.rb
```diff
@@ -52,28 +52,33 @@ def render(context)
...
+        context.template_name = old_template_name
       end
     end
 
     private
 
-    def load_cached_partial(context)
+    def load_cached_partial(template_name, context)
       cached_partials = context.registers[:cached_partials] || {}
-      template_name = context.evaluate(@template_name_expr)
```
fw42: hm interesting, this was evaluated twice before?

fw42:
Awesome. Maybe one day we can even have Liquid lines in Ruby stack traces :-)

:+1:

lib/liquid/tags/include.rb
dylanahsmith: yup

dylanahsmith:
> It might be a good idea to make Context#template_name into a stack, so you can see the full hierarchy, not sure how useful that would be though.

We could store the full backtrace by duplicating the stack.  It would be not just template names, but also line numbers.  In that case Liquid::Error would just store a backtrace, and the line_number attribute could be converted to a method that would get the line number from the backtrace.  Should I implement that?

pushrax:
That would be pretty cool actually, and would allow the Liquid stack to be prepended to the backtrace.

fw42:
Is this ready to be merged?

dylanahsmith:
I think we will still need to keep track of the current template_name separately from the template stack, since everything above it on the stack has a line number associated with it, but the current template we don't want to have to keep updating the line number on the context, we just get it based on the tag's line number when there is an error.

I rebased, and I think this is ready to :ship:

pushrax:
:+1:

arthurnn:

PR #575, opened 2015-05-25
Raise Liquid::ZeroDivisionError instead of ZeroDivisionError.

dylanahsmith:
@pushrax & @tjoyal for review

## Problem

We should be raising a Liquid::Error when there is an error in the template rather than a bug, so `{{ 1 | divided_by: 0 }}` should cause a ZeroDivisionError.

## Solution

Raise a `Liquid::ZeroDivisionError` instead of `::ZeroDivisionError` similar to how we raise Liquid::ArgumentError instead of ArgumentError for template errors.

### DIFF
--- a/lib/liquid/errors.rb
+++ b/lib/liquid/errors.rb
@@ -57,4 +57,5 @@ module Liquid
   StackLevelError = Class.new(Error)
   TaintedError = Class.new(Error)
   MemoryError = Class.new(Error)
+  ZeroDivisionError = Class.new(Error)
 end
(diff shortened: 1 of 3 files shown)


pushrax:
:ok:

fw42:
:+1:

tjoyal:
:+1:

I wonder if we should add coverage for this or not in https://github.com/Shopify/liquid/blob/master/test/integration/standard_filter_test.rb

dylanahsmith:
There is already test coverage, so I'm not too worried.

PR #576, opened 2015-05-27
Allow the exception handler to convert exceptions to hide error messges

dylanahsmith:
@fw42, @pushrax for review
cc @arthurnn

## Problem

It should be possible to control what is output for non-Liquid::Error exceptions, since these exceptions might expose internal details of the application.

## Solution

Allow the exception handler to return a string, which will be output for the error, or an exception that the original exception will be replaced with.

https://github.com/Shopify/liquid/pull/573 made it possible to suppress the output of errors.  This can now be done with an exception handler that always returns `''`, so I removed that unreleased feature.

I kept the old behaviour of allowing `true` to be returned to re-raise the exception, although this could have been implemented with an exception handler that just called `raise`.  I changed the internal usage of that within liquid for clarity.

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -13,7 +13,7 @@ module Liquid
   #   context['bob']  #=> nil  class Context
   class Context
     attr_reader :scopes, :errors, :registers, :environments, :resource_limits
-    attr_accessor :exception_handler, :render_errors
+    attr_accessor :exception_handler
 
     def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_errors = false, resource_limits = nil)
       @environments     = [environments].flatten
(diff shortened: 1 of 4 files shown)


lib/liquid/context.rb
```diff
@@ -68,9 +67,22 @@ def handle_error(e, token = nil)
         e.set_line_number_from_token(token)
       end
 
+      output = nil
+
+      if exception_handler
+        result = exception_handler.call(e)
+        case result
+        when Exception
```
fw42: I don't understand why this works. If you call `raise` above, doesn't that make the exception bubble up? Why does this work without a `rescue` here?

fw42: oh sorry, that's not what's happening here. misread.

lib/liquid/context.rb
```diff
@@ -68,9 +67,22 @@ def handle_error(e, token = nil)
...
+          e = result
+          e.set_line_number_from_token(token) if e.is_a?(Liquid::Error)
```
fw42: can you add a test for this line? I think removing it wont break any tests right now, will it?

fw42:
I'm fine with this :+1:

lib/liquid/context.rb
dylanahsmith: Modified a test to use `line_numbers: true` so this code path gets tested.

pushrax:
:+1: - I agree with the backwards compatibility method you chose

PR #577, opened 2015-05-28
Fix exception from using an empty string for the table row collection.

dylanahsmith:
@fw42 & @tjoyal for review

## Problem

Liquid like `{% tablerow product in collections.frontpage.products cols:3  %}` was resulting in a `Liquid error: undefined method `each' for \"\":String` error when the shop doesn't have a frontpage collection.  This empty string was coming from an EmptyDrop that we returned when a collection couldn't be found.

This was inconsistent with the `for` tag, which would treat an empty string as an empty collection.  Both the `for` and `tablerow` tags will treat a non-empty string as a collection with a single item.  It is also inconsistent with how liquid worked with ruby 1.8.7, where String#each was the same as String#each_line.

## Solution

Treat an empty string the same as an empty collection.

### DIFF
--- a/lib/liquid/utils.rb
+++ b/lib/liquid/utils.rb
@@ -17,7 +17,7 @@ module Liquid
       index = 0
 
       # Maintains Ruby 1.8.7 String#each behaviour on 1.9
-      return [collection] if non_blank_string?(collection)
+      return collection != ''.freeze ? [collection] : [] if collection.is_a?(String)
 
       collection.each do |item|
         if to && to <= index
--- a/lib/liquid/utils.rb
+++ b/lib/liquid/utils.rb
@@ -17,7 +17,9 @@ module Liquid
       index = 0
 
       # Maintains Ruby 1.8.7 String#each behaviour on 1.9
-      return collection != ''.freeze ? [collection] : [] if collection.is_a?(String)
+      if collection.is_a?(String)
+        return collection.empty? ? [] : [collection]
+      end
 
       collection.each do |item|
         if to && to <= index
--- a/lib/liquid/utils.rb
+++ b/lib/liquid/utils.rb
@@ -8,10 +8,6 @@ module Liquid
       end
     end
 
-    def self.non_blank_string?(collection)
-      collection.is_a?(String) && collection != ''.freeze
-    end
-
     def self.slice_collection_using_each(collection, from, to)
       segments = []
       index = 0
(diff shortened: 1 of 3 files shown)


lib/liquid/utils.rb
```diff
@@ -17,7 +17,7 @@ def self.slice_collection_using_each(collection, from, to)
       index = 0
 
       # Maintains Ruby 1.8.7 String#each behaviour on 1.9
```
fw42: does this comment still make sense now?

fw42:
any backwards-compatibility concerns here?

lib/liquid/utils.rb
```diff
@@ -17,7 +17,7 @@ def self.slice_collection_using_each(collection, from, to)
...
-      return [collection] if non_blank_string?(collection)
+      return collection != ''.freeze ? [collection] : [] if collection.is_a?(String)
```
fw42: I find this line hard to read (the combination of a `if` suffix and a ternary). Can you make this a bit more explicit?

Maybe
```ruby
if collection.is_a?(String)
  return collection != ''.freeze ? [collection] : []
end
```

fw42:
:+1:

lib/liquid/utils.rb
parkr: is `collection != ''.freeze` faster than `!collection.empty?`

lib/liquid/utils.rb
dylanahsmith: yes, this is still for backwards compatibility

lib/liquid/utils.rb
dylanahsmith: `!collection.empty?` is faster and more readable, so I changed this to

```ruby
      if collection.is_a?(String)
        return collection.empty? ? [] : [collection]
      end
```

dylanahsmith:
> any backwards-compatibility concerns here?

No, because the code this affects is currently outputting a liquid error.

tjoyal:
I think your changes will need to be reconsolidated with 
https://github.com/Shopify/liquid/blob/table-row-blank-string-collection/lib/liquid/tags/for.rb#L194

Should `for.rb#L194` be this?
`collection.respond_to?(:each) || collection.is_a?(String)`

I think this can be done safely, but you would need to re-validate.

dylanahsmith:
> Should `for.rb#L194` be this?
> `collection.respond_to?(:each) || collection.is_a?(String)`

Actually, I think the `collection.is_a?(String)` check is also redundant.  We could probably also move the `collection.respond_to?(:each)` check into `Utils.slice_collection` for consistency, and to avoid another potential error from some other non-iterable being used.

tjoyal:
:+1:

PR #578, opened 2015-05-28
Handle some more standard filter errors.

dylanahsmith:
@tjoyal & @pushrax for review

Similar to pull #575, this tries to avoid some non-Liquid::Error exceptions from being raised from filters:

* Rounding NaN raises a Liquid::FloatDomainError instead of FloatDomainError
* modulo 0 raises a Liquid::ZeroDivisionError instead of ZeroDivisionError
* `remove` and `remove_first` casts both arguments to strings to avoid an `TypeError: wrong argument type Fixnum (expected Regexp)` exception


pushrax:
:ok:

fw42:
:+1:

PR #579, opened 2015-05-28
Ensure For@reversed is a boolean

pushrax:
Currently running some AST comparisons between the strict and lax parser, this is the one difference that appears in almost every template. Nice to keep one less string around as well.


fw42:
:+1:

PR #581, opened 2015-06-02
Raise SyntaxError if raw tag is unclosed

pushrax:
See test case.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -42,7 +42,7 @@ module Minitest
       assert_match expected, Template.parse(template).render!(assigns), message
     end
 
-    def assert_match_syntax_error(match, template, registers = {})
+    def assert_match_syntax_error(match, template, assigns = {})
       exception = assert_raises(Liquid::SyntaxError) {
         Template.parse(template).render(assigns)
       }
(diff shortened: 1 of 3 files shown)


test/test_helper.rb
```diff
@@ -42,7 +42,7 @@ def assert_template_result_matches(expected, template, assigns = {}, message = n
       assert_match expected, Template.parse(template).render!(assigns), message
     end
 
-    def assert_match_syntax_error(match, template, registers = {})
+    def assert_match_syntax_error(match, template, assigns = {})
       exception = assert_raises(Liquid::SyntaxError) {
         Template.parse(template).render(assigns)
```
fw42: hu? how did this ever work? or was this method not called anywhere?

fw42:
:+1:

test/test_helper.rb
pushrax: It worked because `Template.parse` will raise an exception before `render` is called (and its arguments are evaluated).

gauravmc:
Why not the error when the tag is initialized (like elsewhere) instead of doing it parse time?

pushrax:
@gauravmc this is not quite fixing your original case directly – it's actually ensuring the presence of an `{% endraw %}` following a `{% raw %}`.

PR #582, opened 2015-06-02
Ensure raw tag has no arguments

pushrax:
We usually don't enforce the lack of arguments to zero-argument tags (e.g. break, continue, etc), but I think it's a good practise.

cc @gauravmc

### DIFF
--- a/lib/liquid/tags/raw.rb
+++ b/lib/liquid/tags/raw.rb
@@ -1,6 +1,15 @@
 module Liquid
   class Raw < Block
+    Syntax = /\A\s*\z/om
     FullTokenPossiblyInvalid = /\A(.*)#{TagStart}\s*(\w+)\s*(.*)?#{TagEnd}\z/om
+
+    def initialize(tag_name, markup, options)
+      super
+
+      unless markup =~ Syntax
+        raise SyntaxError.new(@options[:locale].t("errors.syntax.tag_unexpected_args".freeze, tag: tag_name))
+      end
+    end
 
     def parse(tokens)
       @body = ''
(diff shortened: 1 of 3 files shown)


parkr:
:+1:

lib/liquid/tags/raw.rb
```diff
@@ -1,7 +1,16 @@
 module Liquid
   class Raw < Block
+    Syntax = /\A\s*\z/om
```
gauravmc: Don't need `o`.. probably also `m`

gauravmc:
:+1:

test/integration/tags/raw_tag_test.rb
```diff
@@ -24,6 +24,7 @@ def test_open_tag_in_raw
   end
 
   def test_invalid_raw
-    assert_match_syntax_error /tag was never closed/, '{% raw } foo {% endraw %}'
```
fw42: where did this test go? why did you delete it?

pushrax: It ends up hitting the new case and getting the other syntax error, so I edited it to new line at that location.

fw42: I would prefer to keep all three tests. It's not obvious to me that the new tests cover the exact same codepaths that the old test did.

pushrax: Sure.

fw42:
:+1:

gauravmc:
Would need a gem bump after.. require this is Shopify: https://github.com/Shopify/shopify/pull/46102

PR #583, opened 2015-06-02
Raise a Liquid::ArgumentError in slice filter for invalid integers.

dylanahsmith:
@tjoyal & @fw42 for review

If an invalid integer is passed to slice for the offset of length arguments, then raise a Liquid::ArgumentError instead of a non-Liquid::Error, since this is a template bug.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -323,7 +323,7 @@ module Liquid
       begin
         Integer(num)
       rescue ::ArgumentError
-        raise Liquid::ArgumentError, "invalid integer '#{num}'"
+        raise Liquid::ArgumentError, "invalid integer"
       end
     end
 
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -46,8 +46,8 @@ module Liquid
     end
 
     def slice(input, offset, length = nil)
-      offset = Integer(offset)
-      length = length ? Integer(length) : 1
+      offset = to_integer(offset)
+      length = length ? to_integer(length) : 1
 
       if input.is_a?(Array)
         input.slice(offset, length) || []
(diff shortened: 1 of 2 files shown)


tjoyal:
What was the original error looking like? Could this new custom error be abused into generating a new xss exploit point?

dylanahsmith:
> What was the original error looking like? Could this new custom error be abused into generating a new xss exploit point?

I was seeing the error `can't convert nil into Integer`

If the input were a string, then it would also include the input

```
irb> Integer('<script>alert(1)</script>')
ArgumentError: invalid value for Integer(): "<script>alert(1)</script>"
```

but that would only be exploitable if the visitor had control of the input.

tjoyal:
:+1: then as it was already exploitable

lib/liquid/standardfilters.rb
```diff
@@ -317,6 +317,16 @@ def default(input, default_value = "".freeze)
 
     private
 
+    def to_integer(num)
+      return num if num.is_a?(Integer)
+      num = num.to_s
+      begin
+        Integer(num)
```
fw42: can you add a test for this line? i.e. one where the argument is a string that can be converted to an integer?

fw42:
:+1:

lib/liquid/standardfilters.rb
```diff
@@ -317,6 +317,16 @@ def default(input, default_value = "".freeze)
...
+      rescue ::ArgumentError
+        raise Liquid::ArgumentError, "invalid integer '#{num}'"
```
fw42: do we really need to put the `#{num}` in here? does that give us much? I'm concerned now about the possible XSS you mentioned.

dylanahsmith: it can help debugging, but I guess it can be removed

lib/liquid/standardfilters.rb
dylanahsmith: done

lib/liquid/standardfilters.rb
dylanahsmith: removed ` '#{num}'` from the error message

PR #585, opened 2015-06-03
Template Specific tags

gauravmc:
@fw42 @dylanahsmith @pushrax as per the discussion in https://github.com/Shopify/shopify/pull/46102/files#r31581939, simple prototype of template specific tags.


gauravmc:
Closed in favour of #590

PR #586, opened 2015-06-03
Avoid an exception from checking if a string contains a non-string.

dylanahsmith:
@tjoyal & @fw42 for review

## Problem

The contains operator can raise an exception if the left operand is a string, but the right operand isn't.

`{% if string contains non_string %}`

because String#include? will raise a `TypeError: no implicit conversion of #{obj.class} into String` exception, which isn't a Liquid::Error despite being a template bug.

## Solution

If the left operand is a string, then convert the right operand to a string using `.to_s` to avoid exceptions.

### DIFF
--- a/lib/liquid/condition.rb
+++ b/lib/liquid/condition.rb
@@ -16,7 +16,12 @@ module Liquid
       '>='.freeze => :>=,
       '<='.freeze => :<=,
       'contains'.freeze => lambda do |cond, left, right|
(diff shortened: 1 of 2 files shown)


fw42:
:+1: thanks for doing all this exception-related cleanup!

tjoyal:
If I only look at the changes :+1: :smile_cat:

PR #588, opened 2015-06-03
add global_filters

boobooninja:
Allow Template.parse options to accept a "global_filters" array of filters that would be applied to every liquid variable.

Example:
```
template = Liquid::Template.parse("Apply filter to {{ one }} and {{ two }} and {{ three }}.", global_filters: ['truncate: 5'])
template.render({'one' => 'first variable', 'two' => 'second variable', 'three' => 'third variable'})
=> "Apply filter to fi... and se... and th...."
```


Rakefile
```diff
@@ -76,3 +76,7 @@ desc "Run example"
 task :example do
   ruby "-w -d -Ilib example/server/server.rb"
 end
+
+task :console do
+  exec 'irb -I lib -r ./lib/liquid.rb'
```
fw42: I like this, but can you make it a separate PR? also I think `-r liquid` would be enough, don't need the `-r lib/...`

fw42:
I'm not totally opposed to what you are trying to do, but the implementation feels wrong to me (modifying the template while it's being parsed, basically).

@pushrax @dylanahsmith thoughts?

test/integration/variable_test.rb
```diff
@@ -79,4 +79,10 @@ def test_hash_with_default_proc
   def test_multiline_variable
     assert_equal 'worked', Template.parse("{{\ntest\n}}").render!('test' => 'worked')
   end
+
+  def test_global_options_on_variable
+    template = Template.parse(%({{test}}), global_filters: ['truncate: 5'])
```
dylanahsmith: I don't see why liquid syntax should be used to define global filters.  I think taking a ruby proc would make more sense, and I think this could be used to implement taint checking.

dylanahsmith:
I agree that a feature like this could be useful, since it could be used to automatically escape output by default, similar to rails views, rather then the current practice of escaping user input.

test/integration/variable_test.rb
pushrax: Agree

boobooninja: @dylanahsmith Thanks for the feedback. Can you elaborate a little bit more on what you mean by using a Proc instead of using liquid syntax?

I'm guessing you would want these filters applied at some other point in the process than I currently have implemented? aka during parsing.

dylanahsmith: It isn't needed during parsing, just during rendering, so it would make sense to store the proc as part of the Liquid::Context.  So instead of [`taint_check(obj)` in Variable#render](https://github.com/Shopify/liquid/blob/7b4398d0c4f5b5b766677c8ad1ed159fed4cb041/lib/liquid/variable.rb#L77) you could do something like `context.global_variable_filter.call(obj)`.

boobooninja: @dylanahsmith I understand that it is not really needed until rendering so could be applied at that time and even perhaps passed in as an option to the render method. I am wondering why you don't like using liquid syntax though. The idea is to make it as easy for the end-user as possible.
This may be due to my lack of fully understanding the codebase but why should the user need to create a Proc for each filter they want to pass in as a global filter when they already know the liquid filter syntax?

dylanahsmith: "end-user" is a vague term.  Do you mean the template author?  Because there is no where in the template for them to specify a global filter.

If you mean the programmer that is embedding liquid into their application, then no, they aren't necessarily more familiar with liquid syntax, and allowing them to use any ruby code gives them more control.  E.g. they can easily call `CGI.escapeHTML(input)` if they want to use this for escaping, but they couldn't easily add taint checking.

pushrax:
Merged in #610

PR #590, opened 2015-06-03
Local Tags

gauravmc:
@dylanahsmith @fw42 @pushrax this comes from the discussions happening around https://github.com/Shopify/shopify/pull/46102#discussion_r31594674.

With this change you can now inherit from `Liquid::Document` and do things like: override methods like `registered_tags`, so that parser ignores all global tags and access only the ones that come from it. Better way to do something like #585 for our use case.

See usage here: https://github.com/Shopify/shopify/commit/24c07ee1bc4ac50a87298f3b47634a5147bb31fc

### DIFF
--- a/test/unit/tokenizer_unit_test.rb
+++ b/test/unit/tokenizer_unit_test.rb
@@ -22,18 +22,20 @@ class TokenizerTest < Minitest::Test
   end
 
   def test_calculate_line_numbers_per_token_with_profiling
-    template = Liquid::Template.parse("", :profile => true)
-
-    assert_equal [1],       template.send(:tokenize, "{{funk}}").tokens.map(&:line_number)
-    assert_equal [1, 1, 1], template.send(:tokenize, " {{funk}} ").tokens.map(&:line_number)
-    assert_equal [1, 2, 2], template.send(:tokenize, "\n{{funk}}\n").tokens.map(&:line_number)
-    assert_equal [1, 1, 3], template.send(:tokenize, " {{\n funk \n}} ").tokens.map(&:line_number)
+    assert_equal [1],       tokenize("{{funk}}", true).map(&:line_number)
+    assert_equal [1, 1, 1], tokenize(" {{funk}} ", true).map(&:line_number)
+    assert_equal [1, 2, 2], tokenize("\n{{funk}}\n", true).map(&:line_number)
+    assert_equal [1, 1, 3], tokenize(" {{\n funk \n}} ", true).map(&:line_number)
   end
 
   private
 
-  def tokenize(source)
-    tokenizer = Liquid::Tokenizer.new(source)
-    tokenizer.tokens
+  def tokenize(source, line_numbers = false)
+    tokenizer = Liquid::Tokenizer.new(source, line_numbers)
+    tokens = []
+    while t = tokenizer.shift
+      tokens << t
+    end
+    tokens
   end
 end
--- a/test/unit/tokenizer_unit_test.rb
+++ b/test/unit/tokenizer_unit_test.rb
@@ -24,15 +24,16 @@ class TokenizerTest < Minitest::Test
   def test_calculate_line_numbers_per_token_with_profiling
     template = Liquid::Template.parse("", :profile => true)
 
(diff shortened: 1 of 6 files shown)


lib/liquid/ruby_tokenizer.rb
```diff
@@ -0,0 +1,28 @@
+module Liquid
+  class RubyTokenizer
```
gauravmc: `RubyTokenizer` coz `Tokenizer ` can come from liquid-c

pushrax: We should just add `Tokenizer` to ruby liquid and patch it in liquid c.

dylanahsmith: Agreed.  We should just have an interface similar to liquid-c, where it passes a Tokenizer object rather than an array of tokens to Document.parse, and Tokenizer just implements the `shift` method, which you can delegate to `@tokens.shift` in the ruby implementation.

From liquid-c it might be easier to define LiquidC::Tokenizer to delegate to until it exposes line numbers.

gauravmc: I've updated this to just be `Tokenizer`, following along the lines of liquid-c's implementation of it

test/unit/tokenizer_unit_test.rb
```diff
@@ -24,15 +24,24 @@ def test_tokenize_blocks
   def test_calculate_line_numbers_per_token_with_profiling
     template = Liquid::Template.parse("", :profile => true)
 
-    assert_equal [1],       template.send(:tokenize, "{{funk}}").map(&:line_number)
-    assert_equal [1, 1, 1], template.send(:tokenize, " {{funk}} ").map(&:line_number)
-    assert_equal [1, 2, 2], template.send(:tokenize, "\n{{funk}}\n").map(&:line_number)
-    assert_equal [1, 1, 3], template.send(:tokenize, " {{\n funk \n}} ").map(&:line_number)
+    assert_equal [1],       to_a(template.send(:tokenize, "{{funk}}")).map(&:line_number)
```
fw42: now that this thing is it's own class, we don't need this `template.send(:tokenize)` hack anymore, you can just unit-test the Tokenizer class, no?

fw42: or can't you use the `tokenize` test helper method here?

fw42:
Code looks good to me, but @pushrax and @trishume should take a look.

(Does this PR play nice with liquid-c?)

trishume:
This will interfere a little bit with liquid-c by making it impossible to turn off the tokenizer, but it is easy to fix in liquid-c. 

Basically liquid-c has always defined a `Liquid::Tokenizer` class with initialize and shift methods, then it overrides the `tokenize` method to either use the class or the original ruby method base on the liquid c enabled flag. With this change the liquid-c tokenizer should override the tokenizer in this PR with a compatible one so that either way it will use the liquid-c tokenizer, regardless of the flag.

Other than that should be fine, but I haven't worked on liquid-c in a while so there might be some new stuff it will interfere with.

gauravmc:
@trishume I'd PRed this in liquid-c for that, have a look https://github.com/Shopify/liquid-c/pull/22

test/unit/tokenizer_unit_test.rb
gauravmc: yea true

dylanahsmith: Yup, the `tokenize` helper just needs an options hash, then the `line_numbers: true` option can be passed into the tokenizer.

dylanahsmith: LGTM other than this.

gauravmc: done

arthurnn:
Did we add a changelog for this? do we need one?

gauravmc:
@arthurnn donno if we need one cc @fw42 @dylanahsmith

fw42:
I don't think this PR needs a changelog entry, it's neither a new feature nor a bugfix and should be mostly backwards compatible. It's just a refactor, really.

Issue #591, opened 2015-06-04
Using "blank" as a variable name


fw42:
Weird things happen. Saw something like this in production:
```
{% assign blank = "" %}
...
{% assign something = blank %}
...
{{ something }}
```

which renders `"Liquid error: undefined method 'to_liquid' for :blank?:Symbol"`, which is kinda shitty.

Should have some better behaviour.

nickpearson:
I've always thought the `== blank` check was strange, since it's not really testing for equality.

If going forward Liquid allows `blank` to be used as a variable name, how about adding `is` as an alternative to `==` for special checks like `blank` and `empty`? `{% if something is blank %}`

pushrax:
`blank` is a keyword in liquid, like `true`, and is parsed in the exact same way. I don't think it should be assignable. There are two issues here, the first (outputting `blank`) is fixed by #592, the second (assigning to `blank`) should work in the same way as assigning to `true`.

pushrax:
If people agree with my thoughts on the treatment of assigning to `blank` then we can close this, since it's what the current behaviour is.

fw42:
I think the current behaviour (you can't assign to a variable named "blank") seems fine to me.

fw42:
Arguably we should error and not be silent...

fw42:
I think we can close this issue for now since I wouldn't consider this a "bug" at this point. Please reopen if you disagree.

pushrax:
Yeah, in the same way that we should error if you assign to `true` or `14` (but we don't).

PR #592, opened 2015-06-04
blank and empty as variable names

fw42:
Fixes https://github.com/Shopify/liquid/issues/591.

Not sure if this is a good fix, feels kinda hacky, but seems to work.

@pushrax @dylanahsmith


pushrax:
Yeah, this is basically what I was thinking. Hopefully external tags don't rely on this.

dylanahsmith:
That doesn't really seem hacky.

:+1:

arthurnn:

PR #593, opened 2015-06-04
Rename 'has_key?' and 'has_interrupt?'

fw42:
rubocop doesn't like `has_bla?` and `is_bla?` etc. and me neither... Should be safe to change, right?

@pushrax @dylanahsmith


pushrax:
weeee more breaking changes

dylanahsmith:
sure :+1:

arthurnn:
is this public API change? if so, I think we need a changelog

PR #595, opened 2015-06-05
Fix uniq filter with string input

fw42:
because people do stupid shit in templates and run the uniq filter on string inputs and such...

@arthurnn @dylanahsmith


dylanahsmith:
:+1:

pushrax:
:+1:

arthurnn:

PR #596, opened 2015-06-05
Run tests with latest liquid/c gem

fw42:
Turns out https://github.com/Shopify/liquid/pull/592 is actually broken if liquid/c is present.

This PR makes it so that we run tests twice, once with and once without liquid/c enabled.

(Tests in this PR are going to fail for now. Justin is working on a fix in liquid/c.)

@pushrax @trishume


pushrax:
I'm not sure about using `master` always, it means that when you check out an old liquid version the tests may fail...

Should we lock the liquid-c SHA in the gemfile?

pushrax:
Also, it will fail on non-MRI rubies :(

fw42:
Hmm good points

trishume:
nice 2x2 test matrix, quadruple the error messages if you screw up something globally :smiling_imp: 

In terms of version issues, maybe gate it by an environment variable that is set in CI?

test/test_helper.rb
```diff
@@ -15,6 +15,16 @@
 end
 Liquid::Template.error_mode = mode
 
+if ENV['LIQUID-C'] == '1'
+  puts "-- LIQUID C"
+  begin
+    require 'liquid/c'
+  rescue LoadError
+    puts "LiquidC not supported."
```
parkr: Nitpick, but `$stderr.puts` here?

parkr:
:+1:

fw42:
How about instead of using master, we use the latest released version of the liquid-c gem?

pushrax:
@fw42 I think that still has the same problem, and also makes updating liquid and liquid-c in tandem difficult.

fw42:
Hm I don't like the idea of hardcoding a liquid/c ref/version, we will forget to bump that.

PR #599, opened 2015-06-15
Handle nils gracefully when using the "sort" filter

carsonreinke:
Since there is no reject filter, allow sorting when a value could be nil.


carsonreinke:
I guess I mean compact, not reject.

fw42:
FWIW I would be okay with adding a `compact` filter

fw42:
Hmm. Not super sure how I feel about this. Can you explain your use-case?

pushrax:
:+1: to adding a compact filter instead of this. As someone who prefers static type systems this feels wrong to me.

carsonreinke:
Yep, it looked weird.

PR #600, opened 2015-06-15
Added "compact" filter

carsonreinke:
Instead of changing "sort" filter (https://github.com/Shopify/liquid/pull/599), add a compact filter.

Obviously, compact using a property is a little different.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,9 @@ Metrics/BlockNesting:
   Exclude:
     - 'lib/liquid/block_body.rb'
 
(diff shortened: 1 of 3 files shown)


fw42:
Looks like a Rubocop error. We should just ignore the "module/class has too many lines" errors for now.

fw42:
Code looks good to me. Thanks! :+1:

fw42:
@pushrax can you review this too?

carsonreinke:
Awesome!

PR #601, opened 2015-06-16
Use to_integer instead of to_i on arguments

davidcornu:
`truncate` and `truncatewords` use `to_i` on arguments which can cause exceptions.

PR #602, opened 2015-06-16
Use to_integer for truncatewords filter

fw42:
This makes the `truncatefilter` raise a proper `Liquid::ArgumentError` if someone passes something as argument that is not an integer.

@dylanahsmith @pushrax


fw42:
Closing in favour of https://github.com/Shopify/liquid/pull/601. @davidcornu

PR #603, opened 2015-06-19
Format changelog attribution to include one name only

pushrax:
I felt that a disproportional amount of each line was taken up with attribution, not sure if others agree.


fw42:
sure, I don't care

PR #604, opened 2015-06-23
sorting on nested properties

untra:
pr to provide support for sorting on nested properties in collections. Created to address [Jekyll #3351](https://github.com/jekyll/jekyll/issues/3351)

```
{% assign sorted_posts = filtered_posts | sort: 'book.author' %}
```

Comes with tests. @fw42

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -116,7 +116,9 @@ module Liquid
       ary = InputIterator.new(input)
       if property.nil?
         ary.sort
-      elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
+      elsif property.include? '.'
+        nested_sort(input, property)
+      elsif ary.first.respond_to?(:[]) && accepts?(ary.first, property)
         ary.sort { |a, b| a[property] <=> b[property] }
       elsif ary.first.respond_to?(property)
         ary.sort { |a, b| a.send(property) <=> b.send(property) }
@@ -375,6 +377,43 @@ module Liquid
       end
     rescue ArgumentError
       nil
+    end
+
+    def nested_respond_to?(input, property)
+      return true if property.empty?
+      p = property.split('.').first
+      q = property[/#{p}.?([\w.]*)/, 1]
+      if input.respond_to?(:[]) && accepts?(input, p)
+        return nested_respond_to?(input[p], q)
+      elsif input.respond_to?(p)
+        return nested_respond_to?(input.send(p), q)
+      end
+      false
+    end
+
+    def nested_send(input, property)
+      return input if property.empty?
+      p = property.split('.').first
+      q = property[/#{p}.?([\w.]*)/, 1]
+      if input.respond_to?(:[]) && accepts?(input, p)
+        return nested_send(input[p], q)
+      elsif input.respond_to?(p)
+        return nested_send(input.send(p), q)
+      end
+      nil
+    end
+
+    def accepts?(input, index)
+      !input[index].nil?
+      rescue TypeError
+        false
+    end
+
+    def nested_sort(input, property = nil)
+      ary = InputIterator.new(input)
+      if nested_respond_to?(ary.first, property)
+        ary.sort { |a, b| nested_send(a, property) <=> nested_send(b, property) }
+      end
     end
 
     def apply_operation(input, operand, operation)
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -16,7 +16,7 @@ Metrics/AbcSize:
 # Offense count: 16
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 314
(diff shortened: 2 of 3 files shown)


untra:
Travis build fails because the addition of my tests makes the StandardFiltersTest class too big. :muscle:

fw42:
Yeah don't worry about the Rubocop fail, I will fix that later

fw42:
Hm I'm kinda worried that adding something like this to the sort filter will make people assume that something similar works for all other filters (`map` for example). If we want to add something like this, I think it should be more generic.

@dylanahsmith @pushrax what do you think?

lib/liquid/standardfilters.rb
```diff
@@ -377,6 +379,43 @@ def to_date(obj)
...
+        return nested_send(input[p], q)
+      elsif input.respond_to?(p)
+        return nested_send(input.send(p), q)
+      end
+      nil
+    end
+
+    def accepts?(input, index)
+      !input[index].nil?
+      rescue TypeError
```
fw42: why is this necessary?

fw42: is there a test that will fail if you remove this?

untra: `nested_send` is called by the `nested_sort`, which should now be called by `sort` when `property.include? '.'` (line 119, standardfilters.rb) . If you remove it, the tests I wrote will not pass.

untra: I added a number of helper methods that should be called by sort when the property being passed  is nested. I put them with the private methods. Is this the right location for that code?

I would also like to add tests for these helper methods. What is the right place to put these tests? They cannot be called from standard_filter_test.rb through `@filters`

untra:
@dylanahsmith @pushrax @fw42 any thoughts? Myself and others *really* want this functionality. Would nested_sort this work better as it's own filter?
```
{% assign sorted_posts = filtered_posts | nested_sort: 'book.author' %}
```
This approach would mean no changes to the sort filter, and `nested_sort` would look like it's own distinct filter. It would also mean `map` should get it's own `nested_map` filter.
Thoughts?

Fresheyeball:
Come on people! Mergy merge merge.... otherwise we have to live on a fork. No one wants that. We don't want that. You don't want that. Its no fun.

fw42:
Custom filters can be implemented easily without having to fork Liquid. This is not blocking anything.

Fresheyeball:
Fair enough. I'd just like the sorting of nested properties to be in liquid for max re-usablity.

pushrax:
Before we merge something like this, I'd want to have a thorough discussion of how to generalize the syntax and make it applicable to all filters that access members, like @fw42 mentioned earlier. One-off syntax extensions like this fragment the language and make it much more difficult to maintain.

Fresheyeball:
maybe I am confused. How is the proposed code a one-off syntax extension? Its just extending the functionality of existing syntax by extending the functionality of an existing function. 

``` 
  | sort: "foo"
``` 

is existing syntax

```
  | sort: "foo.bar"
```

does not seem like a new syntax to me. Please help me understand your concern.

pushrax:
For one, it's not backward compatible. At the moment, `| sort: "foo.bar"` will work if `[]("foo.bar")` is defined.

Syntax in Liquid is not formally specified (yet) so the boundary is a little fuzzy, but you're right that I am probably using the wrong terminology. Semantics then.

Fresheyeball:
@pushrax I removed it because @untra explained the objection to me. I'm not a ruby guy, and it did not make immediate sense. Its a clear and sensible objection.

untra:
@Fresheyeball lets make our own filter, like @fw42 suggested. We were unaware that was a option. That way we won't be rushing them into a merge.

The best approach to generalizing this sort of functionality would require using the helper functions I wrote, `nested_respond_to?` and `nested_send`. Ideally, these should operate just like native ruby `respond_to?` and `send`, but will split inputs on the '.' and recurse. Take a look at the implementation of `nested_sort`:
```ruby
    def nested_sort(input, property = nil)
      if nested_respond_to?(ary.first, property)
        ary.sort { |a, b| nested_send(a, property) <=> nested_send(b, property) }
      end
    end
```
and the current implementation of `sort`:
```ruby
    def sort(input, property = nil)
      if property.nil?
        ary.sort
        ary.sort { |a, b| a[property] <=> b[property] }
      elsif ary.first.respond_to?(property)
        ary.sort { |a, b| a.send(property) <=> b.send(property) }
      end
    end
```
Similar, no? These helper functions need their own tests though. So I have two questions:
- Where can I / should I put tests for `nested_respond_to?` and `nested_send`, and other helpers I want to write?
- What other filters besides `map` should get generalized this way?

remimarenco:
Is it live somewhere? It would be very useful :)

untra:
Closing this PR because stale
still might be a nice liquid feature to add

PR #605, opened 2015-06-25
add global filters using procs

boobooninja:
apply global filters by passing an array of procs to the render method as :global_filters option

### DIFF
--- a/lib/liquid/tag.rb
+++ b/lib/liquid/tag.rb
@@ -1,7 +1,7 @@
 module Liquid
   class Tag
     attr_accessor :options, :line_number
-    attr_reader :nodelist, :warnings
+    attr_reader :nodelist, :warnings, :tag_name
     include ParserSwitching
 
     class << self
--- a/test/unit/tag_unit_test.rb
+++ b/test/unit/tag_unit_test.rb
@@ -13,4 +13,9 @@ class TagUnitTest < Minitest::Test
     tag = Tag.parse("long_tag", "param1, param2, param3", [], {})
     assert_equal("long_tag param1, param2, param3", tag.raw)
   end
+
+  def test_tag_name_should_return_name_of_the_tag
+    tag = Tag.parse("some_tag", [], [], {})
+    assert_equal 'some_tag', tag.tag_name
+  end
 end


fw42:
what's the use case for this? what problem does this solve?

pushrax:
As discussed in https://github.com/Shopify/liquid/pull/588

fw42:
Ah I see, I forgot about that

boobooninja:
Yes, sorry I didn't mention this is my attempt at implementing the suggestions in #588 

The original idea was to allow a filter to be applied to all variables in the template without having to modify the template and actually place the filter in all the tags.

lib/liquid/context.rb
```diff
@@ -47,6 +48,22 @@ def add_filters(filters)
       @strainer = nil
     end
 
+    def add_global_filters(global_filters)
+      gf = [global_filters].flatten.compact
+
+      gf.each do |p|
+        if p.respond_to?(:call)
```
fw42: what's this check for? why would you call `add_global_filters` with something that doesn't respond to `call`?

lib/liquid/template.rb
```diff
@@ -183,20 +183,15 @@ def render(*args)
...
-        end
+        context.add_filters(options[:filters]) if options[:filters]
 
-        if options[:exception_handler]
-          context.exception_handler = options[:exception_handler]
-        end
-      when Module
-        context.add_filters(args.pop)
-      when Array
+        context.add_global_filters(options[:global_filters]) if options[:global_filters]
```
fw42: Just to clarify: This applies the global filters *after* the "normal" filters have already been applied?

fw42:
Can you add a test with both global and "normal" filters and assert that they are applied in the order that you want them to? Also add a test with more than one global filter and assert the correct order there.

fw42:
I'm not super familiar with this part of the code. For example I'm not sure why `add_filters` resets the `@strainer` instance variable and if `add_global_filters` should do the same.

@dylanahsmith could you take a look?

lib/liquid/context.rb
```diff
@@ -47,6 +48,22 @@ def add_filters(filters)
       @strainer = nil
     end
 
+    def add_global_filters(global_filters)
+      gf = [global_filters].flatten.compact
```
dylanahsmith: There is no need to have multiple procs get called.  The proc that gets passed in can just call multiple methods.

lib/liquid/template.rb
dylanahsmith: I don't think `:global_filters` is the best name, since `Liquid::Strainer.global_filter` has a similar name but completely different meaning.  It sounds like it could be adding these filters to that list of global filters or it could sound like this list overrides the list of global filters.  What about `:auto_filter` as the option name?

pushrax: Maybe `after_filter`?

fw42: `default_filters`?

dylanahsmith: `default_filters` sounds like it would only be used if there aren't any other filters for that variable.

I considered something like `after_filter` but was concerned about the inverse, it not applying when there aren't any filters. It also sound like it could be applied after every filter.  It doesn't have the advantage of making it clear that it is applied after the other filters rather, which is ambiguous with `auto_filter`.

pushrax: `before_output`?

dylanahsmith:
> I'm not super familiar with this part of the code. For example I'm not sure why `add_filters` resets the `@strainer` instance variable and if `add_global_filters` should do the same.

That's not necessary, because these aren't actual filters, it is just a proc that gets called for variables after all filters have been applied to it.

lib/liquid/template.rb
boobooninja: I agree on the name conflict @dylanahsmith. Actually I don't really understand why the current Liquid::Strainer.global_filter is called what it called. I believe these are custom filters that have not been registered:
Liquid::Template.register_filter(TextFilter)
but rather just passed in at render time:
@template.render({}, :filters => [TextFilter])

Just a suggestion, but maybe they should be called custom_filters or registered_filters?

The reason I went with the 'global' word was to capture the idea that this filter was being applied to all variables in the template.

PR #607, opened 2015-06-26
Add very simple spaceless tag

### LINKED ISSUE #216: Whitespace issues
We need to do something about the amount of generated whitespace in rendered Liquid templates. We have several issues and pull requests about that right now (#215, #214, #194, #171, #162) and no suitable fix.

I want to open this for discussion because it's not only annoying for third-party users but also considerably slowing down a lot of Shopify storefronts because of megabytes of whitespace resulting from big nested loops.

A few ideas have shown up:
* Add Erb like tags (this slows down parsing a lot, as can be seen in #214).
* Automatically collapse all white-space (this will break stuff where whitespace is intended, email templates for instance).
* Create a filter for collapsing white space which can manually be wrapped around loops (not sure if this will make things much faster, but at least the output will be smaller - on the other hand, it has to be done by hand for each template).
* Automatically collapse whitespace in loops if (and only if) the rendered output contains only whitespace (this will catch some of our worst-cases, but won't fix the problem entirely).
* ...

Comments?

@boourns @hornairs @csfrancis @dylanahsmith @camilo @arthurnn
cc @jcrawford @nickpearson @avalanche123 @ragalie

pushrax:
Based on Django's equivalent tag. https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#spaceless

Helps address #216

### DIFF
--- a/lib/liquid/tags/spaceless.rb
+++ b/lib/liquid/tags/spaceless.rb
@@ -1,0 +1,24 @@
+module Liquid
+  # The spaceless tag strips whitespace between HTML tags using a simple regex.
+  #
+  # == Usage:
+  #    {% spaceless %}
+  #      <h1>
+  #        Hello
+  #      </h1>
+  #    {% endspaceless %}
+  #
+  class Spaceless < Block
+    HTML_STRIP_SPACE_REGEXP = />\s+</
+
+    def render(context)
+      self.class.strip_html_whitespace(super)
+    end
+
+    def self.strip_html_whitespace(input)
+      input.gsub(HTML_STRIP_SPACE_REGEXP, '><')
+    end
+  end
+
+  Template.register_tag('spaceless'.freeze, Spaceless)
+end
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -105,6 +105,11 @@ module Liquid
       input.to_s.gsub(/\r?\n/, ''.freeze)
     end
 
+    # Remove whitespace between HTML tags
+    def strip_html_whitespace(input)
+      Spaceless.strip_html_whitespace(input.to_s)
+    end
+
     # Join elements of the array with certain character between them
     def join(input, glue = ' '.freeze)
       InputIterator.new(input).join(glue)
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -417,4 +417,9 @@ class StandardFiltersTest < Minitest::Test
   def test_cannot_access_private_methods
     assert_template_result('a', "{{ 'a' | to_number }}")
   end
+
+  def test_strip_html_whitespace
+    html = "<a>\n\n   <b>  \n  <c>a</c> \r\n   </b>   </a>"
+    assert_template_result('<a><b><c>a</c></b></a>', "{{ a | strip_html_whitespace }}", { 'a' => html })
+  end
 end # StandardFiltersTest
--- a/test/integration/tags/standard_tag_test.rb
+++ b/test/integration/tags/standard_tag_test.rb
@@ -293,4 +293,9 @@ class StandardTagTest < Minitest::Test
   def test_multiline_tag
     assert_template_result '0 1 2 3', "0{%\nfor i in (1..3)\n%} {{\ni\n}}{%\nendfor\n%}"
   end
+
+  def test_spaceless
+    html = "<a>\n\n   <b>  \n  <c>a</c> \r\n   </b>   </a>"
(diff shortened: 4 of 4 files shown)


fw42:
This seems super naive and error-prone to me :-/

`<a title="Foo < bar">` will break this, for example... I know that's not strictly valid HTML, but that doesn't mean people don't use it.

lib/liquid/tags/spaceless.rb
```diff
@@ -0,0 +1,24 @@
...
+  #
+  # == Usage:
+  #    {% spaceless %}
+  #      <h1>
+  #        Hello
+  #      </h1>
+  #    {% endspaceless %}
+  #
+  class Spaceless < Block
+    HTML_STRIP_SPACE_REGEXP = />\s+</
```
fw42: Shouldn't this be a multiline regexp?

fw42: oh I guess `\s+` matches `\n`

pushrax:
It actually won't, since it only removes space between `>` and a following `<`. So, it will act weirdly on `<a titlye="Foo > < Bar"` but I don't think it will break anything.

pushrax:
Note: I completely stole their regex here: https://github.com/django/django/blob/47fcbe506c04019a12e16221843e25a52249b1ab/django/utils/html.py#L207

lib/liquid/tags/spaceless.rb
parkr: It only works in html. Any worries there?

pushrax: Is there something else where this would be useful?

nickpearson:
Should this tag take a parameter to change the behavior to collapse consecutive whitespace characters into a single space rather than removing it altogether? I can see this causing confusion for areas where a space should be retained, such as:

```
{% spaceless %}
  <h3>
    <b>Click</b> <a href="#">here</a>
  </h3>
{% endspaceless %}
```

which would render as `<h3><b>Click</b><a href="#">here</a></h3>`, visible as:

### Click[here](#)

One could argue `Spaceless` should be just that, space-less, but I thought I'd bring this up since I ran into this when implementing my own HTML compacting. (I solved it by handling whitespace differently for block-level and inline tags, though that's a lot more complex and isn't perfect since a block-level tag can be rendered as an inline element via CSS.)

Maybe more intelligent whitespace handling than this should exist in a different tag in the future and `Spaceless` should be left as simple as it is.

pushrax:
I think for now it should be simple, with the clearly defined rule that space is always/only removed between tags. Any extra space you put inside a tag is kept. It's easy to write your HTML with this in mind.

seangeng:
Would love to see this implemented!  Compressing our HTML in shopify templates is something I've been eagerly awaiting :)

penibelst:
I [compress HTML with Liquid](http://jch.penibelst.de/) since one year. The proposed `spaceless` tag is something I could not use for any compression step. I think the only safe use case are elements inside the `head`.

In my solution users pass an array of elements to clip whitespace around them. Like @nickpearson says, the user must decide which elements are block-level. The compression algorithm is dead simple:

1. Take an element `e`.
2. Remove whitespace before `<e`.
3. Remove whitespace after `</e>`.

That’s why I propose to expand the new tag with an **optional list of elements:**

```html
{% spaceless "p", "div" %}
<html></html>
{% endspaceless %}
```

And as filter:

```html
{{ "<html></html>" | strip_html_whitespace: "p", "div" }}
```

It’s easy to implement and makes the new solution much more powerful. Now people could put a *whole* page inside the `spaceless` tag.

penibelst:
An implementation with optional elements:

```ruby
def spaceless(input, elements = nil)
  # D-J-A-N-G-O. The D is silent.
  return input.gsub(/>\s+</, "><") unless elements

  elements.each do |element|
    # before
    input.gsub!(/\s+<#{element}/, "<#{element}")
    # after
    input.gsub!(/<\/#{element}>\s+/, "</#{element}>")
  end
  input
end

require "minitest/autorun"

class TestSpaceless < Minitest::Test
  def test_no_elements
    assert_spaceless "<a><b><c>a</c></b></a>", "<a>\n\n   <b>  \n  <c>a</c> \r\n   </b>   </a>"
  end

  def test_elements
    assert_spaceless "<a> b  <b>c</b>  </a><c>a</c>a", "<a> b  <b>c</b>  </a> <c>a</c>  a", %w(a, c)
  end

  private

  def assert_spaceless(expected, input, elements = nil)
    assert_equal expected, spaceless(input, elements)
  end
end
```

PR #608, opened 2015-07-02
Add Liquid::Tag#tag_name

gauravmc:
This has been needed a few times in Shopify, which ends up in `instance_variable_get` shittiness. Added to `attr_reader` list.

@fw42 @pushrax cc @quelledanielle


quelledanielle:
:+1: would be super useful right now

pushrax:
Sure, we discussed this already so I'm going to go ahead and merge.

PR #610, opened 2015-07-04
add global_filters (v3)

boobooninja:
add global filters using a proc
only add one proc and not an array
add tests to make sure global_filters are applied after native filters

With regard to pull request #588 and #605

Regarding my comment in #605 pertaining to naming:
I propose we use names like 'native_filters' and 'custom_filters' to represent the current two types of filters. Those that are native to Liquid and those that are custom developed by the user and then registered with Liquid or supplied to render via options.
This would then allow us to use the 'global_filters' name to accurately represent the filters applied to all variables in the template. I understand we may want to modify this to something like 'appended_global_filters' to reflect the fact they are applied after the explicit filters, or we note this in documentation.

p.s. I apologize for the multiple pull requests. I am trying to learn the best way to amend a commit related to a pull request as well as rebase that work on top of all the changes made to the upstream. Would love any info you can provide for best practices.

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -13,7 +13,7 @@ module Liquid
   #   context['bob']  #=> nil  class Context
   class Context
     attr_reader :scopes, :errors, :registers, :environments, :resource_limits
-    attr_accessor :exception_handler, :template_name, :partial
+    attr_accessor :exception_handler, :template_name, :partial, :global_filter
 
     def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_errors = false, resource_limits = nil)
       @environments     = [environments].flatten
@@ -32,6 +32,7 @@ module Liquid
 
       @interrupts = []
       @filters = []
+      @global_filter = nil
     end
 
     def warnings
@@ -50,6 +51,10 @@ module Liquid
       filters = [filters].flatten.compact
       @filters += filters
       @strainer = nil
+    end
+
+    def apply_global_filter(obj)
+      global_filter.nil? ? obj : global_filter.call(obj)
     end
 
     # are there any not handled interrupts?
--- a/test/integration/template_test.rb
+++ b/test/integration/template_test.rb
@@ -211,4 +211,18 @@ class TemplateTest < Minitest::Test
     end
     assert exception.is_a?(Liquid::ZeroDivisionError)
   end
+
+  def test_global_filter_option_on_render
+    global_filter_proc = ->(output) { "#{output} filtered" }
+    rendered_template = Template.parse("{{name}}").render({ "name" => "bob" }, global_filter: global_filter_proc)
+
+    assert_equal 'bob filtered', rendered_template
+  end
+
+  def test_global_filter_option_when_native_filters_exist
+    global_filter_proc = ->(output) { "#{output} filtered" }
+    rendered_template = Template.parse("{{name | upcase}}").render({ "name" => "bob" }, global_filter: global_filter_proc)
+
+    assert_equal 'BOB filtered', rendered_template
+  end
 end
--- a/test/unit/context_unit_test.rb
+++ b/test/unit/context_unit_test.rb
@@ -466,4 +466,18 @@ class ContextUnitTest < Minitest::Test
     assert contx
     assert_nil contx['poutine']
   end
+
+  def test_apply_global_filter
+    global_filter_proc = ->(output) { "#{output} filtered" }
+
+    context = Context.new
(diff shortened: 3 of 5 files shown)


lib/liquid/context.rb
```diff
@@ -47,6 +48,14 @@ def add_filters(filters)
       @strainer = nil
     end
 
+    def add_global_filters(global_filters)
+      @global_filters = global_filters if global_filters.respond_to?(:call)
```
fw42: same question as in the other PR (sorry if I missed the answer): why would it not respond to `:call`? and shouldn't we error in that case?

fw42:
@dylanahsmith ping

lib/liquid/context.rb
pushrax: agree

lib/liquid/context.rb
```diff
@@ -47,6 +48,14 @@ def add_filters(filters)
       @strainer = nil
     end
 
+    def add_global_filters(global_filters)
```
pushrax: Perhaps this should be singular now (`global_filter`)

dylanahsmith: Also, `add` implies that you are adding to a list.  An attribute writer like `attr_writer :global_filter` or `def global_filter=(global_filter)` would be more appropriate.

lib/liquid/context.rb
dylanahsmith: Agreed.  You should either raise an `::ArgumentError` if the argument doesn't respond to `:call` or remove the check entirely.

lib/liquid/context.rb
pushrax: Simply `attr_accessible :global_filter` is probably the way to go here.

pushrax:
After addressing the comments, can you rebase off master (to remove the class length cop) and then fix the rubocop offenses? You can find them in the test output.

Thanks for being so patient!

boobooninja:
@pushrax @fw42 @dylanahsmith I made changes and pushed code again to this PR. Wasn't sure if you guys are notified of such actions.

test/unit/context_unit_test.rb
```diff
@@ -466,4 +466,46 @@ def test_context_initialization_with_a_proc_in_environment
...
+
+    context = Context.new
+    context.global_filter = global_filter_proc
+    assert_equal 'hi filtered', context.apply_global_filter('hi')
+
+    context = Context.new
+    assert_equal 'hi', context.apply_global_filter('hi')
+  end
+
+  def test_add_global_filter_when_it_contains_several_actions
```
dylanahsmith: I don't think this test is necessary anymore, since the code above tests the same code path in the code under test.  Doing more in the global filter proc doesn't test anything new about liquid.

test/unit/context_unit_test.rb
```diff
@@ -466,4 +466,46 @@ def test_context_initialization_with_a_proc_in_environment
     assert contx
     assert_nil contx['poutine']
   end
+
+  def test_add_global_filter
```
dylanahsmith: This is just a combination of what test_apply_global_filter and test_apply_global_filter_when_no_global_filter_exist does, so it can also be removed.

test/integration/template_test.rb
```diff
@@ -211,4 +211,19 @@ def test_exception_handler_does_reraise_if_it_returns_true
     end
     assert exception.is_a?(Liquid::ZeroDivisionError)
   end
+
+  def test_global_filter_option_on_render
```
dylanahsmith: These tests you added to TemplateTest are almost the same as the ones you added to VariableTest.  I don't know which is the best file to add the tests, but keep the tests in one of them and remove the additional tests in the other.

dylanahsmith:
The code looks go to me other than the redundant tests.  It looks like you will also need to rebase on master to resolve merge conflicts.

PositiveAlex:
Hey, @boobooninja, how's it going? Could you please tell how can I help you with this task and/or when we could try to use this feature?

boobooninja:
@dylanahsmith Hopefully I addressed the latest comments and all looks good. I know @PositiveAlex is excited to use this feature. :)

dylanahsmith:
LGTM

PositiveAlex:
So, what is the status of this feature, @boobooninja ?

pushrax:
@PositiveAlex: I just merged it into Liquid master. It'll be present in the next major release of Liquid (4.0), but since that will take some time you can use it now by specifying a git ref of Liquid in your Gemfile.

PR #611, opened 2015-07-04
Remove standard exception rescue in escape filter.

### LINKED ISSUE #521: Catch-all rescue in `escape` filter might cause issues?
In the StandardFilter library: https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb#L36-L38

The `escape` filter has had a nasty catchall rescue from the first import, but this likely breaks things (and could let unescaped code slip into the template) and is causing me issues.

What specific exception was supposed to be rescued? Happy to submit a PR to fix this if I can get some background.

dylanahsmith:
@fw42 & @pushrax for review

Fixes #521

If there are exceptions happening, then they will likely result in a security vulnerability from returning the unescaped input.  Now that we report non-Liquid::Error exceptions in Shopify, we should be able to ship this without silently breaking things.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -33,7 +33,7 @@ module Liquid
     end
 
     def escape(input)
-      CGI.escapeHTML(input).untaint rescue input
+      CGI.escapeHTML(input).untaint
     end
     alias_method :h, :escape
 


fw42:
:+1:

pushrax:
Definitely :+1:

PR #612, opened 2015-07-05
Use node to refer to objects from the nodelist rather than token.

dylanahsmith:
## Problem

The BlockBody code and profiling code was referring to nodes in the nodelist as tokens, which is inconsistent and confusing.

## Solution

The variables and methods appeared to be implementation details, so I went ahead and renamed token to node in the rendering code.

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -84,10 +84,10 @@ module Liquid
             break
           end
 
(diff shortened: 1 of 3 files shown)


fw42:
Makes sense to me :+1:

pushrax:
Sure :shipit:

PR #613, opened 2015-07-05
Add taint warnings to the context rather than the template.

dylanahsmith:
@eapache & @fw42 for review

## Problem

The taint checking is adding warnings to the tag objects so that it can be retrieved from Template#warnings, but that should only be done for parse warnings, since Liquid::Template#warnings won't return warnings on included templates.

Also, Liquid::Template#warnings normally only contains exception objects, but the taint checker was adding the warnings as strings.  This makes code using the warnings fragile, and also prevents us from including line number and template name context to the errors.

Additionally, using Liquid::Template#warnings means that the whole template node graph needs to be walked to get the warnings.

## Solution

Add a warnings method to the Liquid::Context, and use it as a place to add the taint warnings.  The context is re-used for included templates, so will catch taint warnings from snippets.  A TaintedError exception object is added to the warnings array rather than just an error message.

### DIFF
--- a/lib/liquid/variable.rb
+++ b/lib/liquid/variable.rb
@@ -74,7 +74,7 @@ module Liquid
       @filters.inject(context.evaluate(@name)) do |output, (filter_name, filter_args, filter_kwargs)|
         filter_args = evaluate_filter_expressions(context, filter_args, filter_kwargs)
         context.invoke(filter_name, output, *filter_args)
-      end.tap{ |obj| taint_check(obj) }
+      end.tap{ |obj| taint_check(context, obj) }
     end
 
     private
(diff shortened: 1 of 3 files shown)


lib/liquid/variable.rb
```diff
@@ -106,16 +106,20 @@ def evaluate_filter_expressions(context, filter_args, filter_kwargs)
       parsed_args
     end
 
-    def taint_check(obj)
-      if obj.tainted?
+    def taint_check(context, obj)
+      if obj.tainted? && Template.taint_mode != :lax
```
fw42: small style nitpick... I would prefer this method to `return unless obj.tainted?` and `return if Template.taint_mode == :lax` rather than this big whole-method-if-block.

fw42:
LGTM. This might require a change in Shopify as well. :+1:

eapache:
LGTM

lib/liquid/variable.rb
dylanahsmith: agreed and made the change.

dylanahsmith:
> This might require a change in Shopify as well.

Yup, I'll change it to get these warnings from the `context.warnings` rather than `template.warnings`

PR #614, opened 2015-07-05
Implement line numbers without the Liquid::Token class.

dylanahsmith:
@Shopify/liquid for review

## Problem

Using the `line_numbers: true` option causes a significant decrease in performance, since it uses a subclass of String with a `@line_number` instance variable for tokens with that option, rather than using String directly.

Benchmark on master without line numbers

```
              parse:       18.5 (±10.8%) i/s -       1103 in  59.996823s
        parse & run:       11.9 (±8.4%) i/s -        713 in  60.011475s
```

with line numbers

```
              parse:       14.1 (±7.1%) i/s -        839 in  60.023189s
        parse & run:       10.9 (±9.2%) i/s -        654 in  60.083140s
```

The Liquid::Token objects are also passed to tags as the markup argument so that tag parse errors can include the line numbers.  This means that the performance impact would likely also affect an implementation of line number support in liquid-c, since it would still have to create Liquid::Token objects to pass to tags.

## Solution

Keep track of the current line number in the Tokenizer class which can be retrieved by BlockBody before getting the next token.

Instead of passing a Liquid::Token markup argument to the tag class, I chose to use the `options` argument to pass the line number.  Since the options argument is keeping track of parse state as well as options, I refactored it using a new ParseContext class, which implements `ParseContext#[]` for backwards compatibility.  This also allows us to append parse warnings to a single array in order to avoid having to walk the node tree afterwards to construct the list of parse warnings.

The performance without line number support is comparable to before:

```
              parse:       18.4 (±5.4%) i/s -       1096 in  60.034994s
        parse & run:       11.9 (±8.4%) i/s -        711 in  59.996768s
```

but the performance with line number support is much better

```
              parse:       17.6 (±5.7%) i/s -       1048 in  60.025864s
        parse & run:       11.5 (±8.7%) i/s -        690 in  60.010647s
```


lib/liquid/context.rb
```diff
@@ -62,10 +63,10 @@ def pop_interrupt
       @interrupts.pop
     end
 
-    def handle_error(e, token = nil)
+    def handle_error(e, line_number = nil)
       if e.is_a?(Liquid::Error)
-        e.template_name = template_name
-        e.set_line_number_from_token(token)
+        e.template_name ||= template_name
```
fw42: hm why did this change from `=` to `||=`? next line same question

fw42:
@trishume would you mind reviewing this as well?

lib/liquid/block_body.rb
```diff
@@ -12,44 +12,37 @@ def initialize
...
+            @blank = false
+          else
+            @nodelist << token
+            @blank &&= !!(token =~ /\A\s*\z/)
           end
-        rescue SyntaxError => e
-          e.set_line_number_from_token(token)
-          raise
         end
+        parse_context.line_number = tokenizer.line_number
```
trishume: Is this redundant with the one at the start of the method?

dylanahsmith: the line number needs to be updated in the pasre context before each token, and the assignment at the start of the methid is outside of the while loop.  I tried using loop and break, but object profiling showed that break makes an object allocation.

lib/liquid/parse_context.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    def initialize(options = {})
+      @template_options = options ? options.dup : {}
+      @locale = @template_options[:locale] ||= I18n.new
+      @warnings = []
+      self.partial = false
+    end
+
+    def [](option_key)
+      @options[option_key]
```
trishume: `@template_options[option_key]`?

lib/liquid/parse_context.rb
```diff
@@ -0,0 +1,39 @@
+module Liquid
+  class ParseContext
+    attr_accessor :partial, :locale, :line_number
+    attr_reader :warnings, :error_mode
+
+    def initialize(options = {})
+      @template_options = options ? options.dup : {}
```
trishume: Do you also want to set `@options` here so that it always exists even if `partial=` hasn't been called?

lib/liquid/parse_context.rb
```diff
@@ -0,0 +1,39 @@
...
+      @locale = @template_options[:locale] ||= I18n.new
+      @warnings = []
+      self.partial = false
```
trishume: The whole thing that setting this is what causes `@options` to be initialized was confusing to me when reading this code. Is there an easier to read way of doing this?

lib/liquid/parse_context.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def partial_options
+      @partial_options ||= begin
+        dont_pass = @template_options[:include_options_blacklist]
+        if dont_pass == true
+          { locale: locale }
+        elsif dont_pass.is_a?(Array)
+          opts = @template_options.dup
+          dont_pass.each { |o| opts.delete(o) }
```
trishume: I would rewrite this using `Hash#reject`: `opts = @template_options.reject {|k,v| dont_pass.include?(k)}`

lib/liquid/parse_context.rb
```diff
@@ -0,0 +1,39 @@
+module Liquid
+  class ParseContext
```
trishume: I'm tempted to make this delegate `method_missing` to the `@options` hash so that it maintains full API compatibility with old tags. But that is hacky and I'm not sure this will actually break any tags given that it does have `[]`.

tobi: definitely don't use method_missing


- tobi
CEO Shopify

trishume:
Looks good, great job, lots of bad stuff was cleaned up. :+1:

lib/liquid/context.rb
dylanahsmith: because that is what set_line_number_from_token was doing so that the line number isn't replaced when an exception is re-raised multiple times as was happening in BlockBody

lib/liquid/block_body.rb
pushrax: Maybe line_number should reference the last parsed token instead of the next?

lib/liquid/parse_context.rb
pushrax: We don't need to worry too much about API compatibility - breaking changes are already present in Liquid 4.

lib/liquid/parse_context.rb
pushrax: I think this is the most DRY way, it's an unfortunate syntax but very Ruby.

lib/liquid/parse_context.rb
pushrax: Unfortunately that increases the time complexity, not sure if it matters here though.

lib/liquid/parser_switching.rb
```diff
@@ -1,16 +1,14 @@
...
       when :strict then strict_parse_with_error_context(markup)
       when :lax    then lax_parse(markup)
       when :warn
         begin
           return strict_parse_with_error_context(markup)
         rescue SyntaxError => e
-          e.set_line_number_from_token(markup)
-          @warnings ||= []
-          @warnings << e
+          @options.warnings << e
```
pushrax: It's probably worth renaming options to avoid future confusion.

pushrax:
This is definitely the right way to do line numbers, I'm really glad you were able to get rid of the Token class.

lib/liquid/block_body.rb
dylanahsmith: > Maybe line_number should reference the last parsed token instead of the next?

Do you mean in the tokenizer?  That seems like it would be confusing, since it has already consumed that token, so line_number wouldn't reflect the tokenizer's current position.

lib/liquid/parse_context.rb
trishume: True but in practice both the hash and `dont_pass` should be very small (2-3 items) so I doubt it will even affect the benchmarks at all.

lib/liquid/block_body.rb
pushrax: Sure, the current way is fine with me. I was just wondering if it might make the common use-case easier.

lib/liquid/parse_context.rb
dylanahsmith: this is done once per parse, so I wasn't worried about time complexity.  I changed it to use reject.

lib/liquid/parser_switching.rb
dylanahsmith: I kept the `@options` instance variable because liquid-c refers to it directly in a [Liquid::Variable#lax_parse monkey patch](https://github.com/Shopify/liquid-c/blob/35e9aee48d639ae1d3ac9ba77616aca9800eab7d/lib/liquid/c.rb#L46).  However, I tried to get rid of direct usage of `@options` in liquid for now, and made sure Tag and Variable both have methods to use instead of `@options` so that liquid-c can be decoupled from a specific instance name for options.

PR #616, opened 2015-07-08
Handle non-int range lookup arguments

pushrax:
Also moved out the `to_*` methods into `Utils`.

@dylanahsmith @fw42


dylanahsmith:
:+1:

PR #617, opened 2015-07-09
Rename options instance variable in Variable and Tag.

dylanahsmith:
@pushrax for review

Addresses pull request comment https://github.com/Shopify/liquid/pull/614#discussion_r33938850.

I updated liquid-c in https://github.com/Shopify/liquid-c/pull/24 to not use the `@options` instance variable directly.


pushrax:
:+1:

PR #621, opened 2015-07-09
Add a lax warning mode to see where we can make the lax parser stricter.

dylanahsmith:
@fw42 & @pushrax for review

## Problem

There are a lot of implicit liquid quirks due to not using anchored regexes, including things like unterminated quotes being used as a filter separator.  However, I suspect there are some quirks that actually don't happen in production or are rare enough to fix by hand.

## Solution

I decided we should log these edge cases when the happen in the lax parser, using warning messages that are specific enough to search the logs for, along with the relevant part of the markup.

We already have a warning mode, but the strict warnings aren't specific enough for this purpose, so I added a `:lax_warn` error_mode.  We could get rid of this error mode after making use of these warnings.

I anchored regexes that were used for a single match, and non-greedy captures to capture text at the start of regex scans.  I also tried to make sure `nil` is captured rather than an empty string for these edge cases, so avoid extra string allocations.

## Benchmark

on master

```
              parse:     18.673  (±16.1%) i/s -      1.075k
        parse & run:     12.302  (±16.3%) i/s -    720.000
```

on this branch

```
              parse:     18.792  (±16.0%) i/s -      1.077k
        parse & run:     12.453  (±16.1%) i/s -    729.000
```

### DIFF
--- a/lib/liquid/variable.rb
+++ b/lib/liquid/variable.rb
@@ -10,7 +10,16 @@ module Liquid
   #   {{ user | link }}
   #
   class Variable
-    FilterParser = /\s*(?:#{FilterSeparator}|(['"\|]+?))\s*((?:\s+|#{QuotedFragment}|#{ArgumentSeparator})+)/o
+    capture_ignored_variable_prefix = /([\s,\|'"]+?)??/
+    capture_expression = /(#{QuotedFragment})/o
+    capture_ignored_filter_prefix = /([^\|]+?)??/
+    capture_filters = /(#{FilterSeparator}.*)/o
+    VariableSyntax = /\A\s*#{capture_ignored_variable_prefix}\s*#{capture_expression}\s*(?:#{capture_ignored_filter_prefix}\s*#{capture_filters})?\z/om
+
+    capture_lax_seperator = /(['"\|]+?)/
+    capture_filter = /((?:\s+|#{QuotedFragment}|#{ArgumentSeparator})+)/o
+    FilterParser = /\s*(?:#{FilterSeparator}|#{capture_lax_seperator})\s*#{capture_filter}/o
+
     attr_accessor :filters, :name, :line_number
     attr_reader :parse_context
     alias_method :options, :parse_context
@@ -35,7 +44,7 @@ module Liquid
 
     def lax_parse(markup)
       @filters = []
-      return unless markup =~ /\A\s*([\s,\|'"]+?)??\s*(#{QuotedFragment})\s*(?:([^\|]+?)??\s*(#{FilterSeparator}.*))?\s*\z/om
+      return unless markup =~ VariableSyntax
 
       add_syntax_warning("variable prefixed with ignored characters: #{$1.inspect}") if $1
       name_markup = $2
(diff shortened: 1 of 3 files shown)


lib/liquid/variable.rb
```diff
@@ -10,7 +10,16 @@ module Liquid
...
-    FilterParser = /(?:\s+|#{QuotedFragment}|#{ArgumentSeparator})+/o
+    capture_ignored_variable_prefix = /([\s,\|'"]+?)??/
+    capture_expression = /(#{QuotedFragment})/o
+    capture_ignored_filter_prefix = /([^\|]+?)??/
+    capture_filters = /(#{FilterSeparator}.*)/o
+    VariableSyntax = /\A\s*#{capture_ignored_variable_prefix}\s*#{capture_expression}\s*(?:#{capture_ignored_filter_prefix}\s*#{capture_filters})?\z/om
+
+    capture_lax_seperator = /(['"\|]+?)/
+    capture_filter = /((?:\s+|#{QuotedFragment}|#{ArgumentSeparator})+)/o
+    FilterParser = /\s*(?:#{FilterSeparator}|#{capture_lax_seperator})\s*#{capture_filter}/o
```
pushrax: :sob:

lib/liquid/variable.rb
```diff
@@ -10,7 +10,16 @@ module Liquid
...
   #
   class Variable
-    FilterParser = /(?:\s+|#{QuotedFragment}|#{ArgumentSeparator})+/o
+    capture_ignored_variable_prefix = /([\s,\|'"]+?)??/
+    capture_expression = /(#{QuotedFragment})/o
+    capture_ignored_filter_prefix = /([^\|]+?)??/
+    capture_filters = /(#{FilterSeparator}.*)/o
+    VariableSyntax = /\A\s*#{capture_ignored_variable_prefix}\s*#{capture_expression}\s*(?:#{capture_ignored_filter_prefix}\s*#{capture_filters})?\z/om
+
+    capture_lax_seperator = /(['"\|]+?)/
```
pushrax: seperator -> separator

lib/liquid/variable.rb
dylanahsmith: save some tears for the test

```ruby
    template = Liquid::Template.parse("{{ test | '': 'prepend ' }}", error_mode: :lax_warn)
    assert_equal "prepend wat", template.render!('test' => 'wat')
```

pushrax:
It's too late in the day for me to parse this, will take a look tomorrow morning. Definitely :+1: to the idea though.

lib/liquid/variable.rb
dylanahsmith: changed everywhere that I made that typo

fw42:
also @trishume if you have a minute

pushrax:
This looks correct, but makes me slightly sad. I wonder if there's a way to leverage the formatter to do this.

pushrax:
That said, :+1: all the regexes look right to me.

trishume:
Could this same purpose not be served by analyzing templates outside of production? Talk to @pushrax about the Shopify template comparator that he used.

pushrax:
That is a good point actually – the AST comparisons could be made smart enough to detect exactly what the issue is. Right now it's just a dumb diff though.

dylanahsmith:
Yes, these changes could be used in or out of production by just parsing with `error_mode: :lax_warn` and picking up the errors.  I'm not sure how long it takes to parse all of the templates outside of production, but it would be more complete.  However, we are already parsing these templates in production, so just adding logging is a quick way of roughly assessing the impact of a change to how these captured characters will be handled.

trishume:
Everything looks good :+1:
Still worried a little bit since the logic in that function is notoriously hard to mess with properly while maintaining equivalence.

I'm in favour of reverting this after we're done with it (or simply never merging it and running a comparator job on the branch) because it makes the logic harder to understand and makes it almost look like the regexes are parsing properly when it should be clear to anyone reading that they aren't.

Also IIRC @pushrax's method for running a comparator job on prod templates isn't that difficult and can be run on a laptop in an afternoon. I recommend doing that, you can be more sure that your data is good than through logging and aggregation. Also if you get it to work it would make for good groundwork for future comparisons of liquid versions.

dylanahsmith:
ok, I'm going to close this and do it outside of production, but I'll use the code here to log the warnings that I'm interested in.  Once I find changes that I think I can introduce, I can do another pass two compare the AST of the proposed changes with liquid master.

PR #623, opened 2015-07-13
Allow use of curlies in a properly quoted string

knu:
Currently Liquid does not even allow a closing curly brace to be put inside of a variable, but you sometimes need to deal with the character.

### DIFF
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -35,7 +35,7 @@ module Liquid
   QuotedFragment              = /#{QuotedString}|(?:[^\s,\|'"]|#{QuotedString})+/o
   TagAttributes               = /(\w+)\s*\:\s*(#{QuotedFragment})/o
   AnyStartingTag              = /\{\{|\{\%/
-  PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/om
+  PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}(?:(?:[^'"{}]+|#{QuotedString})*?|.*?)#{VariableIncompleteEnd}/om
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/om
   VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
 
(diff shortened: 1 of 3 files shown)


fw42:
Hm I think we tried to fix this before and we didn't merge it because it negatively impacted performance by quite a bit. @trishume might remember.

fw42:
Does this bug exist in the strict parser as well?

nickpearson:
It's not ideal, but you can work around this with a `capture`. Using your test case (`{{funk | replace: "}", '}}' }}`) as an example:

```
{% capture match %}}{% endcapture %}
{% capture replacement %}}}{% endcapture %}
{{ funk | replace: match, replacement }}
```

```ruby
# the above Liquid is in the `text` variable
Liquid::Template.parse(text).render({ 'funk' => '{x}' })
#=> {x}}
```

trishume:
See https://github.com/Shopify/liquid/pull/170

I don't think the strict parser would help since this is an issue with the tokenizer, which comes before the strict parser.

It is possible that liquid-c could fix this without a performance hit but that would make it act differently than base liquid.

pushrax:
The problem is that we separate the tokenizer and parser, which is necessary for the regex approach but not for the strict approach. I don't think we're at the state where we could merge the two though :(

dylanahsmith:
We should try implementing this in liquid-c, since performance was the main concern with #170.  #170 also seems to have implemented support for `\` escaping, but I don't know if that is safe to add without breaking things.

lib/liquid.rb
```diff
@@ -35,7 +35,7 @@ module Liquid
   QuotedFragment              = /#{QuotedString}|(?:[^\s,\|'"]|#{QuotedString})+/o
   TagAttributes               = /(\w+)\s*\:\s*(#{QuotedFragment})/o
   AnyStartingTag              = /\{\{|\{\%/
-  PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/om
+  PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}(?:(?:[^'"{}]+|#{QuotedString})*?|.*?)#{VariableIncompleteEnd}/om
```
dylanahsmith: Why is `[^'"{}]+|` needed?

dylanahsmith:
This is missing integration tests, which are preferred since they get used with liquid-c.

lib/liquid.rb
knu: You mean it could be `(?:#{QuotedString})*?|.*?)`?  No, that would only save a quoted string at the beginning of a variable block (which normally begins with a variable name) and the following `.*?` would stop at the next `}` that might be in the next quoted string.

I included `{}` in the character class because it looked like the tokenizer has to be generous with curly brace mismatch for the "lax" parser to work.

knu:
Thanks for taking a look, guys.  I actually didn't know a fix for this had once been rejected due to performance issues, but if there's any kind of benchmark index/threshold Liquid must surpass, I'm willing to improve the implementation.

devdreamsolution (approved):

knu:
A new parser/tokenizer was introduced in 5.6.1 and this does not apply any more.

Superseded by https://github.com/Shopify/liquid/pull/2063

PR #624, opened 2015-07-14
Allow variable and tag end characters to be quoted.

dylanahsmith:
@pushrax and @trishume for review
cc @knu

Closes #623
Depends on https://github.com/Shopify/liquid-c/pull/27

## Problem

Add support for quoting string string in variables and tags to contain end variable and tag characters.

## Solution

When encountering a quote character inside a variable or tag, the tokenizer will include everything up to the end quote character (if one is present) even if it includes the tag/variable end characters.

I tried to keep the tag contents regex as simple as possible.  I also added integration tests that fail without using the liquid-c without the same support for this case.

### DIFF
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -35,7 +35,8 @@ module Liquid
   QuotedFragment              = /#{QuotedString}|(?:[^\s,\|'"]|#{QuotedString})+/o
   TagAttributes               = /(\w+)\s*\:\s*(#{QuotedFragment})/o
   AnyStartingTag              = /\{\{|\{\%/
-  PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/om
+  tag_contents                = /(?:#{QuotedString}|.)*?/m
+  PartialTemplateParser       = /#{TagStart}#{tag_contents}#{TagEnd}|#{VariableStart}#{tag_contents}#{VariableIncompleteEnd}/om
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/om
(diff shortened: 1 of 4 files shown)


fw42:
Can you benchmark the performance impact of this?

lib/liquid.rb
```diff
@@ -35,7 +35,8 @@ module Liquid
   QuotedFragment              = /#{QuotedString}|(?:[^\s,\|'"]|#{QuotedString})+/o
   TagAttributes               = /(\w+)\s*\:\s*(#{QuotedFragment})/o
   AnyStartingTag              = /\{\{|\{\%/
-  PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/om
+  tag_contents                = /(?:#{QuotedString}|.)*?/m
```
fw42: hm this being a local variable and not a constant looks kinda odd...

knu: Adding `[^"'}]+|` before QuotedString would reduce the number of backtracks.

dylanahsmith: there should only be backtracking for unterminated quotes the way it is, which is an edge case rather than a common case.

dylanahsmith: also, the local variable is because I don't want it to be used outside of liquid

trishume:
:+1:
Again would be great if we could run a breakage test job.

dylanahsmith:
We need to make it easier to test the impact of possibly breaking changes on existing templates, since that is what is blocking changes like this from happening.  This is the problem having a language that isn't strict, it doesn't leave room for extension without possibly breaking existing code.

devdreamsolution (approved):

Issue #625, opened 2015-07-17
Chained access to multidimensional hash using variables as keys


Mingan:
We're in a process of upgrading our website to use Jekyll 3. Apart from the normal issues, we've encountered some strange problems and one of them seems to be with Liquid.

In the past, we used Jekyll 2.5.3 with Liquid **2.6.1**, after update we have Jekyll 3.0.0.pre.beta8 and Liquid **3.0.3**.

Construction like 

    site.data.menu[include.menu][include.locale]

with variables `include.menu` and `include.locale` used to be legal and worked as expected. (We used this in a for loop to generate menu.)

In the version it fails with:

    Liquid syntax error: Expected end_of_string but found open_square in "{{site.data.menu[include.menu][include.locale]}}"

We managed to sidestep the problem by extracting the first level into a variable and then accessing the second one like:

    {% assign menus = site.data.menu[include.menu] %}
    {% for item in menus[include.locale] %}
         ...
    {% endfor %}

I couldn't find any mention about this functionality at all let alone about changes but since it's possible to access the hash in two steps I think it should be possible to do it in a single step as well.

fw42:
Interesting. I confirmed the problem. This indeed works with the lax parser (the default of Jekyll 2) but not the strict parser (the default of Jekyll 3).

Not sure why. It looks correct to me. @trishume @pushrax?

trishume:
Yah this is a bug in the strict parser. It only tries to parse one set of square brackets after a variable name.

This specific case could trivially be fixed by changing the `if` on this line to a `while`:
https://github.com/Shopify/liquid/blob/master/lib%2Fliquid%2Fparser.rb#L78

That might actually be a correct expression of the intended grammar but we should probably think about what the grammar should be and when we want those square brackets to make sure we allow them in all the right places.

fw42:
We should backport this fix to the 3.0 branch so that Jekyll can pick it up...

@parkr FYI

fw42:
Thanks for reporting this issue!

Mingan:
Thank you, for fixing it so amazingly fast! :heart:

PR #626, opened 2015-07-17
Fix bracket thing

fw42:
Fixes https://github.com/Shopify/liquid/issues/625.

@trishume @pushrax

### DIFF
--- a/lib/liquid/parser.rb
+++ b/lib/liquid/parser.rb
@@ -75,7 +75,7 @@ module Liquid
 
     def variable_signature
       str = consume(:id)
-      if look(:open_square)
+      while look(:open_square)
         str << consume
         str << expression
         str << consume(:close_square)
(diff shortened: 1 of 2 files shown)


trishume:
Cool, as far as I can tell this is the correct fix that matches what we would expect the grammar to be. It is only this way because the liquid grammar is simple, this wouldn't be the correct way to handle things if liquid had other postfixes on identifiers like function call brackets.

:+1:

pushrax:
:+1:

fw42:
Backported to 3-0-stable in https://github.com/Shopify/liquid/commit/8f84ddb5ce8b3296b781ee269d1f1d9ce0ea3750 and pushed 3.0.4 to Rubygems.

@parkr you might want to bump Liquid in Jekyll 3.

parkr:
> @parkr you might want to bump Liquid in Jekyll 3.

`~> 3.0, >= 3.0.4`? Is 3.0.0-3.0.3 unusable?

fw42:
ah ok ~> 3.0 should do. I thought you were locked 3.0.3 or something.

PR #627, opened 2015-07-21
[RFC] Inject Liquid template name into Ruby backtrace

fw42:
Not intended to be merged, this is just a quick and dirty idea that I think might be useful and that I would like some feedback on.

We sometimes see Ruby exceptions coming from Liquid and it's often difficult to track down which piece of Liquid code triggered it, so it would be nice to have the template name and line number as part of the Ruby backtrace. This PR does that (in a super ugly way, but it's a proof of concept).

Example:

template
```
foo

bar

{{ boom }}

bla
```

test.rb
```
require 'liquid'

class Boom
  def to_liquid
    raise 'broken'
  end
end

template = Liquid::Template.parse(File.read('template'), line_numbers: true)

output = template.render(
  { 'boom' => Boom.new },
  {
    template_name: 'template.liquid',
    exception_handler: ->(e){ true }
  }
)

puts output
```

output:
```
test.rb:5:in `to_liquid': broken (RuntimeError)
	from /Users/flo/Projects/shopify/liquid/lib/liquid/context.rb:217:in `find_variable'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/variable_lookup.rb:36:in `evaluate'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/context.rb:192:in `evaluate'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/variable.rb:76:in `render'
	from template.liquid:5
	from /Users/flo/Projects/shopify/liquid/lib/liquid/block_body.rb:90:in `render_node'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/block_body.rb:72:in `block in render'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/block_body.rb:59:in `each'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/block_body.rb:59:in `render'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/template.rb:210:in `block in render'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/template.rb:244:in `with_profiling'
	from /Users/flo/Projects/shopify/liquid/lib/liquid/template.rb:209:in `render'
	from test.rb:11:in `<main>'
```

Thoughts? How can we make this better? Is this even useful?

@Shopify/liquid @trishume

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -87,11 +87,32 @@ module Liquid
         when String
           output = result
         else
-          raise if result
+          if result
+            inject_shit_into_backtrace(e, line_number)
+            raise e
+          end
         end
       end
       errors.push(e)
       output || Liquid::Error.render(e)
+    end
+
+    def inject_shit_into_backtrace(e, line_number)
+      return unless template_name && line_number
+
+      new_backtrace = []
+
+      found = false
+      e.backtrace.each_with_index do |line, index|
+        if !found && line =~ /lib\/liquid\/block_body\.rb:\d+:in `render_node'/
+          new_backtrace << "#{template_name}:#{line_number}"
+          found = true
+        end
+
+        new_backtrace << line
+      end
+
+      e.set_backtrace(new_backtrace)
     end
 
     def invoke(method, *args)
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -190,6 +190,10 @@ module Liquid
         if options[:exception_handler]
           context.exception_handler = options[:exception_handler]
         end
+
+        if options[:template_name]
+          context.template_name = options[:template_name]
+        end
       when Module
         context.add_filters(args.pop)
       when Array


pushrax:
:+1: to the idea. Just need to be super-careful not to blow away any existing information in the backtrace by accident.

fw42:
it's not super clear where exactly in the backtrace the new line should show up and how to reliably determine that in each case... thoughts?

trishume:
TIL Ruby has a `#set_backtrace` method, when I read the PR description I was expecting something like wrapping rendering in a call to a dynamically created method which does nothing, but is named to include the template and line. This is much less of a hacky and seems to be the kind of thing `set_backtrace` was intended for.

I'm cool with this idea, especially since it only screws with things if you opt in with an exception handler, and only when an exception actually happens. The main code paths are no less clean because of this.

Another option (not sure if more or less clean) would be to wrap unexpected exceptions in a wrapper exception class (`Liquid::ErrorWrapper`?) that includes the template and line number.

This example implementation could of course use some tweaks, perhaps on method naming :stuck_out_tongue_winking_eye:

dylanahsmith:
Interesting, this is the first time I have seen set_backtrace used.

> it's not super clear where exactly in the backtrace the new line should show up and how to reliably determine that in each case... thoughts?

Does it really make sense to try to interleave the backtraces, so there are no ruby methods generated from the liquid code?  I had been thinking that the liquid backtrace could be exposed separately in exception notifications, along with other request metadata, which would be less fragile and avoid this concern.

PR #631, opened 2015-07-23
Fix a timezone test failure.

### LINKED ISSUE #630: test_date may fail in certain timezones
Hi,

As part of the effort of the Debian GNU/Linux distribution to provide reproducible binary packages, it has been noticed that the test_date test case can fail when run in a certain timezone (TZ=UTC-14), because of some arithmetic problems around midnight. See
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=788694
for the Debian report and 
https://reproducible.debian.net/rb-pkg/unstable/amd64/ruby-liquid.html
for the full build log.

I don't know exactly what would be the best way to deal with that. Is fixing deterministically the timezone during the test could be considered?

Cheers,

Cédric

dylanahsmith:
@pushrax for review

Fixes #630

The test had a hard coded timestamp and formatted date, so we should also hard code the timezone to a failure in other timezones.

### DIFF
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -263,8 +263,10 @@ class StandardFiltersTest < Minitest::Test
 
     assert_equal '', @filters.date('', "%B")
 
-    assert_equal "07/05/2006", @filters.date(1152098955, "%m/%d/%Y")
-    assert_equal "07/05/2006", @filters.date("1152098955", "%m/%d/%Y")
+    with_timezone("UTC") do
+      assert_equal "07/05/2006", @filters.date(1152098955, "%m/%d/%Y")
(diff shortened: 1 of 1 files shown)


pushrax:
:+1:

pushrax:
Let's backport this as well.

fw42:
Cool :+1:

dylanahsmith:
Backported to 3-0-stable with commit d891c2b7abbdc69afe3fd42f7249b4714365ead1

pushrax:
Might have to backport to 2.6 as well for the debian people.

dylanahsmith:
Backported to 2-6-stable with commit 435babc051a60d63c8cfa0024d9267b5b4a63203

PR #632, opened 2015-07-24
Properly rescue ::ArgumentError in the date filter

knu:
Currently Utils.to_date rescues `ArgumentError` which actually refers to `Liquid::ArgumentError`, and `::ArgumentError` raised by `Time.parse` will be leaked.

### DIFF
--- a/lib/liquid/utils.rb
+++ b/lib/liquid/utils.rb
@@ -72,7 +72,7 @@ module Liquid
       when String
         Time.parse(obj)
       end
-    rescue ArgumentError
+    rescue ::ArgumentError
       nil
     end
   end
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -420,6 +420,11 @@ class StandardFiltersTest < Minitest::Test
     assert_template_result('a', "{{ 'a' | to_number }}")
   end
 
+  def test_date_raises_nothing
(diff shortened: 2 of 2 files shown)


fw42:
Oh, good catch.

@dylanahsmith?

fw42:
we might want to actually raise a Liquid::ArgumentError instead of nil

pushrax:
Either approach is ok with me.

dylanahsmith:
> we might want to actually raise a Liquid::ArgumentError instead of nil

We actually return a `nil` if nothing matches the case statement.  Do we want an inconsistency in that way?

fw42:
Good point, it should be consistent. I think raising Liquid::ArgumentError would be better, but I'm not sure about backwards compatibility, so I would say lets leave it as is for now? yay? nay?

pushrax:
Yeah let's leave it for now and just merge as-is.

fw42:
Thank you for your contribution!

knu:
Thanks!  Please backport this to 3.0 also. (2.6 doesn't seem to have this problem)

knu:
It'd be OK if Utils.to_date raised Liquid::ArgumentError, but I think the `date` filter should rescue it and just return the original input, because filters are expected to be generous to input (data) while they can be strict about parameters given (code).

PR #636, opened 2015-07-31
Don't allow identifiers to end in a dash in strict parse mode.

dylanahsmith:
@pushrax & @trishume for review

## Problem

We were allowing identifiers to end in a dash, but I don't think this was intended.  E.g. `foo-` would be a valid identifier in strict mode.

## Solution

Make sure identifiers end with a word character.

### DIFF
--- a/lib/liquid/lexer.rb
+++ b/lib/liquid/lexer.rb
@@ -13,7 +13,7 @@ module Liquid
       '?'.freeze => :question,
       '-'.freeze => :dash
     }
-    IDENTIFIER = /[a-zA-Z_][\w-]*\??/
+    IDENTIFIER = /[a-zA-Z_](?:[\w-]*\w)?\??/
     SINGLE_STRING_LITERAL = /'[^\']*'/
     DOUBLE_STRING_LITERAL = /"[^\"]*"/
     NUMBER_LITERAL = /-?\d+(\.\d+)?/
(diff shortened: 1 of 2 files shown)


pushrax:
For some reason I thought this was intentional. Tomorrow I'm going to get the comparator set up properly and generally so we can see if this will affect production.

:+1: otherwise.

lib/liquid/lexer.rb
```diff
@@ -13,7 +13,7 @@ class Lexer
       '?'.freeze => :question,
       '-'.freeze => :dash
     }
-    IDENTIFIER = /[a-zA-Z_][\w-]*\??/
+    IDENTIFIER = /[a-zA-Z_](?:[\w-]*\w)?\??/
```
pushrax: This disallows identifiers with multiple dashes in a row (e.g. `a--b`).

dylanahsmith: Nope.  I updated an added test to cover this case as well though.

pushrax: Derp, misread the logic. :+1:

dylanahsmith:
> For some reason I thought this was intentional.

Oh, maybe it was.  It seems weird to have a identifier end with a dash, but I guess the same could be said of an ending underscore.  I guess the restrictions only matter on the start of the token to make it easier to distinguish between identifiers and numbers.

PR #637, opened 2015-08-02
Allow first and last filters to take count argument

tdg5:
It seems to me that this functionality would be useful to many Liquid users. As may be obvious, use of the count argument with dot notation is not supported.

@fw42 could I trouble you for a code review at your convenience?

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -265,8 +265,10 @@ module Liquid
     # Example:
     #    {{ product.images | first | to_img }}
     #
-    def first(array)
-      array.first if array.respond_to?(:first)
+    def first(array, count = nil)
+      return unless array.respond_to?(:first)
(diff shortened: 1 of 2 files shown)


pushrax:
LGTM.

dylanahsmith:
Note that this won't work for drops that don't take a count argument (e.g. https://github.com/Shopify/liquid/blob/27c6b8074a3c1f54179d62c97d8bef53700eb5a2/test/integration/drop_test.rb#L70), and this won't be detected until this starts getting used.  I know this is at least the case in Shopify, so we should make sure to address this there, but we should also make not of this in the CHANGELOG.

tdg5:
I added a commit making note of the change in `History.md`. I added it under the changes for `4.0.0 / not yet released / branch "master"`. Let me know if this was overzealous.

devdreamsolution (approved):

PR #645, opened 2015-08-21
Add a url_decode standard filter

larcher:
I ran into a place where I wanted a `url_decode` filter while experimenting with Huginn. 

I found issue #259, where a `url_encode` and `url_decode` filters were discussed, but it doesn't look like that ever made it into Liquid core? And there was #421, which did add the `url_encode` filter, but no `url_decode`.

So here it is, `url_decode` -- just the inverse of `url_encode`, calls `CGI::unescape` instead of `escape`.


pushrax:
Merged in #658

PR #652, opened 2015-09-25
Fix "sort" filter on empty array to return empty array

mcary:
Can we adjust the "sort" filter to return empty array?  This would allow it to be chained with other array-like filters.  Currently calling "sort" followed by "map" results in an array containing one nil, which seems very broken.

I've added a test and verified that the test fails without this fix (and passes with it).

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -116,6 +116,8 @@ module Liquid
       ary = InputIterator.new(input)
       if property.nil?
         ary.sort
+      elsif ary.empty? # The next two cases assume a non-empty array.
+        []
       elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
         ary.sort { |a, b| a[property] <=> b[property] }
       elsif ary.first.respond_to?(property)
@@ -130,6 +132,8 @@ module Liquid
 
       if property.nil?
         ary.sort { |a, b| a.casecmp(b) }
+      elsif ary.empty? # The next two cases assume a non-empty array.
+        []
       elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
         ary.sort { |a, b| a[property].casecmp(b[property]) }
       elsif ary.first.respond_to?(property)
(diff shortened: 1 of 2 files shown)


fw42:
Huh. Yeah. This sucks. Fix looks reasonable to me. @pushrax?

dylanahsmith:
:+1:

Looks like the test failures are from a new version of rubocop.  I opened #653 to take care of those.

dylanahsmith:
Please rebase on master to fix the test failures.

pushrax:
Agree and looks good. This should be fixed for `sort_natural`, `uniq`, and `compact` as well.

mcary:
Any ideas how to address the count performance issue?  I did look at size/length/empty, but I guess count is the only one in Enumerable.

Would it be too bold to assume `ary.first == nil` means empty?  Or how would you feel about a trick like:

```Ruby
   ary.first(1).empty?
```

`first` returns an array when passed a number, even if the number is 1, at least on Ruby 2.1.  And the returned array seems to be limited by the size of `ary`.

dylanahsmith:
If you are actually concerned about the performance of `.count` to check if the enumerable is empty, then we could add an `empty?` method on InputIterator.

This implementation will be constant time without the extra array allocation from using `.first(1).empty?`

```ruby
  def empty?
    @input.each { return false }
    true
  end
```

mcary:
Hrm interesting.  That `empty?` implementation does look constant-time as well, but with more lines of code.  Your call on readability vs. small constant factor of performance.

Either way, it does seem valuable to add an `empty?` method to the InputIterator class now that we're talking about calling it from several filters.

mcary:
Interesting that the `first` filter is not as flexible as the others; it requires the input to respond to `first` rather than attempting `Array(...)` as the InputIterator would.  (I noticed it when searching for occurrences of `.first`.)

mcary:
I pushed up a revised commit that addresses the feedback so far.

mcary:
Oops, and now I've rebased onto master as well.

dylanahsmith:
:+1:

PR #653, opened 2015-09-25
Fix offenses from the new version of rubocop.

dylanahsmith:
@fw42 for review

I noticed that CI is now picking up new rubocop offenses, since we don't lock the version in the Gemfile.  Should we be locking the version?

Either way, this should deal with these new offenses.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -42,7 +42,7 @@ module Liquid
     end
 
     def url_encode(input)
-      CGI.escape(input) rescue input
+      CGI.escape(input) unless input.nil?
     end
 
     def slice(input, offset, length = nil)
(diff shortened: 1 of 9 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -42,7 +42,7 @@ def escape_once(input)
     end
 
     def url_encode(input)
-      CGI.escape(input) rescue input
```
dylanahsmith: We already removed the `rescue input` for the escape filter in https://github.com/Shopify/liquid/pull/611, so we may as well do it here as well for the same reasons.

fw42:
Locking the version makes sense!

fw42:
:+1:

dylanahsmith:
Ok.  I've locked the version to the latest version of rubocop.

PR #658, opened 2015-10-21
Add a url_decode standard filter (rework #645)

pushrax:
Rebased and removed the rescue from #645. I think this is reasonable to include in core.

@dylanahsmith or @fw42


fw42:
Sure :+1:

PR #661, opened 2015-10-28
Rename before_method as liquid_method_missing

tjoyal:
The method name `before_method` does not reflect the drop invocation behaviour. It will not be called "before" another method that already exists.

At first I wrote this with support for the previous method and fallback mechanism, but that added complexity for no obvious reasons. (we could backport to an earlier version once this is merged if we feel the need to. That would also require us to publish a gem update)

### DIFF
--- a/History.md
+++ b/History.md
@@ -15,6 +15,7 @@
 * Ruby 1.9 support dropped (#491) [Justin Li]
 * Liquid::Template.file_system's read_template_file method is no longer passed the context. (#441) [James Reid-Smith]
 * Remove support for `liquid_methods`
+* Rename Drop method `before_method` as `liquid_method_missing` (#661) [Thierry Joyal]
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


fw42:
@dylanahsmith @pushrax thoughts?

dylanahsmith:
:+1:

pushrax:
I have always thought of this concept as "basically method_missing" – what do you think of a name like `liquid_method_missing` or some variant of that? The current name is definitely confusing though, so :+1: in any case.

tjoyal:
Yea, probably better to go with a better name.

`liquid_method_missing` as it is an existing concept?
`invoke_drop_method_missing` as it is explicit?
`invoke_method_missing` as it is easier to read?

dylanahsmith:
:+1: for `liquid_method_missing`

tjoyal:
PR updated and renamed

fw42:
:+1:

PR #662, opened 2015-10-28
Keep original stack trace in Liquid::ArgumentError

nickpearson:
By only reporting the original `ArgumentError`'s `message`, the original backtrace is lost, making it difficult to track down some application errors that sprout up in Liquid. (This happend to me with an error in a `to_liquid` method.)

Presently, the backtrace shows the call stack starting in `Liquid::Strainer#invoke` where `raise` is called inside the `rescue` rather than in the application code where the error actually occurs. This update passes the original backtrace along with the message so it's clear where an error is actually happening.

Ping @fw42 for review.

### DIFF
--- a/lib/liquid/strainer.rb
+++ b/lib/liquid/strainer.rb
@@ -52,7 +52,7 @@ module Liquid
         args.first
       end
     rescue ::ArgumentError => e
-      raise Liquid::ArgumentError.new(e.message)
+      raise Liquid::ArgumentError, e.message, e.backtrace
     end
   end
 end
--- a/test/unit/strainer_unit_test.rb
+++ b/test/unit/strainer_unit_test.rb
@@ -29,6 +29,18 @@ class StrainerUnitTest < Minitest::Test
     end
   end
 
+  def test_stainer_argument_error_contains_backtrace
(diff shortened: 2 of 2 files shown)


fw42:
Hm we don't really want backtraces to show up in the rendered output (they expose internal details to the user)... not 100% sure this would happen here, but I would be very careful.

dylanahsmith:
Liquid::ArgumentError already has a stack trace which has internal details (filenames and line numbers) this is just using the third argument to `raise` which can be used to set the backtrace, so `Liquid::ArgumentError` has the same backtrace as `::ArgumentError`

Seems reasonable to me.

fw42:
Good point... :+1:

fw42:
Will merge this if you make the tests pass

nickpearson:
Updated PR to handle JRuby-style error message.

I used `assert_match` to match both `(1 for 0)` and `(given 1, expected 0)`, but I can update the test to use `assert_includes` instead if that's your preference.

fw42:
Cool. Thanks Nick!

nickpearson:
Thanks!

Issue #664, opened 2015-11-06
Standard filter "escape" fails when input is nil


tanelj:
When input value passed to `escape` filter is `nil` then error has been raised.

Example:
```
Liquid::Template.parse("{{ somevar | escape }}").render
```

Error:
```
Error `ActionView::Template::Error: undefined method `gsub' for nil:NilClass`

...lib/ruby/2.1.0/cgi/util.rb:  37:in `escapeHTML'
...liquid-74cc41ce74ee/lib/liquid/standardfilters.rb:  36:in `escape'
...liquid-74cc41ce74ee/lib/liquid/strainer.rb:  50:in `invoke'
...liquid-74cc41ce74ee/lib/liquid/context.rb: 103:in `invoke'
...liquid-74cc41ce74ee/lib/liquid/variable.rb:  78:in `block in render'
```

It's introduced by issue #611.

Maybe it should use same patter as `url_encode` filter:

```
def escape(input)
  CGI.escapeHTML(input).untaint unless input.nil?
end
```

PR #665, opened 2015-11-06
Fixed issue where "nil" value for "escape" filter breaks rendering

tanelj:
Fix for issue #664.


fw42:
Look good to me! Thanks for the fix. @dylanahsmith :+1:?

dylanahsmith:
:+1: Thanks for the fix!

PR #667, opened 2015-11-09
Remove nil and empty string check in invoke_drop.

dylanahsmith:
@tjoyal & @fw42 for review

The `method_or_key && method_or_key != EMPTY_STRING` checks aren't necessary in invoke_drop, since we now use `Set#include?` instead of `Object#respond_to?` to see if the method is invokable, which doesn't have any problems with `nil` or `""`

### DIFF
--- a/lib/liquid/drop.rb
+++ b/lib/liquid/drop.rb
@@ -23,8 +23,6 @@ module Liquid
   class Drop
     attr_writer :context
 
-    EMPTY_STRING = ''.freeze
-
     # Catch all for the method
     def liquid_method_missing(_method)
       nil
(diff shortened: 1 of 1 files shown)


fw42:
Can you check if the shopify test suite passes with this? If so, then :ship:

tjoyal:
Make much sense. Just sad they is not any failing tests :smile_cat: 

:+1: :shipit:

fw42:
There is on failing tests because we don't run tests on Ruby 1.8 anymore

dylanahsmith:
> Can you check if the Shopify test suite passes with this?

checked and CI was :green_apple: when using this on a branch.

> Just sad they is not any failing tests :smile_cat:

If I add `raise if !method_or_key || method_or_key == ""` at the start of `invoke_drop`, then DropsTest#test_nil_value_access and DropsTest#test_empty_string_value_access tests fail.

PR #672, opened 2015-11-20
Test mapping over procs

gmalette:
Mapping over an array of drops that return procs will return an array of procs. Any other filter after that will operate on the proc. 

Without the patch, the template would render: 

```
"#<Proc:0xXXXXXX@/home/vagrant/src/liquid/test/integration/standard_filter_test.rb:255 (lambda)>#<Proc:0xXXXXXX@/home/vagrant/src/liquid/test/integration/standard_filter_test.rb:258 (lambda)>"
```

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -173,7 +173,8 @@ module Liquid
         if property == "to_liquid".freeze
           e
         elsif e.respond_to?(:[])
-          e[property]
+          r = e[property]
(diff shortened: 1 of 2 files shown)


gmalette:
@dylanahsmith

fw42:
Ping?

fw42:
Looks reasonable to me :+1:

gmalette:
Lol I had forgotten all about this. I'll merge tomorrow

pushrax:
LGTM

PR #673, opened 2015-11-21
make file system assignment thread-safe

Azdaroth:
Had some weird issues with thread safety during multiple requests when assigning custom file system globally via class-level accessor. Solved it eventually by assigning it per instance level when rendering templates, which seems to be the most proper solution, but I think it's good to improve it anyway and make it thread-safe.

@fw42, could you review it?

### DIFF
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -16,8 +16,6 @@ module Liquid
     attr_accessor :root
     attr_reader :resource_limits, :warnings
 
-    @@file_system = BlankFileSystem.new
-
     class TagRegistry
       def initialize
         @tags  = {}
@@ -64,11 +62,11 @@ module Liquid
       attr_writer :taint_mode
 
       def file_system
-        @@file_system
+        Thread.current[:liquid_file_system] ||= BlankFileSystem.new
       end
 
       def file_system=(obj)
-        @@file_system = obj
+        Thread.current[:liquid_file_system] = obj
       end
 
       def register_tag(name, klass)


Azdaroth:
What's the status of this? Could anyone review it?

lib/liquid/template.rb
```diff
@@ -64,11 +62,11 @@ class << self
       attr_writer :taint_mode
 
       def file_system
-        @@file_system
+        Thread.current[:liquid_file_system] ||= BlankFileSystem.new
```
carsonreinke: Isn't this a race condition?

Azdaroth: Well, it's `Thread.current`, how would that be a race condition? I think it's a pretty common pattern to ensure thread-safety, no?

carsonreinke: Yep you are right, my bad.

dylanahsmith: Won't this return nil for other threads if the filesystem was only set once at initialization?

Azdaroth: I think it never returns nil, if `Thread.current[:liquid_file_system]` is nil it will assign `BlankFileSystem.new` and return it, same thing for every separate thread.

dylanahsmith: Oh right, but `BlankFileSystem.new` is still wrong a multi-threaded program that initialized this to something else.  So this will at least be a breaking change, and might also be awkward for multi-threaded programs, since they will need to initialize this whenever starting a thread that might render liquid.

You said yourself in the PR description that

> Solved it eventually by assigning it per instance level when rendering templates, which seems to be the most proper solution

so why not allow the filesystem to be overridden on a template instance so this isn't a breaking change.

Azdaroth: Oh, indeed, I see the issue now. About overriding the filesystem on instance level -  it's already possible, here: https://github.com/Shopify/liquid/blob/d8bda2c8923c6b3048678016dae49a4fecd963fc/lib/liquid/tags/include.rb#L105

But I'm wondering actually if that isn't a desired behaviour in such case. I think it's less surprising that each thread would require separate assignment of file system or it would default to the default one otherwise than the situation when the threads override the global file system at the same time. Indeed it's a breaking change, but looks like a step in the right direction.

dylanahsmith: I think it would be quite surprising to set Liquid::Template.file_system and then find it returns a BlankFileSystem instance on another thread.  The point is that BlankFileSystem isn't an appropriate default, so the application would want to specify a different default file system during initialization.  Having a per-thread default seems more complicated to reason about than using `context.registers[:file_system] =` to change that state for the code that uses a different file system.

Azdaroth: Ok, seems fair. Maybe it would be best to deprecate this class-level accessor then and use instead: `context.register.fetch(: file_system) { BlankFileSystem.new }` so that no crash between threads could happen. This PR I guess can be closed.

dylanahsmith:
I'm going to close this, since `template.registers[:file_system] = file_system` can be used instead.

PR #679, opened 2015-12-18
removed redundancy at htmltags file

akilegaspi:
Just a heads up on this part since there's a multiple declaration of index, nothing special.


dylanahsmith:
Thanks

PR #680, opened 2015-12-23
Ensure truncate is operating on a string

jcheatham:
@fw42 & @dylanahsmith for review

Adds a fix for the truncate filter when it's the first/only filter on a non-string.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -63,10 +63,11 @@ module Liquid
     # Truncate a string down to x characters
     def truncate(input, length = 50, truncate_string = "...".freeze)
       return if input.nil?
+      input_str = input.to_s
       length = Utils.to_integer(length)
       l = length - truncate_string.length
       l = 0 if l < 0
-      input.length > length ? input[0...l] + truncate_string : input
+      input_str.length > length ? input_str[0...l] + truncate_string : input_str
     end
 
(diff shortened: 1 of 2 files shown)


fw42:
I'm curious. What's your use case? What problem does this solve?

jcheatham:
In Zendesk people will invoke something like `{{ticket.comment | truncate:50}}` which is an object with other properties, but can also be treated like a string.  So you can do something like `{{ticket.comment | upcase | truncate:50}}` but without this or a custom filter like:

```ruby
def truncate(input, *args)
  super(input.to_s, *args)
end
```

truncate gets invoked on the object which will then fail, e.g.:

```
  1) Error:
StandardFiltersTest#test_truncate_calls_to_liquid:
NoMethodError: undefined method `length' for #<TestThing:0x007fa11358f708 @foo=1>
    liquid/lib/liquid/standardfilters.rb:71:in `truncate'
    liquid/lib/liquid/strainer.rb:50:in `invoke'
```

Right now we're using that custom filter, but it'd be nice to fix for everybody since according to the comment just above truncate it should be operating on a string... but it's never actually invoking `to_s` on the input to ensure that's the case :)

pushrax:
We do this in all the other string filters, so it makes sense to me. :+1:

dylanahsmith:
:+1:

Issue #678, opened 2015-12-05
Slow for loops


chrismytton:
I've been running into some performance issues with `for` loops being slow when iterating over lots of items. It seems that even when the `for` loop has no body it's still quite slow.

I've put together a benchmark which reproduces this problem with some real-world data I've been working with. https://gist.github.com/chrismytton/a34ace5123c27e11dd55

Related to https://github.com/Shopify/liquid/issues/508

parkr:
Was just running an object stackprof of Jekyll and noticed the for loop creates a lot of object:

```text
   5335880 (145.3%)       39984   (1.1%)     block (2 levels) in Liquid::For#render
```

From:

```console
$ cd jekyll
$ export BENCHMARK=true
$ bundle install
$ script/stackprof object
==================================
  Mode: object(100)
  Samples: 3671823 (0.00% miss rate)
  GC: 0 (0.00%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
    801771  (21.8%)      801771  (21.8%)     Jekyll::URL#sanitize_url
    358200   (9.8%)      358200   (9.8%)     Jekyll::Utils#add_permalink_suffix
    492668  (13.4%)      314216   (8.6%)     Jekyll::URL#generate_url_from_drop
   1954884  (53.2%)      148840   (4.1%)     Jekyll::Document#url
  12339233 (336.1%)       88465   (2.4%)     Liquid::Block#render_all
   1986808  (54.1%)       85815   (2.3%)     Jekyll::Drops::Drop#[]
    503967  (13.7%)       74127   (2.0%)     Jekyll::Collection#url_template
    429840  (11.7%)       71640   (2.0%)     block in Jekyll::Collection#url_template
     71183   (1.9%)       71183   (1.9%)     Liquid::Variable#taint_check
    214776   (5.8%)       55921   (1.5%)     Kramdown::Parser::Kramdown#parse_spans
    178452   (4.9%)       50287   (1.4%)     block in Jekyll::URL#generate_url_from_drop
     48992   (1.3%)       48992   (1.3%)     #<Module:0x007fc36b4b8fd0>.sanitized_path
     44743   (1.2%)       44743   (1.2%)     Kramdown::Utils::Html#escape_html
     63392   (1.7%)       43122   (1.2%)     Kramdown::Utils::Html#html_attributes
   5335880 (145.3%)       39984   (1.1%)     block (2 levels) in Liquid::For#render
[...]

PR #681, opened 2016-01-06
Reuse 'forloop' hash to save memory allocations

fw42:
Tries to address https://github.com/Shopify/liquid/issues/678.

Not sure if this makes sense or if there is some side-effects that I'm missing (but if I am, then at least this PR should reveal some missing test coverage, since everything is green).

Before:
```
ruby_empty_loop        0.030000   0.000000   0.030000 (  0.026284)
for_loop_no_body       0.840000   0.010000   0.850000 (  0.853402)
for_loop_with_body     1.800000   0.000000   1.800000 (  1.807363)
```

After:
```
ruby_empty_loop        0.030000   0.000000   0.030000 (  0.027673)
for_loop_no_body       0.430000   0.000000   0.430000 (  0.430292)
for_loop_with_body     1.240000   0.010000   1.250000 (  1.259245)
```

(See https://github.com/Shopify/liquid/issues/678 for the benchmark script I'm using.)

The idea here is that I don't see any reason why consecutive loop iterations each create a new Hash Ruby object for the `forloop` variable. Instead, re-use the one from the previous iteration and just update the keys (if necessary).

Also refactor the `For#render` method a little bit to (hopefully) make it a bit more readable, it was getting pretty long and hairy.

@dylanahsmith @pushrax


pushrax:
Good call. I don't see any reason why this would be an issue, as it makes no sense for people to store things on this hash (let alone expect the stored value to go away in the next iteration). LGTM.

lib/liquid/tags/for.rb
```diff
@@ -170,6 +114,74 @@ def strict_parse(markup)
...
+
+          segment.each_with_index do |item, index|
+            context[@variable_name] = item
+
+            loop_vars['index'.freeze] = index + 1
+            loop_vars['index0'.freeze] = index
+            loop_vars['rindex'.freeze] = length - index
+            loop_vars['rindex0'.freeze] = length - index - 1
+            loop_vars['first'.freeze] = (index == 0)
+            loop_vars['last'.freeze] = (index == length - 1)
```
dylanahsmith: Should we instead use a liquid drop? Then we can just increment a single index between iterations of the loop, and calculate these values lazily on methods on the drop.

dylanahsmith:
There is the possibility of a side effect from the `forloop` hash being stored in another variable, but that seems unlikely.

lib/liquid/tags/for.rb
fw42: hm interesting idea, I will try that

pushrax: This is a good idea too.

fw42:
Updated. Even a bit faster now:
```
ruby_empty_loop        0.030000   0.000000   0.030000 (  0.028785)
for_loop_no_body       0.340000   0.000000   0.340000 (  0.349778)
for_loop_with_body     1.170000   0.000000   1.170000 (  1.169294)
```

pushrax:
Awesome stuff. My only concern with the drop approach is that some custom tags might try to access these indexes with the `[]` operator; is this something that might be the case?

fw42:
> My only concern with the drop approach is that some custom tags might try to access these indexes with the [] operator

That should work, `Liquid::Drop` implements `#[]`.

pushrax:
Also, this is probably a good time to remove the duplication in `TableRow`.

pushrax:
> That should work, `Liquid::Drop` implements `#[]`.

Right :facepalm:

lib/liquid/forloop_drop.rb
```diff
@@ -0,0 +1,40 @@
...
+    def initialize(name, length, parentloop)
+      @name = name
+      @length = length
+      @parentloop = parentloop
+      @index = 0
+    end
+
+    attr_reader :name, :length, :parentloop
+
+    def increment
```
dylanahsmith: Do we want to make this protected so it isn't invokable from liquid?

pushrax: :+1:

dylanahsmith:
:+1:

lib/liquid/forloop_drop.rb
fw42: hm if I make it protected, I can't call it from the `For` class either though :-/

fw42: I'm just gonna ghetto `send` it

JulienItard:
any idea when it will be released with V4 ?

PR #683, opened 2016-01-08
Liquid::TablerowloopDrop

fw42:
Same as https://github.com/Shopify/liquid/pull/681 but for table rows.

@pushrax @dylanahsmith


lib/liquid/tags/table_row.rb
```diff
@@ -28,36 +28,21 @@ def render(context)
...
-            'col'.freeze       => col + 1,
-            'col0'.freeze      => col,
-            'rindex'.freeze    => length - index,
-            'rindex0'.freeze   => length - index - 1,
-            'first'.freeze     => (index == 0),
-            'last'.freeze      => (index == length - 1),
-            'col_first'.freeze => (col == 0),
-            'col_last'.freeze  => (col == cols - 1)
-          }
+          context['tablerowloop'.freeze] = tablerowloop
```
dylanahsmith: Shouldn't this be moved out of the loop now that we are re-using the object?

fw42: good point, yeah

dylanahsmith:
:+1:

lib/liquid/tags/table_row.rb
```diff
@@ -28,36 +28,21 @@ def render(context)
...
-
-          if col == cols && (index != length - 1)
-            col  = 0
-            row += 1
-            result << "</tr>\n<tr class=\"row#{row}\">"
+          if tablerowloop.col_last && !tablerowloop.last
+            result << "</tr>\n<tr class=\"row#{tablerowloop.row + 1}\">"
           end
+
+          tablerowloop.send(:increment!)
```
parkr: why use `send` here?

fw42: because the method isn't `public` (if it were, people could call it from within the template, which we don't want).

pushrax:
:ship:  I wish we didn't have duplication between for and tablerow, but oh well.

PR #687, opened 2016-01-23
Performance improvement: `default` filter

pathawks:
See discussion on jekyll/jekyll#4348

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -341,9 +341,12 @@ module Liquid
       raise Liquid::FloatDomainError, e.message
     end
 
-    def default(input, default_value = "".freeze)
-      is_blank = input.respond_to?(:empty?) ? input.empty? : !input
-      is_blank ? default_value : input
+    def default(input, default_value = ''.freeze)
+      if !input || input.respond_to?(:empty?) && input.empty?
+        default_value
+      else
+        input
+      end
     end
 
     private


parkr:
Travis build error is not due to this change, but due to a failure on `ruby-head` to install `liquid-c`:

```text
linking shared-object liquid_c.so
/usr/bin/ld: cannot find -lgmp
collect2: ld returned 1 exit status
make: *** [liquid_c.so] Error 1
```

fw42:
Cool, thanks for the contribution!

fw42:
@pushrax, agree?

lib/liquid/standardfilters.rb
```diff
@@ -341,9 +341,13 @@ def floor(input)
       raise Liquid::FloatDomainError, e.message
     end
 
-    def default(input, default_value = "".freeze)
-      is_blank = input.respond_to?(:empty?) ? input.empty? : !input
-      is_blank ? default_value : input
+    def default(input, default_value = EMPTY_STR)
```
pushrax: Where is `EMPTY_STR` defined?

pushrax:
Checking for falsy values explicitly makes sense, sure. @pathawks can you rebase from master to fix the build?

lib/liquid/standardfilters.rb
pathawks: Whoops, good catch.

`''.freeze` is used several times throughout this file; it may not be a bad idea to have an `EMPTY_STR` constant, but that is a discussion for a different PR.

pathawks:
> can you rebase from master to fix the build?

:white_check_mark:

lib/liquid/standardfilters.rb
```diff
@@ -341,9 +341,13 @@ def floor(input)
       raise Liquid::FloatDomainError, e.message
     end
 
-    def default(input, default_value = "".freeze)
-      is_blank = input.respond_to?(:empty?) ? input.empty? : !input
-      is_blank ? default_value : input
+    def default(input, default_value = ''.freeze)
+      return default_value unless input
+      if input.respond_to?(:empty?) && input.empty?
```
pushrax: Sorry I didn't comment this in the previous review, but these two lines may as well be `if !input || input.respond_to ...`

pathawks: Fixed.

Issue #685, opened 2016-01-21
Get all variables used in a template


anitsirc:
Hi, 

Is there a way of finding out all variables being used in a template, or at least the ones that it couldn't replace when rendering the template? Both, the ones used as `{{ name }}` and also the ones used in other tags `{% if name == 'John' %}` 

Thanks :)

fw42:
Not that I know of, no. Sorry :(

pushrax:
You could add some hooks to the variable lookup function if you are ok with running a custom liquid version.

dsignr:
Hi, is there a solution for this yet? I'd also like to find all variables used in a template. Thank you.

fw42:
No solution, sorry. And nothing planned so far. But I'm open to look at suggestions.

dsignr:
Somewhat hacky, but I put together this for my use case:

```ruby
def get_variables(t)
      variables = []
      nodes = Liquid::Template.parse(t).root.nodelist
      nodes.each do |n|
        variables << n.name.name if n.is_a? Liquid::Variable
      end #End loop
      #Return
      variables
 end #End method

>> get_variables("{{client_name}} is such a {{animal}}")
>> ["client_name", "animal"]
```

Not even sure if this is ok. What do you think? :D

pushrax:
Assuming you only care about top level variables that will work. You also need to check that `n.name` is a `VariableLookup` though, if you have something like `{{ "foo" }}` then it'll be a string.

dsignr:
Wow, I didn't know that you could have variables like that. Thanks for the suggestion, thanks!

hugopl:
If you export all your variables using Liquid::Drop, you can run your liquid template against mock liquid drop with same methods as the usual liquid drop, but registering what variables were used.

JanStevens:
With https://github.com/Shopify/liquid/pull/1025 being available in 4.0.2 this can now be achieved using the following code

```ruby
Liquid::ParseTreeVisitor.for(Liquid::Template.parse(liquid_string).root).add_callback_for(Liquid::VariableLookup) do |node|
  [node.name, *node.lookups].join('.')
end.visit.flatten.compact
```

viral810:
```ruby
liquid_string = "Hi {{ name }}"

vars = Liquid::Template
.parse(liquid_string)
.root
.nodelist
.select { _1.is_a? Liquid::Variable }
.map(&:name)
.map(&:name)
.uniq

vars # => ["name"]
```

[...]

Issue #490, opened 2014-11-22
Addition to strict mode: fail on nonexistent variable dereferencing


parkr:
In https://github.com/jekyll/jekyll/issues/3008, @gjtorikian wrote of one woe of using Liquid: it swallows references to non-existent variables.

It would be amazing, when `error_mode = :strict`, to throw an error in this case. It would really help Liquid scale out to larger teams, I think, and would help surface annoying rip-your-hair-out issues that Liquid allows for today.

What do you think?

Carpetsmoker:
FYI all, I did some basic work on this, see the commit above. I need to go fix our templates first, but just so that you know it's in the works ;-)

gjtorikian:
Great! Thank you!

parkr:
> @pushrax what do you think?

:point_up:

pushrax:
@Carpetsmoker I added a few comments to your commit, feel free to submit a PR after you take a look at them.

gjtorikian:
Any updates here? :stuck_out_tongue_winking_eye:

parkr:
It would be nice. Very nice.

fw42:
@Carpetsmoker: Are you interested in addressing the comments and finishing this? If not, I can take a look myself.

fw42:
@parkr: Is this something you would want to include in Jekyll 3? If yes, I will take a look soon (since I know you want to release soon). If it can wait, then I will focus on other things for now :-)

parkr:
> Is this something you would want to include in Jekyll 3?

It would be marvelous! We're more lenient in Jekyll 3 with our Liquid version so if it is added to Liquid later, we could turn it on later as well (if it is not automatically enabled).

Carpetsmoker:
@fw42 I did this for my previous job, which is my, well, no longer my job ;-) I don't mind finishing this as such, but at the moment I'm using my girlfriend's OSX machine, and I don't really want to mess around with Ruby and such (apparently you need homebrew and whatnot, and I've never really used OSX).

So yeah, I'll work on it, but not in the next week orso. Anyone else can of course also work on it if she or he has the inclination to do so ;-)

mikebridge:
So is it correct to say that when this is on, an error will be generated when a missing member of a hash or drop is dereferenced?  i.e. both these would always return errors if "missing" is a missing hash key or drop value:

```liquid
    {{ x.missing }}
    {{ x.y.missing.z }}
```

[...]

Issue #422, opened 2014-08-17
Invalid filters are ignored without warning


tadrinth:
Using a filter that Liquid does not recognize causes the invalid filter to be ignored and no warning will be generated.  

It would be nice if there was a way to get a warning in this case, instead of these errors being silently ignored.  I was expected the strict error mode to catch invalid filters upon parse, and I think that would be the ideal solution.

amw:
Just wasted 30 minutes figuring out why a filter I tried to use didn't generate expected results. Turns out it was not implemented in the version of Jekyll that I used (only on master, not released as gem yet).

🤔😮😒

Silently ignoring a missing filter is definitely not nice...

h0jeZvgoxFepBQ2C:
Waat, stumbled over this one... What kind of strict mode is this, if this isn't triggering an exception here?
Anyone knows a workaround to raise an exception?

PR #691, opened 2016-01-29
Detect undefined variables or filters

apeckham:
This adds `strict_variables` and `strict_filters` options to render. When true, `render` collects a list of undefined variables/filters in #errors, and `render!` raises when an undefined variable/filter is used.

Example:

```ruby
template = Liquid::Template.parse("{{x}} {{y}} {{z.a}} {{z.b}}")
template.render({ 'x' => 1, 'z' => { 'a' => 2 } }, { strict_variables: true })
template.errors
#=> [#<Liquid::MissingVariable: Liquid error: missing variable y>, #<Liquid::MissingVariable: Liquid error: missing variable b in {"a"=>2}>]
```

This lets us write unit tests for our templates. Related:
- #685 Get all variables used in a template
- #490 Fail on nonexistent variable dereferencing
- #422 Invalid filters are ignored without warning

Thanks @Jeiwan for working on this.

The loop over [global_filter, exception_handler, strict_variables, strict_filters] in template.rb is to fix a Rubocop cyclomatic complexity error.

Tests included. Will you consider merging? Happy to make changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,3 +73,34 @@ This is useful for doing things like enabling strict mode only in the theme edit
 
 It is recommended that you enable `:strict` or `:warn` mode on new apps to stop invalid templates from being created.
 It is also recommended that you use it in the template editors of existing apps to give editors better error messages.
+
+### Undefined variables and filters
+
+By default, the renderer doesn't raise or in any other way notify you if some variables or filters are missing, i.e. not passed to the `render` method.
+You can improve this situation by passing `strict_variables: true` and/or `strict_filters: true` options to the `render` method.
+When one of these options is set to true, all errors about undefined variables and undefined filters will be stored in `errors` array of a `Liquid::Template` instance.
+Here are some examples:
+
+```ruby
+template = Liquid::Template.parse("{{x}} {{y}} {{z.a}} {{z.b}}")
+template.render({ 'x' => 1, 'z' => { 'a' => 2 } }, { strict_variables: true })
+#=> '1  2 ' # when a variable is undefined, it's rendered as nil
+template.errors
+#=> [#<Liquid::UndefinedVariable: Liquid error: undefined variable y>, #<Liquid::UndefinedVariable: Liquid error: undefined variable b>]
+```
+
+```ruby
+template = Liquid::Template.parse("{{x | filter1 | upcase}}")
+template.render({ 'x' => 'foo' }, { strict_filters: true })
+#=> '' # when at least one filter in the filter chain is undefined, a whole expression is rendered as nil
+template.errors
+#=> [#<Liquid::UndefinedFilter: Liquid error: undefined filter filter1>]
+```
+
+If you want to raise on a first exception instead of pushing all of them in `errors`, you can use `render!` method:
+
+```ruby
+template = Liquid::Template.parse("{{x}} {{y}}")
+template.render!({ 'x' => 1}, { strict_variables: true })
+#=> Liquid::UndefinedVariable: Liquid error: undefined variable y
+```
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -76,6 +76,9 @@ module Liquid
           end
         rescue MemoryError => e
           raise e
+        rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
+          context.handle_error(e, token.line_number)
+          output << nil
         rescue ::StandardError => e
           output << context.handle_error(e, token.line_number)
         end
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -13,7 +13,7 @@ module Liquid
   #   context['bob']  #=> nil  class Context
   class Context
     attr_reader :scopes, :errors, :registers, :environments, :resource_limits
-    attr_accessor :exception_handler, :template_name, :partial, :global_filter
+    attr_accessor :exception_handler, :template_name, :partial, :global_filter, :strict_variables, :strict_filters
 
     def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_errors = false, resource_limits = nil)
       @environments     = [environments].flatten
@@ -207,6 +207,8 @@ module Liquid
     def lookup_and_evaluate(obj, key)
       if (value = obj[key]).is_a?(Proc) && obj.respond_to?(:[]=)
         obj[key] = (value.arity == 0) ? value.call : value.call(self)
+      elsif strict_variables && obj.respond_to?(:key?) && !obj.key?(key)
+        raise Liquid::UndefinedVariable, "undefined variable #{key}"
       else
         value
       end
--- a/lib/liquid/drop.rb
+++ b/lib/liquid/drop.rb
@@ -24,8 +24,9 @@ module Liquid
     attr_writer :context
 
     # Catch all for the method
-    def liquid_method_missing(_method)
-      nil
+    def liquid_method_missing(method)
+      return nil unless @context.strict_variables
+      raise Liquid::UndefinedDropMethod, "undefined method #{method}"
     end
 
     # called by liquid to invoke a drop
--- a/lib/liquid/strainer.rb
+++ b/lib/liquid/strainer.rb
@@ -26,7 +26,7 @@ module Liquid
     end
 
     def self.add_filter(filter)
-      raise ArgumentError, "Expected module but got: #{f.class}" unless filter.is_a?(Module)
+      raise ArgumentError, "Expected module but got: #{filter.class}" unless filter.is_a?(Module)
       unless self.class.include?(filter)
         send(:include, filter)
         @filter_methods.merge(filter.public_instance_methods.map(&:to_s))
@@ -48,6 +48,8 @@ module Liquid
     def invoke(method, *args)
       if self.class.invokable?(method)
         send(method, *args)
+      elsif @context && @context.strict_filters
+        raise Liquid::UndefinedFilter, "undefined filter #{method}"
       else
         args.first
       end
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -181,12 +181,7 @@ module Liquid
 
         registers.merge!(options[:registers]) if options[:registers].is_a?(Hash)
 
-        context.add_filters(options[:filters]) if options[:filters]
-
-        context.global_filter = options[:global_filter] if options[:global_filter]
-
-        context.exception_handler = options[:exception_handler] if options[:exception_handler]
-
+        apply_options_to_context(context, options)
       when Module, Array
         context.add_filters(args.pop)
       end
@@ -234,6 +229,14 @@ module Liquid
       else
         yield
       end
+    end
+
+    def apply_options_to_context(context, options)
+      context.add_filters(options[:filters]) if options[:filters]
+      context.global_filter = options[:global_filter] if options[:global_filter]
+      context.exception_handler = options[:exception_handler] if options[:exception_handler]
+      context.strict_variables = options[:strict_variables] if options[:strict_variables]
+      context.strict_filters = options[:strict_filters] if options[:strict_filters]
     end
   end
 end
--- a/lib/liquid/variable_lookup.rb
+++ b/lib/liquid/variable_lookup.rb
@@ -55,9 +55,11 @@ module Liquid
           object = object.send(key).to_liquid
 
           # No key was present with the desired value and it wasn't one of the directly supported
-          # keywords either. The only thing we got left is to return nil
+          # keywords either. The only thing we got left is to return nil or
+          # raise an exception if `strict_variables` option is set to true
         else
-          return nil
+          return nil unless context.strict_variables
+          raise Liquid::UndefinedVariable, "undefined variable #{key}"
         end
 
         # If we are dealing with a drop here we have to
--- a/lib/liquid/errors.rb
+++ b/lib/liquid/errors.rb
@@ -56,4 +56,7 @@ module Liquid
   MemoryError = Class.new(Error)
   ZeroDivisionError = Class.new(Error)
   FloatDomainError = Class.new(Error)
+  UndefinedVariable = Class.new(Error)
+  UndefinedDropMethod = Class.new(Error)
+  UndefinedFilter = Class.new(Error)
 end
--- a/test/integration/template_test.rb
+++ b/test/integration/template_test.rb
@@ -27,6 +27,12 @@ class ErroneousDrop < Liquid::Drop
   end
 end
 
+class DropWithUndefinedMethod < Liquid::Drop
+  def foo
+    'foo'
+  end
+end
+
 class TemplateTest < Minitest::Test
   include Liquid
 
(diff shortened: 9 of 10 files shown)


lib/liquid/template.rb
```diff
@@ -183,10 +183,9 @@ def render(*args)
 
         context.add_filters(options[:filters]) if options[:filters]
 
-        context.global_filter = options[:global_filter] if options[:global_filter]
-
-        context.exception_handler = options[:exception_handler] if options[:exception_handler]
-
+        %i(global_filter exception_handler strict_variables strict_filters).each do |option|
+          context.send(:"#{option}=", options[option]) if options[option]
+        end
```
pushrax: This is taking DRY too far IMO.

pushrax:
Great work, I think this will be useful for many people. I'll give this PR an actual review tonight before passing judgement.

fw42:
@hkdsun, can you review this as well please?

apeckham:
When strict_variables is true `{% include "foo" %}` raises "missing variable foo" on context.rb:211. I wrote a failing test and we'll investigate a fix.

apeckham:
Actually, I'm not sure what the correct behavior should be in the above situation, after reading include_tag_test. Let us know your thoughts!

lib/liquid/context.rb
```diff
@@ -27,7 +27,7 @@ def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_erro
       @this_stack_used = false
 
       if rethrow_errors
-        self.exception_handler = ->(e) { raise }
+        self.exception_handler = ->(e) { raise e }
```
fw42: Why this change? I don't think this is equivalent.

lib/liquid/context.rb
```diff
@@ -207,6 +207,8 @@ def find_variable(key)
     def lookup_and_evaluate(obj, key)
       if (value = obj[key]).is_a?(Proc) && obj.respond_to?(:[]=)
         obj[key] = (value.arity == 0) ? value.call : value.call(self)
+      elsif strict_variables && obj.respond_to?(:has_key?) && !obj.has_key?(key)
+        raise Liquid::MissingVariable, "missing variable #{key}"
```
fw42: Instead of "missing", can we say "undefined"?

lib/liquid/template.rb
fw42: Maybe make a new method somewhere? `parsee_options` or `configure_context` or whatever?

lib/liquid/variable_lookup.rb
```diff
@@ -54,6 +54,8 @@ def evaluate(context)
         elsif @command_flags & (1 << i) != 0 && object.respond_to?(key)
           object = object.send(key).to_liquid
 
+        elsif context.strict_variables
+          raise Liquid::MissingVariable, "missing variable #{key} in #{object}"
```
fw42: I'm worried about the `in #{object}` part here. That looks like it could potentially leak sensitive information into the rendered output. Can we remove that?

test/integration/strainer_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+
+class StrainerTest < Minitest::Test
+  include Liquid
+
+  def test_add_filter_when_wrong_filter_class
```
fw42: This test looks like a unit test to me, not an integration test... Why is it in `test/integration/`?

README.md
```diff
@@ -73,3 +73,30 @@ This is useful for doing things like enabling strict mode only in the theme edit
 
 It is recommended that you enable `:strict` or `:warn` mode on new apps to stop invalid templates from being created.
 It is also recommended that you use it in the template editors of existing apps to give editors better error messages.
+
+### Undefined variables and filters
+
+By default, the renderer doesn't raise or in any other way notify you if some variables or filters are missing, i.e. not passed to the `render` method. You can improve this situation by passing `strict_variables: true` and/or `strict_filters: true` options to `render` method. When one of these options is set to true, all missing errors will be stored in `errors` array of a `Liquid::Template` instance.
```
fw42: can you add some line breaks here?

lib/liquid/block_body.rb
```diff
@@ -76,6 +76,9 @@ def render(context)
           end
         rescue MemoryError => e
           raise e
+        rescue MissingVariable, MissingDropMethod, MissingFilter => e
+          context.handle_error(e, token.line_number)
+          output << nil
```
fw42: Why is this necessary? Do any tests fail if you remove this line?

lib/liquid/strainer.rb
```diff
@@ -26,7 +26,7 @@ class << self
     end
 
     def self.add_filter(filter)
-      raise ArgumentError, "Expected module but got: #{f.class}" unless filter.is_a?(Module)
+      raise ArgumentError, "Expected module but got: #{filter.class}" unless filter.is_a?(Module)
```
fw42: This was a bug that you are fixing? Can you add a regression test?

fw42:
This looks pretty cool! I'm tempted to even make this the default behaviour in strict mode.

lib/liquid/template.rb
Jeiwan: Maybe `apply_options_to_context`?

lib/liquid/context.rb
Jeiwan: We want to rethrow the same exception, `raise` without parameters throws `RuntimeError` without a message.

lib/liquid/variable_lookup.rb
Jeiwan: The idea is not to print exception messages into the rendered output. This message is only available in `errors` method.
And there's also a problem, that we cannot get full variable path here, only variable name and current scope. So I decided to add the scope to the message for developers to know where exactly the missing variable is.

test/integration/strainer_test.rb
Jeiwan: Hmmm... yep, will move it to unit tests.

lib/liquid/block_body.rb
Jeiwan: We didn't want to print exception messages to the output. Actually this approach was implemented in https://github.com/Shopify/liquid/issues/490 so I also decided to use it.

lib/liquid/strainer.rb
Jeiwan: Yep, it's a bug I happened to come across. Decided to fix it along the way. The test is already there, `test/integration/strainer_test.rb` I'll just move it to unit tests, as discussed above.

lib/liquid/context.rb
Jeiwan: Fixed

README.md
Jeiwan: Done

hkdsun:
This looks good to me after the latest change :+1:

I can also see it being the default behaviour in strict mode, since we encourage strict mode as a way of validating the correctness of templates

lib/liquid/context.rb
fw42: Do any tests fail if you revert this change here?

nickpearson: I don't believe adding the `e` changes the behavior. Calling `raise` inside a `rescue` block re-raises the rescued error. For example:

```ruby
begin; raise ArgumentError, 'test'; rescue; raise; end
# and
begin; raise ArgumentError, 'test'; rescue => e; raise e; end
# both result in => ArgumentError: test
```

fw42: yeah that's what I thought too. Not sure about the semantics if it's with a lambda, but `raise e` and `raise` is usually not equivalent (one re-raises, the other raises a new exception (with a new backtrace)).

nickpearson: It seems `raise` pickups up the error even when called in a lambda, as long as the lambda is called from within the `rescue` block:

```ruby
h1 = ->(e){ raise }
h2 = ->(e){ raise e }
begin; raise ArgumentError, 'test'; rescue => e; h1.(e); end
begin; raise ArgumentError, 'test'; rescue => e; h2.(e); end
# both still result in => ArgumentError: test
```

README.md
```diff
@@ -73,3 +73,32 @@ This is useful for doing things like enabling strict mode only in the theme edit
 
 It is recommended that you enable `:strict` or `:warn` mode on new apps to stop invalid templates from being created.
 It is also recommended that you use it in the template editors of existing apps to give editors better error messages.
+
+### Undefined variables and filters
+
+By default, the renderer doesn't raise or in any other way notify you if some variables or filters are missing, i.e. not passed to the `render` method.
+You can improve this situation by passing `strict_variables: true` and/or `strict_filters: true` options to `render` method.
```
fw42: "to **the** render method"

README.md
```diff
@@ -73,3 +73,32 @@ This is useful for doing things like enabling strict mode only in the theme edit
...
+When one of these options is set to true, all missing errors will be stored in `errors` array of a `Liquid::Template` instance.
```
fw42: maybe "all errors about undefined variables and undefined filters" instead of "all missing errors"?

README.md
```diff
@@ -73,3 +73,32 @@ This is useful for doing things like enabling strict mode only in the theme edit
...
+Here are some examples:
+
+```ruby
+template = Liquid::Template.parse("{{x}} {{y}} {{z.a}} {{z.b}}")
+template.render({ 'x' => 1, 'z' => { 'a' => 2 } }, { strict_variables: true })
+template.errors
+#=> [#<Liquid::MissingVariable: Liquid error: missing variable y>, #<Liquid::MissingVariable: Liquid error: missing variable b in {"a"=>2}>]
```
fw42: Example still has the `MissingVariable`. I think you renamed that. Can you update here?

lib/liquid/context.rb
```diff
@@ -207,6 +207,8 @@ def find_variable(key)
     def lookup_and_evaluate(obj, key)
       if (value = obj[key]).is_a?(Proc) && obj.respond_to?(:[]=)
         obj[key] = (value.arity == 0) ? value.call : value.call(self)
+      elsif strict_variables && obj.respond_to?(:has_key?) && !obj.has_key?(key)
```
fw42: I think this should be `key?`, not `has_key?` (see https://github.com/Shopify/liquid/pull/593)

lib/liquid/variable_lookup.rb
```diff
@@ -54,6 +54,8 @@ def evaluate(context)
         elsif @command_flags & (1 << i) != 0 && object.respond_to?(key)
           object = object.send(key).to_liquid
 
+        elsif context.strict_variables
```
fw42: Hm how about moving this into the `else`? Like
```ruby
else
  return nil unless context.strict_variables
  raise Liquid::UndefinedVariable, "undefined variable #{key} in #{object}"
end
```
I think that would be a bit clearer?

lib/liquid/variable_lookup.rb
```diff
@@ -54,6 +54,8 @@ def evaluate(context)
...
+          raise Liquid::UndefinedVariable, "undefined variable #{key} in #{object}"
```
fw42: I'm still worried about including `#{object}` in the string. I don't see the benefit, but I do see potential leaks.

lib/liquid/context.rb
hkdsun: Sorry. I deleted a comment from this line by accident. 

@fw42 had said:
>I think this should be key?, not has_key? (see #593)

lib/liquid/context.rb
Jeiwan: Yep, it fails without `e`:
```
  1) Failure:
TemplateTest#test_undefined_filters_raise [/Users/Jeiwan/Work/liquid/liquid/test/integration/template_test.rb:296]:
[Liquid::UndefinedFilter] exception expected, not
Class: <RuntimeError>
Message: <"">
---Backtrace---
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/context.rb:30:in `block in initialize'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/context.rb:84:in `handle_error'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/strainer.rb:52:in `invoke'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/context.rb:103:in `invoke'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/variable.rb:78:in `block in render'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/variable.rb:76:in `each'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/variable.rb:76:in `inject'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/variable.rb:76:in `render'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/block_body.rb:93:in `render_node'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/profiler/hooks.rb:5:in `block in render_node_with_profiling'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/profiler.rb:80:in `profile_node_render'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/profiler/hooks.rb:4:in `render_node_with_profiling'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/block_body.rb:72:in `block in render'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/block_body.rb:59:in `each'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/block_body.rb:59:in `render'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/template.rb:196:in `block in render'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/template.rb:230:in `with_profiling'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/template.rb:195:in `render'
/Users/Jeiwan/Work/liquid/liquid/lib/liquid/template.rb:208:in `render!'
/Users/Jeiwan/Work/liquid/liquid/test/integration/template_test.rb:297:in `block in test_undefined_filters_raise'
---------------

519 runs, 1277 assertions, 1 failures, 0 errors, 0 skips
rake aborted!
```

This happens only with `UndefinedFilter` exception because we don't actually raise it, but just handle it with `@context.handle_error` here: https://github.com/Shopify/liquid/pull/691/files#diff-cc7b2b233f09fa41b38a4a44e7c5bb3cR52 So this `exception_handler` executes not inside a `rescue`.
And we can't raise it because we don't want to interrupt the filter chain. So if we have filters like these: `{'foo' | undefinedfilter1 | upcase | undefinedfilter2}` we just want to end up with `FOO` and skip all undefined filters.

README.md
Jeiwan: Fixed

README.md
Jeiwan: Fixed

README.md
Jeiwan: Sorry, missed this. Fixed now.

lib/liquid/variable_lookup.rb
Jeiwan: :+1:

lib/liquid/variable_lookup.rb
Jeiwan: Here's an example where it can be useful:
```ruby
t = Template.parse("{{a}} {{b}} {{x.a}} {{x.b}}")
t.render({'a' => 1, 'x' => { 'a' => 2}}, { strict_variables: true})
# => "1  2 "
 => [#<Liquid::UndefinedVariable: Liquid error: undefined variable b>, #<Liquid::UndefinedVariable: Liquid error: undefined variable b in {"a"=>2}>]
```
The second error message gives more information to a developer where the missing variable is. If we remove `in #{object}` part we will end up with two identical errors messages:
```ruby
# => [#<Liquid::UndefinedVariable: Liquid error: undefined variable b>, #<Liquid::UndefinedVariable: Liquid error: undefined variable b>]
```

lib/liquid/context.rb
Jeiwan: Fixed

lib/liquid/context.rb
fw42: Hm I actually think it would be more intuitive to interrupt the filter chain. Why do you think it's better not to?

pushrax: Imagine `x | undefined | escape`. However I think I agree that the filter chain should be interrupted **and** the result forced to `nil`.

Jeiwan: I think it's better to not interrupt the filter chain. I allows us to get the most successful render result, i.e. all defined filters will be applied and undefined filters will be just skipped. At least this is what we have now, without the changes in this PR.
But it's up to you to decide, which approach is better here.

pushrax: I would argue that with `strict_filters` enabled you should not be able to reference an undefined filter. Unlike variables (which might not be defined yet since they are controlled by the user), filters are static and if undefined must be a typo.

Jeiwan:
@fw42 @nickpearson @pushrax Check this commit out: https://github.com/urbandictionary/liquid/commit/f42518c86a40828000f712dc6c836cd327ebf99d
I reverted the change in `exception_handler`, and now interrupt the filter chain and return `nil` to the output when `UndefinedFilter` exception is risen. There's one issue though: `errors` now contains only the first undefined filter in the chain, not all of them.

lib/liquid/variable_lookup.rb
```diff
@@ -55,9 +55,11 @@ def evaluate(context)
           object = object.send(key).to_liquid
 
           # No key was present with the desired value and it wasn't one of the directly supported
-          # keywords either. The only thing we got left is to return nil
+          # keywords either. The only thing we got left is to return nil or
+          # raise an exception if `strict_variables` option is set to true
         else
-          return nil
+          return nil unless context.strict_variables
+          raise Liquid::UndefinedVariable, "undefined variable #{key} in #{object}"
```
fw42: I see your point here about how this could be useful for developers, but I still disagree and at least in Shopify's use-case (where we don't necessarily trust the author of the template), I think exposing `object` is too dangerous.

fw42:
I like this. If you remove the `object` in the error message, I'm happy to merge this.

@pushrax, :+1:?

nickpearson:
Only capturing one error seems fine to me, as that's how any other programming language (that I'm aware of) works -- it quits on the first unhandled error since the application state is indeterminate at that point and cannot safely continue.

lib/liquid/template.rb
```diff
@@ -181,12 +181,7 @@ def render(*args)
 
         registers.merge!(options[:registers]) if options[:registers].is_a?(Hash)
 
-        context.add_filters(options[:filters]) if options[:filters]
-
-        context.global_filter = options[:global_filter] if options[:global_filter]
-
-        context.exception_handler = options[:exception_handler] if options[:exception_handler]
-
+        apply_options_to_context(context, options)
```
pushrax: Good refactor.

lib/liquid/variable_lookup.rb
pushrax: Agree completely.

Jeiwan:
I removed `in #{object}` from the message and improved README (added render results with comments to the examples).

pushrax:
Thanks for this contribution!

Jeiwan:
Great! Thank you for merging it.

gauravmc:
FYI, this caused a bunch of `NoMethodError·undefined method `strict_variables' for nil:NilClass` when latest ref of the gem was tried in Shopify core.

lib/liquid/drop.rb
```diff
@@ -24,8 +24,9 @@ class Drop
     attr_writer :context
 
     # Catch all for the method
-    def liquid_method_missing(_method)
-      nil
+    def liquid_method_missing(method)
+      return nil unless @context.strict_variables
```
dylanahsmith: @gauravmc (regarding https://github.com/Shopify/liquid/pull/691#issuecomment-188461364) more specifically a drop doesn't always have the `@context` set.  This is the line that was causing the NoMethodError exceptions.

PR #693, opened 2016-02-01
Add strict variable checking

timraasveld:
Based on https://github.com/bluerail/liquid/tree/strict_variables.

When `Template#render` is called with the `strict_variables: true` option, it will raise an UndefinedVariableException.

It unfortunately does not respect Liquid's `error_mode` option, because Liquid has all its syntax error checking+handling done at parse-time, at which point no variables are define.


timraasveld:
Just discovered that someone else did exactly the same thing three days ago! #691

PR #694, opened 2016-02-01
Force #to_liquid call in InputIterator#each

pushrax:
This will ensure every object is `to_liquid`'d before being processed.

### DIFF
--- a/lib/liquid/extensions.rb
+++ b/lib/liquid/extensions.rb
@@ -60,3 +60,9 @@ class NilClass
     self
   end
 end
+
+class Proc
+  def to_liquid # :nodoc:
+    self
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/liquid/extensions.rb
```diff
@@ -60,3 +60,9 @@ def to_liquid # :nodoc:
     self
   end
 end
+
+class Proc
+  def to_liquid # :nodoc:
+    self
```
dylanahsmith: Liquid only uses a proc for lazy loading a result, so shouldn't we be calling `.to_liquid` on the result of that call rather than on the Proc itself?  I'm not sure if this is the place to do that though, since the result should be memoized.

devdreamsolution (approved):

PR #703, opened 2016-02-22
Adds hmac_sha256

bobjflong:
We would like to use Shopify proxies to generate authentication codes as a security feature.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -1,5 +1,6 @@
 require 'cgi'
 require 'bigdecimal'
+require 'openssl'
 
 module Liquid
   module StandardFilters
@@ -333,6 +334,12 @@ module Liquid
       Utils.to_number(input).floor.to_i
     rescue ::FloatDomainError => e
       raise Liquid::FloatDomainError, e.message
+    end
+
+    def hmac_sha256(input, secret)
+      return nil if input.nil? || secret.nil?
+      digest = OpenSSL::Digest.new('sha256')
+      OpenSSL::HMAC.hexdigest(digest, secret, input)
     end
 
     def default(input, default_value = ''.freeze)
--- a/test/integration/filter_test.rb
+++ b/test/integration/filter_test.rb
@@ -131,6 +131,12 @@ class FiltersTest < Minitest::Test
     assert_equal "Blub", Template.parse("{{ var | capitalize }}").render(@context)
   end
 
+  def test_hmac_sha256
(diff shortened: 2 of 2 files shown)


fw42:
I don't think there's many use-cases for this... But I also have no reason to say no to this, really..

@pushrax @dylanahsmith thoughts?

lib/liquid/standardfilters.rb
```diff
@@ -335,6 +336,11 @@ def floor(input)
       raise Liquid::FloatDomainError, e.message
     end
 
+    def hmac_sha256(input, secret)
+      digest = OpenSSL::Digest.new('sha256')
+      OpenSSL::HMAC.hexdigest(digest, secret, input)
```
pushrax: Does this handle the case where the input or secret is nil? Usually in Liquid we allow nil values in filters and return nil.

dylanahsmith:
I see that we have md5 and sha1 filters in Shopify, but nothing like that in the liquid standard filters.  Do you think this is likely useful outside of Shopify?  Or should we just have this as a filter in Shopify?

pushrax:
Interesting idea. Also don't really have a reason why this shouldn't be allowed. I think it might make more sense to include it directly in Shopify though.

It sounds like a fun challenge to implement this in pure Liquid!

nickpearson:
I have implemented md5 and sha1 as Liquid filters and use them fairly often, so I can see where this filter would be useful.

lib/liquid/standardfilters.rb
bobjflong: added a nil check

bobjflong:
I wasn't aware there was a division between Shopify and Liquid like that. Our use case involves Shopify proxies. I wouldn't be well placed to suggest where the change should live.

bobjflong:
Anything I could do here? Something like this is the last piece of the puzzle for the Intercom integration. Is there someone in Shopify I should contact?

fw42:
I will push this to Shopify core. Will keep you updated.

fw42:
Thanks for your contribution!

fw42:
Just added the following filters to Shopify:
```
  def sha256(input)
    Digest::SHA256.hexdigest(input.to_s)
  end

  def hmac_sha1(input, secret)
    OpenSSL::HMAC.hexdigest('sha1', secret.to_s, input.to_s)
  end

  def hmac_sha256(input, secret)
    OpenSSL::HMAC.hexdigest('sha256', secret.to_s, input.to_s)
  end
```

PR #705, opened 2016-02-24
Strainer#add_filter Raises on Private Override

gauravmc:
#### Issue

We had an issue [in Shopify core](https://github.com/Shopify/shopify/pull/65328#discussion_r53879658) wherein filter module `A` had a private method with the same name as a public method of a registered filter module `B`. `A` was registered after `B` so this meant that the public filter's definition changed internally. This is kind of scary so `Strainer` should protect against it.

#### Solution

This makes it so that a filter module cannot incorrectly override (`Strainer` raises `MethodOverrideError`) a registered method as private.

@fw42 @pushrax cc @Thibaut

### DIFF
--- a/History.md
+++ b/History.md
@@ -18,6 +18,7 @@
 * Ruby 1.9 support dropped (#491) [Justin Li]
 * Liquid::Template.file_system's read_template_file method is no longer passed the context. (#441) [James Reid-Smith]
 * Remove support for `liquid_methods`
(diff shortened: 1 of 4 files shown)


dylanahsmith:
Should we do the same for protected_instance_methods?

pushrax:
Makes sense to me to do this for protected ones as well.

LGTM otherwise

gauravmc:
@dylanahsmith @pushrax Agree re: protected methods. Updated this. Also see changelog entry.

dylanahsmith:
:+1:

PR #713, opened 2016-03-18
github-pages and Gemfile

admhlt:
Alright, this should help.

cc @Tetsuro


_config.yml
```diff
@@ -1,6 +1,8 @@
 title: Liquid Templating Engine
 description: "Liquid is a template language and accompanying rendering engine. It is built for security, so is perfect for rendering custom templates from your users."
 
+repository: https://github.com/Shopify/liquid
```
parkr: This should just be `Shopify/liquid` :smile:

PR #718, opened 2016-03-19
Docs misc fixes

admhlt:
This fixes all (I think) the broken links, and a couple bad examples and unneeded files. We also don't need to commit the Gemfile.lock, since I think GitHub Pages manages all that.

@Tetsuro @parkr @fw42


basics/types.md
```diff
@@ -119,42 +118,4 @@ Adam
 
 You cannot initialize arrays using only Liquid.
```
pushrax: This is technically incorrect, but I'd like to keep it that way :stuck_out_tongue:

Tetsuro:
Thanks! :+1:

basics/types.md
admhlt: I would die for Liquid to get obvious `push` and `pop` methods.

PR #719, opened 2016-03-20
Make use of liquid syntax highlighting

mrzool:
Github supports it, so why not


fw42:
Hmm looks the same to me in the preview

mrzool:
True...no syntax highlighting shows up, even though Liquid is present among the [languages supported by GitHub](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1906).

Too bad!

PR #726, opened 2016-03-21
Install latest bundler on travis

pushrax:
This doesn't seem to fix ruby-head

@fw42


fw42:
it does NOT?

fw42:
still :red_circle:

jeroenvisser101:
How about running `gem update --system`?

PR #729, opened 2016-03-24
Fix string methods for cyrillic

Galathius:


fw42:
Hmm if I remember correctly, we added UTF8/unicode in the past but then had to revert it because it was a huge performance hit.. @dylanahsmith @boourns

fw42:
I'm gonna do some benchmarks on this branch later to double check

.rubocop.yml
```diff
@@ -123,3 +123,6 @@ Style/TrivialAccessors:
 
 Style/WordArray:
   Enabled: false
+
+Style/ColonMethodCall:
+  Enabled: false
```
fw42: why this? can you just use the regular dot notation please?

fw42:
https://github.com/Shopify/liquid/pull/192 for reference, but I think that was about UTF8 identifiers, not about UTF8 inputs for filters.

.rubocop.yml
boourns: We also don't want to merge anything that breaks liquid-c or massively
increases it's complexity.

Galathius: @fw42: This is restricted for use by unicode library

Galathius: @boourns Maybe you have any idea how i can fix it any other way?

fw42: What I mean is why don't you use `Unicode.downcase` instead of `Unicode::downcase`? Then you wouldn't need this rule here.

Galathius: @fw42 oops, you right, sorry.

dylanahsmith:
Ya, UTF-8 identifiers aren't a priority, and are parsed a lot more than these filters will be called.  Unicode aware filters like these seem more important to make it easier to write themes for other languages.

However, this does pull in a gem dependancy with a ruby extension that will prevent other languages from working.  We avoided that in the past with liquid-c, so maybe it makes sense to do something similar in this case as well.  Another gem should be able to override the builtin liquid filters.

Galathius:
@dylanahsmith What do you think about [this gem](https://github.com/diasks2/unicode_case_converter)? I can try to implement it.

devdreamsolution (approved):

PR #731, opened 2016-03-30
Duck typed maths filters

ismasan:
## What

Allow maths filters to work with objects that respond to `#to_number` as well as things that can be coerced into numbers.

## Why

So I can design my own complex types or Drops that can work with maths filters.

## How

If an object responds to `#to_number`, it will be used by `Utils.to_number` when coercing values.

An example:

```ruby
class PriceDrop < Liquid::Drop
  def initialize(amount, currency_symbol)
    @amount, @currency_symbol = product_price, currency_symbol
  end

  def formatted
    "#{@currency_symbol}#{@amount}"
  end

  # .. etc, other methods here

  # this makes it work with Liquid's maths filters
  def to_number
    @amount
  end
end
```

Now it can be used with maths filters

```ruby
price = PriceDrop.new(1000.0, '$')
template = Liquid::Template.parse("net total: {{ price.formatted }}. Tax: {{ price | times: 0.19 }}")
template.render('price' => price) # => "net total: $1000.0. Tax: 190"
```

## Notes / questions

* Standard text filters already use duck typing (by calling `#to_s` on the object), but `Utils.to_number`, used to coerce values in maths filters, does not. I am not sure whether or not this limitation is by design / security reasons, so I'm putting this here up for discussion.
* I went with `#to_number` instead of a more conventional `#to_i` or `#to_f` to avoid accidental coercion (as many types in Ruby-land implement those) and also to mirror `Utils.to_number`. Please advice if this is not a good idea.

### DIFF
--- a/lib/liquid/utils.rb
+++ b/lib/liquid/utils.rb
@@ -52,7 +52,11 @@ module Liquid
       when String
         (obj.strip =~ /\A\d+\.\d+\z/) ? BigDecimal.new(obj) : obj.to_i
       else
-        0
+        if obj.respond_to?(:to_number)
+          obj.to_number
+        else
+          0
+        end
       end
     end
 
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -41,6 +41,16 @@ class TestEnumerable < Liquid::Drop
   end
 end
 
+class NumberLikeThing < Liquid::Drop
+  def initialize(amount)
+    @amount = amount
+  end
+
+  def to_number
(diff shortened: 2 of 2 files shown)


fw42:
I'm not against this... Seems useful to me. @pushrax, @dylanahsmith, thoughts?

dylanahsmith:
Sounds good to me

pushrax:
Thanks for the contribution!

ismasan:
Thank you guys for Liquid! Do we have an idea re. when the next release is going to be?

Issue #730, opened 2016-03-28
Tons of warnings when running Liquid with verbosity.


envygeeks:
They can readily be seen in your own specs (even though you've disabled the repository.) See: https://travis-ci.org/Shopify/liquid/jobs/117479561#L575

While normally I would not complain about warnings, this severely impacts several projects (many of which are my own) but one public project (Jekyll) because now we cannot run our specs with verbosity to cache our own mistakes, we must disable said verbosity because Liquid throws thousands of warnings for our specs/tests.

PR #733, opened 2016-03-30
Fix a bunch of Ruby warnings

fw42:
Closes https://github.com/Shopify/liquid/issues/730. Mostly about uninitialized instance variables.

@pushrax


pushrax:
:sheep:

dylanahsmith:
:+1:

PR #734, opened 2016-03-30
Raise a Liquid::Error when a non-array is passed into the concat filter.

dylanahsmith:
@tjoyal & @pushrax for review

## Problem

Passing a non-array as an argument to the concat filter causes a TypeError to be raised, however, it should have been a Liquid::Error exceptions since it is a bug in the liquid template that shouldn't be reported as an internal error.

## Solution

Explicitly raise the error rather as a Liquid::ArgumentError.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -216,6 +216,9 @@ module Liquid
     end
 
     def concat(input, array)
+      unless array.respond_to?(:to_ary)
+        raise ArgumentError.new("concat filter requires an array argument")
+      end
       InputIterator.new(input).concat(array)
     end
 
@@ -370,7 +373,7 @@ module Liquid
       end
 
       def concat(args)
-        to_a.concat args
+        to_a.concat(args)
       end
 
       def reverse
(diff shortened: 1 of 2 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -370,7 +370,10 @@ def join(glue)
       end
 
       def concat(args)
-        to_a.concat args
+        unless args.respond_to?(:to_ary)
```
fw42: why `to_ary` and not `to_a`? that's what you are calling below

fw42:
:+1:

lib/liquid/standardfilters.rb
dylanahsmith: No, we don't call `args.to_a` below, and Array#concat uses `.to_ary`.  Also, `nil.to_a` returns `[]` and we don't want that implicit conversion.

dylanahsmith: No, we don't call `args.to_a` below, and Array#concat uses `.to_ary`.  Also, `nil.to_a` returns `[]` and we don't want that implicit conversion.

fw42: ah, now I get it

fw42:
lets get this into 4rc2

pushrax:
LGTM :ship:

lib/liquid/standardfilters.rb
tjoyal: Why here instead of in the "public" interface? Seems like a step too deep in the stack?

fw42:
:ship:?

lib/liquid/standardfilters.rb
dylanahsmith: Sure.  I moved it to the concat filter method.

tjoyal:
:+1:

PR #736, opened 2016-03-31
Abs filter

fw42:
The `abs` filter returns the absolute value of a number (i.e. the number itself if it's non-negative and -1 times the number if it's negative).

@pushrax @dylanahsmith @admhlt @freakdesign

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -289,6 +289,12 @@ module Liquid
       array.last if array.respond_to?(:last)
     end
 
+    # absolute value
+    def abs(input)
+      result = Utils.to_number(input).abs
+      result.is_a?(BigDecimal) ? result.to_f : result
+    end
+
     # addition
     def plus(input, operand)
       apply_operation(input, operand, :+)
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -385,6 +385,19 @@ class StandardFiltersTest < Minitest::Test
     assert_template_result "5", "{{ price | minus:'2' }}", 'price' => NumberLikeThing.new(7)
   end
 
+  def test_abs
+    assert_template_result "17", "{{ 17 | abs }}"
+    assert_template_result "17", "{{ -17 | abs }}"
+    assert_template_result "17", "{{ '17' | abs }}"
+    assert_template_result "17", "{{ '-17' | abs }}"
+    assert_template_result "0", "{{ 0 | abs }}"
+    assert_template_result "0", "{{ '0' | abs }}"
+    assert_template_result "17.42", "{{ 17.42 | abs }}"
+    assert_template_result "17.42", "{{ -17.42 | abs }}"
+    assert_template_result "17.42", "{{ '17.42' | abs }}"
+    assert_template_result "17.42", "{{ '-17.42' | abs }}"
+  end
+
   def test_times
     assert_template_result "12", "{{ 3 | times:4 }}"
     assert_template_result "0", "{{ 'foo' | times:4 }}"


fw42:
Updated to move the BigDecimal to float conversion to the very last step (right before rendering), so that we don't lose precision if `abs` is chained with other filters.

fw42:
Hm actually this might still be losing precision (because `Context#find_variable` calls `to_liquid`). Where's the right place to put this conversion?

pushrax:
Would it break things if we don't convert to float?

fw42:
yea it would.. stringifying big decimals behaves differently, since it uses exponent syntax in some cases
```
irb(main):002:0> 14.42.to_s
=> "14.42"
irb(main):003:0> BigDecimal.new("17.42").to_s
=> "0.1742E2"
```

pushrax:
hmm and any filter can theoretically `to_s` right now, so we have to convert to float in every filter

lib/liquid/extensions.rb
```diff
@@ -25,6 +26,12 @@ def to_liquid
   end
 end
 
+class BigDecimal
+  def to_liquid
+    to_f
```
dylanahsmith: This basically means that we shouldn't store big decimals in liquid.  I don't think we should do that, and it doesn't seem related to adding a filter.

lib/liquid/standardfilters.rb
```diff
@@ -349,8 +352,7 @@ def default(input, default_value = ''.freeze)
     private
 
     def apply_operation(input, operand, operation)
-      result = Utils.to_number(input).send(operation, Utils.to_number(operand))
-      result.is_a?(BigDecimal) ? result.to_f : result
+      Utils.to_number(input).send(operation, Utils.to_number(operand))
```
dylanahsmith: Shouldn't `public_send` work here?

lib/liquid/extensions.rb
fw42: what do you mean by "store"? The issue is that when converting strings to numbers (and they have a decimal point), we internally use BigDecimal. But if you chain filters together, this is kinda pointless... but I agree that this could be a separate PR. Would you prefer that I just do this convert in the `abs` filter? (like the other number filters do)

dylanahsmith: Well, if you chain together filters, then we call .to_liquid on the result to avoid using non-liquid values.  The same is true if you assign the result to a variable then retrieve it from a variable tag.  That makes a BigDecimal seem even more pointless.

fw42: what's your suggestion?

dylanahsmith: I guess I don't really see the problem with `Utils.to_number(input).abs` since that keeps it as a BigDecimal.

There are some sketch conversions like one from BigDecimal to a Float in the round filter and Utils.to_number converting a Float to BigDecimal.  I'm not sure what is the best way to handle that in a backwards manner though, since the default behaviour for BigDecimal#to_s is weird.

fw42: yea exactly, the `to_s` thing is the reason why I didn't want to just do `Utils.to_number(input).abs`

dylanahsmith: Well the same problem would be present if `to_s` was called on the input.  E.g. 

```ruby
irb> Liquid::Template.parse('{{ decimal }}').render("decimal" => BigDecimal.new("12.34"))
=> "0.1234E2"
```

so having the filter convert to a float just makes things less consistent.  I suppose this change makes things more consistent, but it doesn't help for preserving precision.

fw42: > having the filter convert to a float just makes things less consistent. 

why is that? converting it to float will keep the same to_s behaviour as the other filters have (and all the other number filters also convert to float)

dylanahsmith: I mean the way `{{ decimal }}` and `{{ decimal | abs }}` are rendered isn't consistent.  I suppose it would make the abs filter consistent with other math filters.

fw42:
updated.. @dylanahsmith do you like this better now?

dylanahsmith:
:+1:

PR #737, opened 2016-03-31
Improve BlockBody#render performance

fw42:
Using the benchmark script from https://github.com/Shopify/liquid/issues/678, on my machine this gives a ~10% performance improvement in that code path (rendering `for`loops with many iterations). I guess mostly due to avoiding the `Array#join`.

@pushrax @dylanahsmith @KnVerey

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -53,7 +53,7 @@ module Liquid
     end
 
     def render(context)
-      output = []
+      output = ""
       context.resource_limits.render_score += @nodelist.length
 
       @nodelist.each do |token|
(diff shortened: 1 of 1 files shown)


dylanahsmith:
How does the liquid benchmark look with and without this patch?

fw42:
Pretty much the same...

Before:
```
        parse & run:     24.671  (±16.2%) i/s -      1.410k
```

After
```
        parse & run:     24.367  (±16.4%) i/s -      1.396k
```

I guess the benchmark doesn't exercise this codepath that much?

pushrax:
I tried this before, and determined that which one is faster depends on the circumstances. IIRC `Array#join` actually preallocates a large string based on the length of the contents, whereas `String#<<` needs to extend the length each time. If you have few children of the block, `String#<<` is faster, in other cases `Array#join` might actually be better.

I even tried doing this function in C and it didn't help much...

fw42:
interesting..

PR #738, opened 2016-04-02
Switch to kramdown/rouge and fix code blocks

admhlt:
By removing the Markdown and highlighter configuration in _config.yml, this switches the Jekyll site to use the default kramdown for Markdown rendering and Rouge for syntax highlighting. This is a necessary change to ensure the site keep working with GitHub Pages, as support for redcarpet is going away in about a month.

I got rid of some complicated CSS and used a bit of HTML to add labels to code blocks instead of having Jekyll auto-label them with CSS. It's not a perfect solution, but I'm okay with it for now.

cc @Tetsuro


admhlt:
I'm going to merge this to get it over with. Will keep an eye on the site today and revert if necessary.

Tetsuro:
:+1:

PR #739, opened 2016-04-05
Add abs filter docs

admhlt:
Adds brief documentation for the new `abs` filter.

cc @fw42

### DIFF
--- a/filters/abs.md
+++ b/filters/abs.md
@@ -1,0 +1,37 @@
+---
+title: abs
+---
+
+Returns the absolute value of a number.
+
+```liquid
+{% raw %}
+{{ -17 | abs }}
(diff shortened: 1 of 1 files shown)


filters/abs.md
```diff
@@ -0,0 +1,37 @@
...
+{% raw %}
+{{ 4 | abs }}
+{% endraw %}
+```
+
+```text
+4
+```
+
+`abs` will work on a string if the string only contains a number.
```
fw42: maybe say "will also work" instead of "will work"?

fw42:
LGTM thanks!

filters/abs.md
admhlt: 👍

PR #215, opened 2013-06-21
Add ERB-like trim mode for tags and outputs

nickpearson:
This change adds support for ERB-like trimming to Liquid.  Tags and outputs are supported, though they behave differently.

**TAG BEHAVIOR**

For any tag starting with `{%-`, as with any ERB segment starting with `<%-`, all tabs and spaces between the beginning of the line and the beginning of the tag will be removed.

For any tag ending with `-%}`, as with any ERB segment ending with `-%>`, the newline character immediately following the tag will be removed.

For example:

``` html
<p>
  {%- if true -%}
  hi
  {%- endif -%}
</p>
```

becomes:

``` html
<p>
  hi
</p>
```

Unlike ERB, this implementation for Liquid is more forgiving in that (a) it also removes any tabs and spaces that precede the newline character, and (b) it does not require a newline to be present.

**OUTPUT BEHAVIOR**

Trim mode for output segments behaves a little differently.

For any output starting with `{{-`, all tabs and spaces between the beginning of the line and the beginning of the output _and the newline preceding it_ will be removed.  (Tabs and spaces preceding the newline will not be removed.)

For any output ending with `-}}`, all tabs and spaces and the newline character immediately following the output _and all tabs and spaces after the newline_ will be removed.

For example:

``` html
<p>
  {{- 'hi' -}}
</p>
```

becomes:

``` html
<p>hi</p>
```

**NOTES**

In the descriptions of functionality, "newline" means `\n` and/or `\r` (specifically, `/\r?\n?/`).  The reason both are optional is so that `{%-` and `{{-` are handled even if they are at the very beginning of the template, and `-%}` and `-}}` are handled even if they are at the very end of a template.  In fact, none of the characters that are part of the trim are required, ensuring the trim-hyphen is never left in a template when it is ready to be parsed.

The beginning and ending trim modes can be used independently.  For example, `{%- if x %}` or `{{ x -}}`.

The reason tabs and spaces preceding the newline in a `{{-` trim are _not_ removed is so that whitespace between multiple output segments can be controlled.  The reason tabs and spaces following a newline in a `-}}` trim _are_ removed is so that indentation can be removed.  For example, consider the following, where underscores are shown in the place of spaces:

``` html
<p>
__{{- 'abc' -}},_
__{{- 'def' -}}
</p>
```

becomes:

``` html
<p>abc,_def</p>
```

The space after the comma is kept, but the indentation spaces before 'abc' and 'def' are removed.

**BENCHMARKS**

Before:

```
[...]

PR #743, opened 2016-04-13
Add support for whitespace control character allowing for ERB like syntax

evulse:
This is a continuation of #215 and #216 . It contains all PR changes that were needed for @nickpearson to complete his PR. @fw42

### DIFF
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -24,6 +24,8 @@ module Liquid
   ArgumentSeparator           = ','
   FilterArgumentSeparator     = ':'
   VariableAttributeSeparator  = '.'
+  WhiteSpaceControl           = /-/
+  WhiteSpace                  = /[ \t]*\r?\n?[ \t]*/
   TagStart                    = /\{\%/
   TagEnd                      = /\%\}/
   VariableSignature           = /\(?[\w\-\.\[\]]\)?/
@@ -43,12 +45,6 @@ module Liquid
   PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/o
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/o
   VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
-  WhiteSpaceControl           = /-/
-  WhiteSpaceTrimCheck         = /(#{TagStart}|#{VariableStart})#{WhiteSpaceControl}|#{WhiteSpaceControl}(#{TagEnd}|#{VariableEnd})/o
-  WhiteSpaceTagStart          = /[ \t]*\r?\n?[ \t]*(#{TagStart}|#{VariableStart})#{WhiteSpaceControl}/o
-  WhiteSpaceTagEnd            = /#{WhiteSpaceControl}(#{TagEnd}|#{VariableEnd})[ \t]*\r?\n?[ \t]*/o
-  WhiteSpaceTrimTags          = /#{WhiteSpaceTagStart}|#{WhiteSpaceTagEnd}/o
-  WhiteSpaceTrimReplace       = '\1\2'
 end
 
 require "liquid/version"
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -39,13 +39,16 @@ module Liquid
   SpacelessFilter             = /^(?:'[^']+'|"[^"]+"|[^'"])*#{FilterSeparator}(?:#{StrictQuotedFragment})(?:#{FirstFilterArgument}(?:#{OtherFilterArgument})*)?/o
   Expression                  = /(?:#{QuotedFragment}(?:#{SpacelessFilter})*)/o
   TagAttributes               = /(\w+)\s*\:\s*(#{QuotedFragment})/o
-  AnyStartingTag              = /\{\{|\{\%/
+  AnyStartingTag              = /#{TagStart}|#{VariableStart}/o
   PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/o
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/o
   VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
-  WhiteSpaceTrimCheck         = /\{[\{\%]-|-[\%\}]\}/
-  WhiteSpaceTrimTags          = /[ \t]*(#{TagStart})-|-(#{TagEnd})[ \t]*\r?\n?|\r?\n?[ \t]*(#{VariableStart})-|-(#{VariableEnd})[ \t]*\r?\n?[ \t]*/
-  WhiteSpaceTrimReplace       = '\1\2\3\4'
+  WhiteSpaceControl           = /-/
+  WhiteSpaceTrimCheck         = /(#{TagStart}|#{VariableStart})#{WhiteSpaceControl}|#{WhiteSpaceControl}(#{TagEnd}|#{VariableEnd})}/o
+  WhiteSpaceTagStart          = /[ \t]*\r?\n?[ \t]*(#{TagStart}|#{VariableStart})#{WhiteSpaceControl}/o
+  WhiteSpaceTagEnd            = /#{WhiteSpaceControl}(#{TagEnd}|#{VariableEnd})[ \t]*\r?\n?[ \t]*/o
+  WhiteSpaceTrimTags          = /#{WhiteSpaceTagStart}|#{WhiteSpaceTagEnd}/o
+  WhiteSpaceTrimReplace       = '\1\2'
 end
 
 require "liquid/version"
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -43,6 +43,11 @@ module Liquid
   PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/o
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/o
   VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
+  WhiteSpaceTrimTagStart      = [/[ \t]*\{\%-/, '{%']
+  WhiteSpaceTrimTagEnd        = [/-\%\}[ \t]*\r?\n?/, '%}']
+  WhiteSpaceTrimVariableStart = [/\r?\n?[ \t]*\{\{-/, '{{']
+  WhiteSpaceTrimVariableEnd   = [/-\}\}[ \t]*\r?\n?[ \t]*/, '}}']
+  WhiteSpaceTrimTags          = [WhiteSpaceTrimTagStart,WhiteSpaceTrimTagEnd,WhiteSpaceTrimVariableStart,WhiteSpaceTrimVariableEnd]
 end
 
 require "liquid/version"
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -44,7 +44,7 @@ module Liquid
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/o
   VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
   WhiteSpaceTrimCheck         = /\{[\{\%]-|-[\%\}]\}/
-  WhiteSpaceTrimTags          = /[ \t]*(\{\%)-|-(\%\})[ \t]*\r?\n?|\r?\n?[ \t]*(\{\{)-|-(\}\})[ \t]*\r?\n?[ \t]*/
+  WhiteSpaceTrimTags          = /[ \t]*(#{TagStart})-|-(#{TagEnd})[ \t]*\r?\n?|\r?\n?[ \t]*(#{VariableStart})-|-(#{VariableEnd})[ \t]*\r?\n?[ \t]*/
   WhiteSpaceTrimReplace       = '\1\2\3\4'
 end
 
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -43,6 +43,7 @@ module Liquid
   PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/o
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/o
   VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
+  WhiteSpaceTrimCheck         = /\{[\{\%]-|-[\%\}]\}/
   WhiteSpaceTrimTagStart      = [/[ \t]*\{\%-/, '{%']
   WhiteSpaceTrimTagEnd        = [/-\%\}[ \t]*\r?\n?/, '%}']
   WhiteSpaceTrimVariableStart = [/\r?\n?[ \t]*\{\{-/, '{{']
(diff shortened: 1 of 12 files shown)


lib/liquid/template.rb
```diff
@@ -144,6 +144,13 @@ def render!(*args)
     def tokenize(source)
       source = source.source if source.respond_to?(:source)
       return [] if source.to_s.empty?
+
+      if source =~ /\{[\{\%]-|-[\%\}]\}/
+        WhiteSpaceTrimTags.each do |white_space|
+          source = source.gsub(white_space[0], white_space[1])
+        end
+      end
+
```
nickpearson: For sake of readability, your block parameter might be better as:
```ruby
WhiteSpaceTrimTags.each do |(regexp, string)|
  source = source.gsub(regexp, string)
```

fw42:
Hmm not sure why Travis CI didn't run your tests here..

lib/liquid/template.rb
```diff
@@ -144,6 +144,13 @@ def render!(*args)
     def tokenize(source)
       source = source.source if source.respond_to?(:source)
       return [] if source.to_s.empty?
+
+      if source =~ WhiteSpaceTrimCheck
+        WhiteSpaceTrimTags.each do |(regexp, string)|
+          source = source.gsub(regexp, string)
```
fw42: Hmm. I'm a bit concerned that this is essentially doing 4 passes (once for each regexp) over the entire source. I can't imagine that to not be a considerable performance hit.

tobi:
Like I said in the other ticket. We really need the white space work to be included in the existing parser tokenizer. Liquid is used in many super high performance situations. 1% slowdown can easily lead to 100s of thousands in hardware costs for Shopify alone. Performance matters a lot and multiple passes are the death of performance

evulse:
New changes bring it down to the tokenizer + whitespace pass but your right no reason it can't be a part of the tokenizer and just skip the whitespace during that step.

evulse:
The overhead of parsing the white space to the parser tokenizer creates more performance issues than the current proposed solution of just clearing it out in one pass first.

evulse:
I think the only way to handle white space as part of the parser tokenizer without a massive performance hit is to move away from the current regex solution and in to a stream parser and hand off tokens as they are found. If the parser remembers the previous token and uses logic based on the next token to decide how to process it then the stream only ever needs to move in one direction and only ever need to store a single previous token in memory at any one stage.

I've attempted regex look forwards and look backs as part of the parser and it just adds way to much performance loss.

evulse:
I've come to the conclusion that this solution is in the top 95% percentile of all possible solutions. I went down the path of changing the tokenizer but to be honest it was starting to feel like this was no longer the same project. The simplicity of the tokenizer is core to the library and as many people use it, changing the tokenizer so heavily would no doubt introduce breaking changes for people who run custom regex for syntax changes. 

Honestly I think the double regex solution above is the perfect balance of simplicity and performance. The regex passes have better performance separated than what they do have combined. For any liquid not using the new syntax it has a very performant regex check to avoid the full regex scan.

nickpearson:
One of the proposed solutions that crops up in the comments of these whitespace issues is to allow newlines inside of Liquid tags and outputs. I had tried this in the past but always got a `Liquid::SyntaxError: Variable '{{' was not properly terminated with regexp: /\}\}/` error.

However, I just tried the following on the master branch, and it works:

```ruby
str = "
a{{
  'b'
}}c{%
  if true
%}d{%
  endif 
%}e
".strip
Liquid::Template.parse(str).render
#=> "abcde"
```

That's of course a contrived example, but it shows that whitespace can be placed inside Liquid tags/outputs where it will be swallowed automatically. Here's a real-world example:

```html
<ul>{%
  for product in products %}
    <li>{{ product.title }}</li>{%
  endfor %}
</ul>
```

which, with two products, will render as:

```html
<ul>
    <li>Shirt</li>
    <li>Pants</li>
</ul>
```

Here's the code to run in a console if you want to try it out:

```ruby
html = "
<ul>{%
  for product in products %}
    <li>{{ product.title }}</li>{%
  endfor %}
</ul>
".strip
data = { 'products' => [{ 'title' => 'Shirt' }, { 'title' => 'Pants' }] }
puts Liquid::Template.parse(html).render(data)
```

The syntax isn't as pretty as the `{{- var -}}` and `{%- if x -%}`, but the important part is that it works.

Is this behavior something that "just works," or is this something Liquid can commit to supporting officially? Official support might be the best option because there's no performance hit (no code changes needed). The only thing to add, I think, would be tests.

dylanahsmith:
@nickpearson Yes, newlines in tags is officially supported.  It was intentionally added in https://github.com/Shopify/liquid/pull/324 along with tests.

I would still like liquid to support the ERB like syntax, since I think it makes liquid code more readable for template writers.  I think we can address Shopify performance concerns with this feature in liquid-c

lib/liquid/template.rb
```diff
@@ -144,6 +144,11 @@ def render!(*args)
     def tokenize(source)
       source = source.source if source.respond_to?(:source)
       return [] if source.to_s.empty?
+
+      if source =~ WhiteSpaceTrimCheck
+          source = source.gsub(WhiteSpaceTrimTags, WhiteSpaceTrimReplace)
```
dylanahsmith: This will affect the content of raw tags as well

dylanahsmith: I think this will also affect line numbers, which will be confusing

lib/liquid.rb
```diff
@@ -39,10 +39,16 @@ module Liquid
...
   TagAttributes               = /(\w+)\s*\:\s*(#{QuotedFragment})/o
-  AnyStartingTag              = /\{\{|\{\%/
+  AnyStartingTag              = /#{TagStart}|#{VariableStart}/o
   PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/o
   TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/o
   VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
+  WhiteSpaceControl           = /-/
+  WhiteSpaceTrimCheck         = /(#{TagStart}|#{VariableStart})#{WhiteSpaceControl}|#{WhiteSpaceControl}(#{TagEnd}|#{VariableEnd})/o
+  WhiteSpaceTagStart          = /[ \t]*\r?\n?[ \t]*(#{TagStart}|#{VariableStart})#{WhiteSpaceControl}/o
+  WhiteSpaceTagEnd            = /#{WhiteSpaceControl}(#{TagEnd}|#{VariableEnd})[ \t]*\r?\n?[ \t]*/o
```
dylanahsmith: This isn't really ERB like in its behaviour.  In ERB it trims a following newline, but doesn't trim spaces after it.  Similarly, the trim on the start tag trims up to the newline, but not including it in ERB.

This is also different from go templates, where it will trim all preceding or following whitespace.

tobi: We don't need to adhere to erb. We should go with what's expected by
designers.
On Fri, Apr 15, 2016 at 1:27 PM Dylan Thacker-Smith <
notifications@github.com> wrote:

> In lib/liquid.rb
> <https://github.com/Shopify/liquid/pull/743#discussion_r59909832>:
>
> >    PartialTemplateParser       = /#{TagStart}.*?#{TagEnd}|#{VariableStart}.*?#{VariableIncompleteEnd}/o
> >    TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/o
> >    VariableParser              = /\[[^\]]+\]|#{VariableSegment}+\??/o
> > +  WhiteSpaceControl           = /-/
> > +  WhiteSpaceTrimCheck         = /(#{TagStart}|#{VariableStart})#{WhiteSpaceControl}|#{WhiteSpaceControl}(#{TagEnd}|#{VariableEnd})/o
> > +  WhiteSpaceTagStart          = /[ \t]*\r?\n?[ \t]*(#{TagStart}|#{VariableStart})#{WhiteSpaceControl}/o
> > +  WhiteSpaceTagEnd            = /#{WhiteSpaceControl}(#{TagEnd}|#{VariableEnd})[ \t]*\r?\n?[ \t]*/o
>
> This isn't really ERB like in its behaviour. In ERB it trims a following
> newline, but doesn't trim spaces after it. Similarly, the trim on the start
> tag trims up to the newline, but not including it in ERB.
>
> This is also different from go templates, where it will trim all preceding
> or following whitespace.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/Shopify/liquid/pull/743/files/a01ac62c891a653692a950e804fc453ded7ebd78#r59909832>
>
-- 
- tobi
CEO Shopify

dylanahsmith: I'm not sure what a designer would expect.  It could cause confusion when all spaces between two words get trimmed, e.g.

```liquid
<p>
  Some starting sentence.
  {% if condition %}
    Some conditional sentences
    that may span multiple lines.
  {% endif %}
  Some following sentence.
</p>
```

I would expect that type of problem would be more common for plain text templates (e.g. for emails).

evulse: I think from a developer point of view the full space clear provides the best expected results. With your issue tag I think its just a matter of developers to know the expected result which is all white space on the current line and the line above/below. If you wanted to nice output on the example provided I would just be dropping inner white space and not outer

```
<p>
  Some starting sentence.
  {% if condition -%}
    Some conditional sentences
    that may span multiple lines.
  {%- endif %}
  Some following sentence.
</p>
```

```
Which gives the below
```
<p>
  Some starting sentence.
  Some conditional sentences
    that may span multiple lines.
  Some following sentence.
</p>
```

Comment (evulse) [inline lib/liquid.rb]: Also 
```
<p>
  Some starting sentence.
  {%- if condition %}
    Some conditional sentences
    that may span multiple lines.
  {%- endif %}
  Some following sentence.
</p>
```
Would provide 
```
<p>
  Some starting sentence.
    Some conditional sentences
    that may span multiple lines.
  Some following sentence.
</p>
```

Comment (evulse) [inline lib/liquid.rb]: I think the doubled sided whitespace clear is insanely useful when your trying to build a url or other string using logic inside a template. It removes the huge one line requirement that is currently needed.

File: lib/liquid/template.rb (hunk 3 above)
Comment (evulse) [inline lib/liquid/template.rb]: I'll added some tests for this. This really means the logic needs to be in the parser.

Comment (dylanahsmith) [inline lib/liquid/template.rb]: Yes, I think Liquid::BlockBody#parse is the right place for this logic.

Comment (evulse) [inline lib/liquid/template.rb]: Need to work something out. Currently the whitespace would be in the previous token so either need to bring the whitespace in to the current token which kills the performance of the tokenizer regex split or create logic that looks back and modifies the stack if it finds a whitespace control symbol.

Comment (evulse) [inline lib/liquid/template.rb]: I just had a look at the liquid-c parser. I think based on that its better to change the tokenizer first as honestly this is where it should be handled to get the outer white space in the same token.  I'll just tweak the performance the best I can in ruby as the c side won't have an issue looking at it.

Comment (dylanahsmith) [inline lib/liquid/template.rb]: > Currently the whitespace would be in the previous token so either need to bring the whitespace in to the current token which kills the performance of the tokenizer regex split or create logic that looks back and modifies the stack if it finds a whitespace control symbol.

In the ruby implementation, I was thinking that BlockBody#parse can modify the last item in the nodelist to rstrip whitespace to trim the previous token.

Trimming the following token might be simplest to handle with the help of the tokenizer by setting a flag on the tokenizer to tell it to lstrip the next token.  That avoids the complication of interacting with child tags which may or may not be blocks.

liquid-c is able to handle things in the tokenizer more easily, since it lazily tokenizes in a way that allows trimming to be turned off while parsing the raw tag.  When matching a TOKEN_RAW token it actually need to read-ahead to find where it ends, so could easily check for an additional dash and trim whitespace from the end.  When matching a tag or variable in tokenizer_next, it could easily check for a dash preceding the end of the tag, then set a flag on the tokenizer to ltrim whitespace on the next token.

**Comment by evulse** [comment]
The new change is slightly slower than the previous approach however I think it still has some optimisations I can do. It does however pass all tests and handles raw tags correctly.

Also I don't generally write in ruby so anything that can be written in a better syntax can you code review

Comments?

**Comment by evulse** [comment]
Also by slower I mean marginally, it could actually end up being faster once I do some more tweaks.

File: lib/liquid/block.rb [hunk 5]
```diff
@@ -1,9 +1,16 @@
 module Liquid
   class Block < Tag
-    IsTag             = /^#{TagStart}/o
```
dylanahsmith: You seem to be working off a really old commit.  See how different the file is now: https://github.com/Shopify/liquid/blob/master/lib/liquid/block.rb

lib/liquid/block.rb
```diff
@@ -13,10 +20,22 @@ def parse(tokens)
       @blank = true
       @nodelist ||= []
       @nodelist.clear
+      @@whitespace ||= false
```
dylanahsmith: Please don't maintain parse state in class variables.

lib/liquid/block.rb
```diff
@@ -13,10 +20,22 @@ def parse(tokens)
...
 
       while token = tokens.shift
         case token
         when IsTag
+          @@whitespace = false
+          if token =~ IsWhiteSpaceTagEnd
+            @@whitespace = true
```
dylanahsmith: I think this flag should be moved to the tokenizer so that it applies to the next token.  It should also be named more appropriately to reflect that it will trim whitespace rather than saying the token is whitespace.

lib/liquid/block.rb
```diff
@@ -13,10 +20,22 @@ def parse(tokens)
...
+          end
+          if token =~ IsWhiteSpaceTagStart
```
dylanahsmith: Using a regex to check a single byte seems like overkill if you are worried about performance.  E.g. `s.getbyte(2) == TRIM_CHAR` where `TRIM_CHAR = '-'.ord`

lib/liquid.rb
```diff
@@ -24,6 +24,8 @@ module Liquid
   ArgumentSeparator           = ','
   FilterArgumentSeparator     = ':'
   VariableAttributeSeparator  = '.'
+  WhiteSpaceControl           = /-/
+  WhiteSpace                  = /[ \t]*\r?\n?[ \t]*/
```
dylanahsmith: This name is too generic when it is for a very specific purpose.

dylanahsmith: I don't think the newline should be optional.  For instance, in the strip_newlines filter we use `/\r?\n/`.

Also, what is the advantage of having it defined like this rather than just stripping all space characters preceding or following a tag using `/\s*/` like with go templates?

lib/liquid/block.rb
```diff
@@ -40,11 +59,27 @@ def parse(tokens)
...
+          @@whitespace = false
+          if token =~ IsWhiteSpaceVariableEnd
+            @@whitespace = true
+          end
+          if token =~ IsWhiteSpaceVariableStart
+            previous_token = @nodelist.pop
+            if previous_token.is_a?(String)
+              previous_token.gsub!(WhiteSpaceEnd, '')
+              @nodelist << previous_token
+            end
```
dylanahsmith: This is doing the same thing as for tags, so the implementation should be shared.  Even the check for the control character could be shared since we we just need to check the third byte for a `-` in both cases.

lib/liquid/block.rb
evulse: That explains the merge conflict message. Updating my fork now as your right its from 2013.

lib/liquid.rb
evulse: I'm not against a full trim. If anything the more it cleans up the whitespace the better. How useful that is i'm not sure. Maybe if your designing a template and you are using empty lines to break apart different blocks

dylanahsmith: I'm not saying a full trim is more useful.  It is just simpler.  If it isn't useful to have it stop stripping whitespace on the second newline preceding or following a tag, then why don't we just use `lstrip` and `rstrip` which are easier to document and understand.

evulse:
Closing due to old commit

PR #744, opened 2016-04-13
Make markup validation a method on Liquid::Raw

DrewMartin:
This lets subclasses overwrite the method if they want additional arguments

@gauravmc @Thibaut


gauravmc:
cc @pushrax we want to make it so that when `Raw` tag is inherited it can accept arguments and send corresponding error for it (see https://github.com/Shopify/shopify/pull/69337).

gauravmc:
:+1:

pushrax:
Seems reasonable to me.

PR #748, opened 2016-04-22
Make Template.tags loop-able

gauravmc:
This makes it so that you can get all registered template tags by looping over them. This is a small followup to [an older PR](https://github.com/Shopify/liquid/pull/590) that made it possible to have custom tag registry for a custom liquid document. Reason for this change is that in our most recent implementation of section liquid template (Please see [this comment](https://github.com/Shopify/shopify/pull/68815#discussion_r59066058) for more.), we want the parsing step to recognize existing tags + few new tags specific to section template.

@dylanahsmith @pushrax cc @Thibaut

### DIFF
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -19,6 +19,8 @@ module Liquid
     @@file_system = BlankFileSystem.new
 
     class TagRegistry
+      include Enumerable
+
       def initialize
         @tags = {}
         @cache = {}
@@ -39,6 +41,10 @@ module Liquid
       def delete(tag_name)
         @tags.delete(tag_name)
         @cache.delete(tag_name)
+      end
+
+      def each(&block)
+        @tags.each(&block)
       end
 
       private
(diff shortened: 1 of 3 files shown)


test/unit/template_unit_test.rb
```diff
@@ -67,4 +67,10 @@ def test_tags_delete
     Template.tags.delete('fake')
     assert_nil Template.tags['fake']
   end
+
+  def test_tags_can_be_looped_over
+    Template.register_tag('fake', FakeTag)
```
dylanahsmith: This is leaking state, because the tag registry is global

dylanahsmith:
Sure, that seems harmless.  :+1:

test/unit/template_unit_test.rb
gauravmc: Hmm that's true. Looks like [this test](https://github.com/Shopify/liquid/blob/23d2beed41cd331dbe81efc2054178dec34b0d4a/test/unit/block_unit_test.rb#L45-L49) also has the same problem.

lib/liquid/template.rb
```diff
@@ -41,6 +43,12 @@ def delete(tag_name)
         @cache.delete(tag_name)
       end
 
+      def each(&block)
+        @tags.each do |tag_name, klass|
```
Thibaut: `@tags.each(&block)`

Issue #749, opened 2016-04-25
Liquid::UndefinedVariable is raised for defined attribute with nil value, when strict_variables: true


nijikon:
``` ruby
2.3.0 :001 > Liquid::Template.parse("{{foo}}").render!({ 'foo' => nil}, strict_variables: true)
Liquid::UndefinedVariable: Liquid error: undefined variable foo
```

Is this desired behaviour?

We want to be strict for variables being defined, nil value is allowed because we handle that with an if.

dylanahsmith:
That seems like a bug to me.

pascalbetz:
We currently have the same problem. Variable is defined but `nil` and this causes mentioned error.
I think the problem lies with `Liquid::Context#find_variable` (https://github.com/Shopify/liquid/blob/master/lib/liquid/context.rb#L163). 


The check for `unless value` (https://github.com/Shopify/liquid/blob/master/lib/liquid/context.rb#L174) does not take `nil` into account which is a valid value when `raise_on_not_found` is set. 

If there is any chance that this gets accepted i'll craft a PR. Let me know.

fw42:
Happy to look at a PR

PR #750, opened 2016-04-25
handle nil value for key when strict_variables is on

nijikon:


lib/liquid/context.rb
```diff
@@ -189,7 +189,7 @@ def find_variable(key)
       if scope.nil?
         @environments.each do |e|
           variable = lookup_and_evaluate(e, key)
```
dylanahsmith: strict_variables really shouldn't be raising in lookup_and_evaluate or we should be calling something else here, since this won't work properly if `@evironments` has a length of more than 1.

For example:

```ruby
template = Liquid::Template.parse("{{ foo }}")
template.assigns["foo"] = "works"
puts template.render!({}).inspect  # => "works"
puts template.render!({}, strict_variables: true).inspect  # raises Liquid::UndefinedVariable
```

dylanahsmith:
Please add a regression test for #749 and the issue I pointed out in https://github.com/Shopify/liquid/pull/750#discussion_r60956416, so we can try to fix this properly this time.

devdreamsolution (approved):

PR #751, opened 2016-04-25
rubocop 0.39.0, drop support for 2.0 and add 2.3.0 in travis

nijikon:


lib/liquid/drop.rb
```diff
@@ -54,7 +54,7 @@ def to_s
       self.class.name
     end
 
-    alias_method :[], :invoke_drop
+    alias [] invoke_drop
```
dylanahsmith: Please keep the existing style and add the following to .rubocop.yml

```
Style/Alias:
  EnforcedStyle: prefer_alias_method
```

lib/liquid/i18n.rb
```diff
@@ -26,7 +26,7 @@ def locale
     def interpolate(name, vars)
       name.gsub(/%\{(\w+)\}/) do
         # raise TranslationError, "Undefined key #{$1} for interpolation in translation #{name}"  unless vars[$1.to_sym]
-        "#{vars[$1.to_sym]}"
+        (vars[$1.to_sym]).to_s
```
dylanahsmith: I don't think the parenthesis are necessary

lib/liquid/variable_lookup.rb
```diff
@@ -1,7 +1,7 @@
 module Liquid
   class VariableLookup
     SQUARE_BRACKETED = /\A\[(.*)\]\z/m
-    COMMAND_METHODS = ['size'.freeze, 'first'.freeze, 'last'.freeze]
+    COMMAND_METHODS = %w(size first last).freeze
```
dylanahsmith: Doesn't this mean that each element won't be frozen now?

lib/liquid/version.rb
```diff
@@ -1,4 +1,4 @@
 # encoding: utf-8
 module Liquid
-  VERSION = "4.0.0.rc2"
+  VERSION = "4.0.0.rc2".freeze
```
dylanahsmith: This was removed on purpose in https://github.com/Shopify/liquid/pull/371, although maybe that is ok now that we don't seem to support ruby 1.9 now.

.rubocop.yml
```diff
@@ -73,9 +74,6 @@ Style/Documentation:
 Style/ClassAndModuleChildren:
   Enabled: false
 
-Style/TrailingComma:
-  Enabled: false
```
dylanahsmith: Please keep the existing style, which was just split into two different styles.

```ruby
Style/TrailingCommaInArguments:
  Enabled: false

Style/TrailingCommaInLiteral:
  Enabled: false
```

.rubocop.yml
```diff
@@ -1,6 +1,7 @@
 inherit_from: ./.rubocop_todo.yml
 
 AllCops:
+  TargetRubyVersion: 2.1
```
dylanahsmith: Is this the minimum ruby version?  If so, .travis.yml indicates it should be 2.0

parkr:
As long as this doesn't totally break compatibility with Ruby 2.0, it's fine. We just ran into a problem where listen, a Jekyll dependency, dropped 2.0 and 2.1 support in a minor bump. This is a major bump (thank you!) but if you drop 2.0 and 2.1 support, we'll have to wait until Jekyll 4 to get all these great updates! Just a request from a heavy user :smile:

.travis.yml
```diff
@@ -1,14 +1,13 @@
+sudo: false
+cache: bundler
 language: ruby
 
 rvm:
-  - 2.0
```
dylanahsmith: Please be focused on your pull request.  You shouldn't need to drop support for ruby 2.0 for code style fixes.

shopmike:
Superseded by https://github.com/Shopify/liquid/pull/1131

PR #765, opened 2016-06-14
Generate index pages for tags and filters

admhlt:
This removes the existing index pages for the basics, tags, and filters directories (which had no content). 

The URL `/basics/` now redirects to the "introduction" page and should never need to be used.

The URL `/tags/` now includes a short summary of tags and then the compiled content from all three pages in that section.

The URL `/filters/` has a short summary of filters and then the compiled content from all the filters pages.

On the new tags and filters index pages, you can click the headings to go to the individual page for that tag or filter.

I did this because I want us to have general links for "tags" and "filters" that we can give people who want to learn more about Liquid, particularly for Jekyll users. Before now, there was no perfect link to give people to say "go check out the docs for Liquid's filters" (it tended to be the Shopify Liquid docs or the 'Liquid for Designers' article).

cc @Tetsuro @parkr

After this gets in I'll update the Jekyll docs on [this page](https://jekyllrb.com/docs/templates/) to point to these new indices.

Filters index


parkr:
Beautiful!!! 😄

PR #746, opened 2016-04-16
Add support for whitespace control character allowing for ERB like trim

evulse:
Rebase of #743. To solve issues #215 and #216

PR #773, opened 2016-06-27
Add whitespace control character and associated tests

evulse:
This adds support for {{- and {%- syntax which will lstrip! and -}} and -%} which will rstrip!

Depends on Shopify/liquid-c#30 to pass all tests.

Resolve issues Shopify/liquid#216, Shopify/liquid#215, Shopify/liquid#214, Shopify/liquid#194, Shopify/liquid#171, Shopify/liquid#162

Previously Pull Request #746

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -1,7 +1,7 @@
 module Liquid
   class BlockBody
-    FullToken = /\A#{TagStart}\s*(\w+)\s*(.*)?#{TagEnd}\z/om
-    ContentOfVariable = /\A#{VariableStart}(.*)#{VariableEnd}\z/om
+    FullToken = /\A#{TagStart}#{WhitespaceControl}?\s*(\w+)\s*(.*?)#{WhitespaceControl}?#{TagEnd}\z/om
+    ContentOfVariable = /\A#{VariableStart}#{WhitespaceControl}?(.*?)#{WhitespaceControl}?#{VariableEnd}\z/om
     TAGSTART = "{%".freeze
     VARSTART = "{{".freeze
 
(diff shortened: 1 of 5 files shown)


evulse:
@fw42 was easier to regenerate pull request on new fork

lib/liquid/block_body.rb
```diff
@@ -48,6 +54,17 @@ def parse(tokenizer, parse_context)
       yield nil, nil
     end
 
+    def whitespace_handler(token, parse_context)
+      if token[2] == WhitespaceControl
+        previous_token = @nodelist.pop
```
dylanahsmith: Won't this pop a node from the nodelist even if the previous token wasn't a string?

pushrax: Should use `.last` as in the C branch.

dylanahsmith:
LGTM

fw42:
The test fails here are because the change in liquid-c isn't merged yet, right?

evulse:
Yes as soon as this is pointed to the new liquid-c commit these pass as expected. Also the liquid-c version passes with this commit included as well.

evulse:
screen shot 2016-08-06 at 11 49 19 pm

Just using Mozilla's Nunjucks on a project and noticed it uses this same syntax. Good to see all the libraries heading in this direction.

fw42:
Thanks for your contribution, great job!

parkr:
Liquid 4?!?!?! Could this be a thing in Liquid 4?!?!

parkr:
💗 🎉 🎊

fw42:
I don't see why not

galopin:
@evulse @fw42 Wondering if this nifty whitespace control feature could be backported into the 3.x branch :interrobang:

PR #781, opened 2016-07-20
Add http for strftime site

MilanAryal:
[strftime](http://strftime.net/) site does not work on secure (https) connection.


admhlt:
Thank you! We must have missed that.

Issue #138, opened 2012-08-19
How to express logic "and","or", "not"?


zhaopengme:
"and" is and.

"or" is or.

but what is "not"  ?

thanks!

courtneynguyen:
I agree with Cam and AllThatIsTheCase. It is difficult to check if product.tags and customer.tags do not contain the tag "wholesale" without a "not contains" operator because of the combined condition.

fw42:
Reopened because of https://github.com/Shopify/liquid/issues/526

Blaisorblade:
From #526:

> same way having support for parentheses would be nice to have

Ah. Sounds like the problem is the parser support for not, because you're not using a parser generator or something like that — and you can't parse the context free grammar of expressions with parentheses with regular expressions such as https://github.com/Shopify/liquid/blob/12d526a05c0654dc67e350f170f91d62fd4875b7/lib/liquid/tags/if.rb#L14. I see why this is not on your roadmap — you'd need to replace that part of the parser.

fw42:
@trishume and @pushrax might have thoughts

pushrax:
We could totally implement parentheses with the new parsers, but it would probably break some weird-ass templates that are using them regardless of whether they do anything or not. We even have a test asserting that parentheses are broken https://github.com/Shopify/liquid/blob/master/test/integration/parsing_quirks_test.rb#L74 :(

trishume:
The other option is to have some kind of super-conditional tag or syntax that is only available with liquid-c or the strict ruby parser. That way it could be opt-in and not break existing templates.

Maybe call it `cond`, `if*` or `if+`. Though the fact that fancy expression features work in that tag but not if would be weird.

Or it could be an opt in for expressions. Maybe some special kind of augmented parentheses in which real expression syntax is valid. Maybe something like the magic test command in bash: `{% if [this && (that || other)] %}`

pushrax:
That would work, but it would also be really confusing going forward. I think if we do this at all we should work towards unifying syntax and formally specifying it.

Blaisorblade:
As a programmer-minded user, I'd be happy to set a flag to enable an extended syntax, where my 
parentheses mean something, and promise I don't use weird templates.

One could also argue that flag should be orthogonal to strict vs lax mode. In fact, when transitioning, I'll probably need to switch to strict mode to check our templates actually parse (i.e., probably, they should contain no parentheses), before enabling a new syntax where these parenthesis suddenly can be used and have meaning.

[...]

PR #787, opened 2016-08-17
Expand boolean support

evulse:
As discussed in #138

### DIFF
--- a/lib/liquid/tags/if.rb
+++ b/lib/liquid/tags/if.rb
@@ -84,22 +84,14 @@ module Liquid
     def strict_parse(markup)
       p = Parser.new(markup)
       condition = parse_binary_comparison(p)
-      p.consume(:space)
       p.consume(:end_of_string)
       condition
     end
 
     def parse_binary_comparison(p)
-      if p.look(:not)
-        n = p.consume(:not)
-        p.consume(:space)
-      end
-      n = p.consume(:bang) if p.look(:bang)
+      n = p.consume?(:not)
       condition = parse_comparison(p)
-      if p.look(:or, 1) || p.look(:and, 1)
-        p.consume(:space)
-        op = (p.type?(:or) || p.type?(:and))
-        p.consume(:space)
+      if op = (p.type?(:or) || p.type?(:and))
         condition.send(op, parse_binary_comparison(p))
       end
       condition.not if n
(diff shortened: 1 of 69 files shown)


lib/liquid/tags/if.rb
```diff
@@ -11,8 +11,8 @@ module Liquid
   #
   class If < Block
     Syntax = /(#{QuotedFragment})\s*([=!<>a-z_]+)?\s*(#{QuotedFragment})?/o
-    ExpressionsAndOperators = /(?:\b(?:\s?and\s?|\s?or\s?)\b|(?:\s*(?!\b(?:\s?and\s?|\s?or\s?)\b)(?:#{QuotedFragment}|\S+)\s*)+)/o
-    BOOLEAN_OPERATORS = %w(and or)
+    ExpressionsAndOperators = /(?:\b(?:\s?and(?:\snot(?=\s))?\s?|\s?or(?:\snot(?=\s))?\s?)\b|(?:\s*(?!\b(?:\s?and(?:\snot(?=\s))?\s?|\s?or(?:\snot(?=\s))?\s?)\b)(?:#{QuotedFragment}|\S+)\s*)+)/o
```
TWiStErRob: Is this supported by the regex?
```
foo and
        not bar
```

test/integration/tags/if_else_tag_test.rb
```diff
@@ -35,6 +35,26 @@ def test_if_or
...
+  def test_if_and_not
+    assert_template_result('',      '{% if a and not b %} YES {% endif %}', 'a' => true, 'b' => true)
+    assert_template_result(' YES ', '{% if a and not b %} YES {% endif %}', 'a' => true, 'b' => false)
+    assert_template_result('',      '{% if a and not b %} YES {% endif %}', 'a' => false, 'b' => true)
+    assert_template_result('',      '{% if a and not b %} YES {% endif %}', 'a' => false, 'b' => false)
+
+    assert_template_result(' YES ', '{% if a and not b and not c %} YES {% endif %}', 'a' => true, 'b' => false, 'c' => false)
+    assert_template_result('',      '{% if a and not b and not c %} YES {% endif %}', 'a' => false, 'b' => false, 'c' => false)
+  end
+
```
TWiStErRob: I think you should test if it supports `{% if true and not 'a' != 'a' %}` which should be true.
Most of the time the conditions are not simple variables, but complex dot-lookups and comparisons.

PR #793, opened 2016-08-28
Fix escape_once not working with hex entities or named entities endin…

### LINKED ISSUE #792: escape_once doesn't ignore hexadecimal entities
{{ "&excl; &#x00021; &#33;" | escape_once }}

Outputs:

    &excl; &amp;#x00021; &#33;

https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references
https://dev.w3.org/html5/html-author/charref

kainjow:
…g with numbers

Fixes #792

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -10,7 +10,7 @@ module Liquid
       '"'.freeze => '&quot;'.freeze,
       "'".freeze => '&#39;'.freeze
     }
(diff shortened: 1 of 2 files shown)


fw42:
Thanks for your contribution! This makes sense to me.

@gauravmc, mind taking a quick look?

kainjow:
I just noticed two issues actually:

- The 'x' for hex entities can be uppercased (just tested briefly with Safari/Chrome/Firefox)
- The hex characters is wrong, it should be stricter `[\da-fA-F]`

Let me go back and fix these. I'll open a new pull request.

PR #795, opened 2016-08-31
Fix escape_once not working with hex entities or named entities ending with numbers

### LINKED ISSUE #792: escape_once doesn't ignore hexadecimal entities

kainjow:
Fixes #792

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -10,7 +10,7 @@ module Liquid
       '"'.freeze => '&quot;'.freeze,
       "'".freeze => '&#39;'.freeze
     }
-    HTML_ESCAPE_ONCE_REGEXP = /["><']|&(?!([a-zA-Z]+|(#\d+));)/
+    HTML_ESCAPE_ONCE_REGEXP = /["><']|&(?!([a-zA-Z]+\d*|(#\d+)|(#[xX][\da-fA-F]+));)/
 
     # Return the size of an array or of an string
     def size(input)
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -134,6 +134,12 @@ class StandardFiltersTest < Minitest::Test
 
   def test_escape_once
     assert_equal '&lt;strong&gt;Hulk&lt;/strong&gt;', @filters.escape_once('&lt;strong&gt;Hulk</strong>')
+    assert_equal '&excl;1&#x00021;2&#33;', @filters.escape_once('&excl;1&#x00021;2&#33;')
+    assert_equal '&ncy;2&#x0043D;3&#1085;', @filters.escape_once('&ncy;2&#x0043D;3&#1085;')
+    assert_equal '&frac34;&#x000BE;&#190;', @filters.escape_once('&frac34;&#x000BE;&#190;')
+    assert_equal '&blk14;&#x02591;&#9617;', @filters.escape_once('&blk14;&#x02591;&#9617;')
+    assert_equal '&UpArrowBar;&DownArrowBar;', @filters.escape_once('&UpArrowBar;&DownArrowBar;')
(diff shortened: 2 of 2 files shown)


test/integration/standard_filter_test.rb
```diff
@@ -134,6 +134,12 @@ def test_escape
 
   def test_escape_once
     assert_equal '&lt;strong&gt;Hulk&lt;/strong&gt;', @filters.escape_once('&lt;strong&gt;Hulk</strong>')
+    assert_equal '&excl;1&#x00021;2&#33;', @filters.escape_once('&excl;1&#x00021;2&#33;')
+    assert_equal '&ncy;2&#x0043D;3&#1085;', @filters.escape_once('&ncy;2&#x0043D;3&#1085;')
+    assert_equal '&frac34;&#x000BE;&#190;', @filters.escape_once('&frac34;&#x000BE;&#190;')
+    assert_equal '&blk14;&#x02591;&#9617;', @filters.escape_once('&blk14;&#x02591;&#9617;')
```
fw42: This one is also different than the ActionView counterpart, which returns `"&amp;blk14;&#x02591;&#9617;"`.

lib/liquid/standardfilters.rb
```diff
@@ -10,7 +10,7 @@ module StandardFilters
       '"'.freeze => '&quot;'.freeze,
       "'".freeze => '&#39;'.freeze
     }
-    HTML_ESCAPE_ONCE_REGEXP = /["><']|&(?!([a-zA-Z]+|(#\d+));)/
+    HTML_ESCAPE_ONCE_REGEXP = /["><']|&(?!([a-zA-Z]+\d*|(#\d+)|(#[xX][\da-fA-F]+));)/
```
fw42: [This](https://github.com/rails/rails/blob/60257141462137331387d0e34931555cf0720886/activesupport/lib/active_support/core_ext/string/output_safety.rb#L8) is what Rails uses. I think we should probably go with that?

fw42:
```ruby
HTML_ESCAPE_ONCE_REGEXP = /["><']|&(?!([a-zA-Z]+|(#\d+)|(#[xX][\dA-Fa-f]+));)/
```

kainjow: No because that doesn't support named entities that contain letters.

fw42: I feel like we should probably stick to what Rails does, because that's what most people (at least in our use case) would expect. Do you think this should be fixed in Rails as well?

kainjow: Just because they both are written in Ruby doesn't mean they should have the same bugs... Also I use Jekyll, so I'd assume most Jekyll users don't care about what Rails does. Neither would Shopify customers. Seems like no one's even using these entities anyways, based on the fact the bug is still there in master.

Similarly I found a [bug](https://github.com/textmate/html.tmbundle/commit/9f812c89f4990a98391701caa77824c94860538f) with TextMate where it wasn't recognizing uppercase hex entities. I guess no one really tests their entities regexes fully :)

The reason I found this in the first place is because I'm [experimenting](https://github.com/kainjow/Jeqyll) with my own port of Liquid/Jekyll.

fw42: > doesn't mean they should have the same bugs

I agree with that, but I'm not convinced that it's a bug. If it's just a behaviour that you would prefer, then I would say no, lets do what Rails does. If it's actually a bug, then I think it should be fixed in both Liquid and Rails (not necessarily by yourself).

@rafaelfranca, what do you think?

rafaelfranca: Yeah, totally agree. If named entities that contains number are allowed in the HTML spec we should fix both Liquid and Rails.

PR #797, opened 2016-09-09
Standard filter truncate / truncatewords: force truncate_string to string

t6d:
Currently, `truncatewords` raises a TypeError when the argument `truncate_string` is an integer. This PR forces string coercion for any value provided for this argument. Thus,

```ruby
assert_equal 'one two1', @filters.truncatewords("one two three", 2, 1)
```

holds true.

The standard filter `truncate` suffered from a similar problem, resulting in a `NoMethodError` when an integer was supplied as an argument to `truncate_string`. Thus, string coercion is forced in this instance, as well.

@Thibaut, please review.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -65,9 +65,10 @@ module Liquid
       return if input.nil?
       input_str = input.to_s
       length = Utils.to_integer(length)
-      l = length - truncate_string.length
+      truncate_string_str = truncate_string.to_s
+      l = length - truncate_string_str.length
       l = 0 if l < 0
-      input_str.length > length ? input_str[0...l] + truncate_string : input_str
+      input_str.length > length ? input_str[0...l] + truncate_string_str : input_str
     end
 
     def truncatewords(input, words = 15, truncate_string = "...".freeze)
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -76,7 +76,7 @@ module Liquid
       words = Utils.to_integer(words)
       l = words - 1
       l = 0 if l < 0
-      wordlist.length > l ? wordlist[0..l].join(" ".freeze) + truncate_string : input
+      wordlist.length > l ? wordlist[0..l].join(" ".freeze) + truncate_string.to_s : input
     end
 
     # Split input string into an array of substrings separated by given pattern.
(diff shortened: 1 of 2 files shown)


kainjow:
Why would you want to pass anything but a string here? I feel the way Liquid handles this currently by throwing a type error is correct.

t6d:
I agree in the sense that passing a number doesn't make much sense, but I disagree as to what should happen when a user accidentally choose to do so. Liquid is often exposed to the user of an internet application and as such should be as resilient as possible. Throwing a low level `TypeError` is in my opinion not a good idea. Developers might expect and thus handle `Liquid::Error` and subclasses correctly, but they might not guard against `TyperErrors`. We should at least repack the error in one of Liquid's official error classes.

kainjow:
If that general rule should be applied, then shouldn't all filters be updated to allow numbers as strings, and vice versa? Why single out this one filter? At a minimum also patch the `truncate` filter to stay consistent.

t6d:
@kainjow, I agree. I'll patch the `truncate` filter accordingly.

Thibaut:
I think `to_s` makes sense rather than `Liquid::ArgumentError`. Liquid generally prefers coercion to strict type checking.

@kainjow numbers can already be passed as strings (see `Utils.to_integer`) and every string filter calls `to_s` on its input. You're right that we're never going to handle every single combination of type/input, but I don't see the harm in fixing cases we see in the wild when the fix doesn't make the code more complex.

👍 to patching `truncate` too.

t6d:
@Thibaut, I patched `truncate` in a similar fashion to `truncate_words`. Feel free to take another look.

lib/liquid/standardfilters.rb
```diff
@@ -65,9 +65,10 @@ def truncate(input, length = 50, truncate_string = "...".freeze)
       return if input.nil?
       input_str = input.to_s
       length = Utils.to_integer(length)
-      l = length - truncate_string.length
+      truncate_string_str = truncate_string.to_s
```
t6d: I'm not sure if introducing a new variable is actually necessary. This simply follows what was done for `input_str`. I personally don't have a problem with converting and reassigning the value of a variable even if it means a type change.

dylanahsmith:
:+1:

Thibaut:
:+1:

PR #798, opened 2016-09-10
Fixing #697 with better exception

### LINKED ISSUE #697: The "include" tag raises Ruby error when non existing variable is passed as attribute
The `include` tag raises Ruby error on render when non existing variable is passed as attribute.

Example code:

```ruby
{% assign foo = "my-partial" %}
{% include bar %}
```

Returns error:

```
undefined method `split' for nil:NilClass
liquid (4.0.0.rc1) lib/liquid/tags/include.rb:47:in `render'
liquid (4.0.0.rc1) lib/liquid/block_body.rb:90:in `render_node'
```

At the moment there is line:

```ruby
context_variable_name = template_name.split('/'.freeze).last
```

It seems that `to_s` should be enough here. Am I right?

```ruby
context_variable_name = template_name.to_s.split('/'.freeze).last
```

PhilibertDugas:
Fix #697 

When including a template which is not defined, the exception raised is 
>undefined method `split` for nil:NilClass

This occurs for a scenario like the following:
`{% include nil %}`
or
`{% include undefined-var %}`

Making the code raise a syntax error to allow better understanding of the include error

### DIFF
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -42,8 +42,9 @@ module Liquid
 
     def render(context)
       template_name = context.evaluate(@template_name_expr)
-      partial = load_cached_partial(template_name, context)
+      raise ArgumentError.new(options[:locale].t("errors.argument.include")) unless template_name
 
+      partial = load_cached_partial(template_name, context)
       context_variable_name = template_name.split('/'.freeze).last
 
       variable = if @variable_name_expr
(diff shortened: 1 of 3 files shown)


lib/liquid/tags/include.rb
```diff
@@ -42,8 +42,9 @@ def parse(_tokens)
 
     def render(context)
       template_name = context.evaluate(@template_name_expr)
-      partial = load_cached_partial(template_name, context)
+      raise SyntaxError.new(options[:locale].t("errors.syntax.include")) if template_name.nil?
```
Thibaut: I think `ArgumentError` would be more appropriate here. Something like:
```ruby
raise ArgumentError.new("illegal template name") unless template_name
```

PhilibertDugas: I've updated the PR with an `ArgumentError` instead of `SyntaxError`

Thibaut:
LGTM (maybe squash)
cc @fw42

PhilibertDugas:
Squashed the commits

PR #805, opened 2016-10-03
dont explode when sorting nil property

richardmonette:
Potential fix for https://github.com/Shopify/shopify/issues/84192

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -125,7 +125,15 @@ module Liquid
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
       elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
-        ary.sort { |a, b| a[property] <=> b[property] }
+        ary.sort do |a, b|
+          a = a[property]
+          b = b[property]
+          if a && b
+            a <=> b
+          else
+            a ? -1 : 1
+          end
+        end
       end
     end
 
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -178,6 +178,24 @@ class StandardFiltersTest < Minitest::Test
     assert_equal [{ "a" => 1 }, { "a" => 2 }, { "a" => 3 }, { "a" => 4 }], @filters.sort([{ "a" => 4 }, { "a" => 3 }, { "a" => 1 }, { "a" => 2 }], "a")
   end
 
+  def test_sort_when_property_is_sometimes_missing_puts_nils_last
+    input = [
+      { "price" => 4, "handle" => "alpha" },
+      { "handle" => "beta" },
+      { "price" => 1, "handle" => "gamma" },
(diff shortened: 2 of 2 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -125,7 +125,7 @@ def sort(input, property = nil)
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
       elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
-        ary.sort { |a, b| a[property] <=> b[property] }
+        ary.sort { |a, b| a[property] && b[property] ? a[property] <=> b[property] : a[property] ? -1 : 1 }
```
fw42: this line is way to confusing... can we add some parentheses?

fw42:
can we have a test to assert the behaviour of comparing nils to other nils? what should happen in that case?

lib/liquid/standardfilters.rb
richardmonette: I can break it up across a few lines instead of using compound ternary statements. Basically, its just saying, if we have values for both properties, then use the combined comparison operator, otherwise, if the first element has the value, its considered larger.

richardmonette:
@fw42 Can add a test but, what would the expectation be? In my opinion, when sorting, and some values are nil, the sortable values should come first in order, and then the nils, in the case of ascending. The nils would be considered the lowest possible value. Order amongst the nil values would not be guaranteed, as there is no sortable value.

richardmonette:
If we are not comfortable with having the nils portion of the sorted array be non-deterministic, I would propose we extend the `ary.first.respond_to?(:[]) && !ary.first[property].nil?` check to make sure all values pass this check, otherwise we raise an error saying the input cannot be sorted on the given property.

richardmonette:
cc @pushrax @dylanahsmith

pushrax:
>  In my opinion, when sorting, and some values are nil, the sortable values should come first in order, and then the nils, in the case of ascending. The nils would be considered the lowest possible value.

isn't this contradictory?

lib/liquid/standardfilters.rb
pushrax: It would also be nice to avoid evaluating `[]` multiple times. 👍  to breaking it up into multiple lines, and also caching the value of `[property]`.

richardmonette:
@pushrax Sorry, my explanation might not be worded as clearly as possible. I've updated the test to try and better convey the intent of the change. 

The example is something like, say we have 3 search results, 2 ProductDrops and 1 ArticleDrop, a given sorting property of 'price'. 

It seems a reasonable result would be something like, the cheapest product first, then the second cheapest, then the ArticleDrop, where the price cannot be determined.

As it stands, this would just caused `ArgumentError: comparison of Hash with Hash failed` which doesn't seem great.

fw42:
I'm ok with this. If @pushrax is too, then 👍 🚢

pushrax:
:+1:

PR #809, opened 2016-10-21
introduce unhandled liquid exception

richardmonette:
The goal of this change is to make it easier to debug exceptions which are generated in the process of rendering liquid.

As it stands, when a liquid error such as `UndefinedDropMethod` is raised, it is augmented with the template name and line number, which aids in debugging the error.

However, for so called "unhandled" errors we do not add this information, which makes debugging errors such as TypeError more challenging.

This PR improves the situation in a couple ways:

Firstly, we add the actual line in question, as an example this might contain "paginate collection.products by test".

Secondly, we wrap StandardErrors in an UnhandledError class, which can carry this additional information (which can use then used in Shopify core, for example, when we report these exceptions.)

@dylanahsmith @pushmatrix @fw42 cc @Thibaut

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -93,10 +93,10 @@ module Liquid
         rescue MemoryError => e
           raise e
         rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
-          context.handle_error(e, token.line_number)
+          context.handle_error(e, token.line_number, token.raw)
           output << nil
         rescue ::StandardError => e
-          output << context.handle_error(e, token.line_number)
+          output << context.handle_error(e, token.line_number, token.raw)
         end
       end
 
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -73,7 +73,7 @@ module Liquid
       @interrupts.pop
     end
 
-    def handle_error(e, line_number = nil)
+    def handle_error(e, line_number = nil, raw_token = nil)
       if e.is_a?(Liquid::Error)
         e.template_name ||= template_name
         e.line_number ||= line_number
@@ -82,7 +82,9 @@ module Liquid
       output = nil
 
       if exception_handler
-        result = exception_handler.call(e)
+        args = [e]
+        args << { line_number: line_number, raw_token: raw_token } if exception_handler.arity == 2
+        result = exception_handler.call(*args)
         case result
         when Exception
           e = result


lib/liquid/block_body.rb
```diff
@@ -93,10 +93,10 @@ def render(context)
         rescue MemoryError => e
           raise e
         rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
-          context.handle_error(e, token.line_number)
+          context.handle_error(e, token.line_number, token.raw)
           output << nil
         rescue ::StandardError => e
-          output << context.handle_error(e, token.line_number)
+          output << context.handle_error(UnhandledError.new(e), token.line_number, token.raw)
```
richardmonette: @wvanbergen Was this kind of what you were thinking, in terms of wrapping the error? My thought was to wrap the error, but still as a different type, so we'd be able to differentiate between error which arise in rendering the liquid vs. something internal that happened in the code which is involved in rendering the liquid cc @dylanahsmith  https://github.com/Shopify/shopify/issues/83649#issuecomment-252432948

lib/liquid/context.rb
```diff
@@ -73,10 +73,11 @@ def pop_interrupt
       @interrupts.pop
     end
 
-    def handle_error(e, line_number = nil)
-      if e.is_a?(Liquid::Error)
+    def handle_error(e, line_number = nil, line = nil)
```
richardmonette: Ah sorry, line is a bad variable name, I was wanting that to be line, as in the raw liquid code on that line.

lib/liquid/block_body.rb
dylanahsmith: :-1: to what he was suggesting there, which is basically ignoring the bug in liquid by treating it as a bug in the template

richardmonette:
Updated to pass the args through the exception handler, with an arity check. Little bit more context on what this achieves:

Say I make an argument error

`{{ 'fall 2016 collection' | split }}`

then I get 

`Liquid error (line 16): wrong number of arguments (1 for 2)`

now, if there is an internal error, we can format it like

`Liquid error (line 16): internal`

where we have the advantage of showing which line the error is on.

Gemfile
```diff
@@ -7,6 +7,7 @@ group :test do
   gem 'spy', '0.4.1'
   gem 'benchmark-ips'
   gem 'rubocop', '0.34.2'
+  gem 'byebug'
```
richardmonette: will 🔥 before shipping

lib/liquid/context.rb
```diff
@@ -1,3 +1,5 @@
+require 'byebug'
```
richardmonette: will 🔥 before shipping

lib/liquid/context.rb
```diff
@@ -73,7 +73,7 @@ def pop_interrupt
       @interrupts.pop
     end
 
-    def handle_error(e, line_number = nil)
+    def handle_error(e, line_number = nil, line = nil)
```
richardmonette: raw_token?

pushrax: Yeah that's pretty clear.

richardmonette: Updated!

lib/liquid/context.rb
```diff
@@ -82,7 +82,11 @@ def handle_error(e, line_number = nil)
       output = nil
 
       if exception_handler
-        result = exception_handler.call(e)
+        exception_handler_arity = exception_handler.arity
+        args = [e]
+        args << line_number if exception_handler_arity > 1
+        args << line if exception_handler_arity > 2
+        result = exception_handler.call(*args)
```
pushrax: How about we pass a hash as the second argument?

richardmonette: Ah, excellent idea, saves us if we want to add more extra data later, than having to mess around with an additional arg, will do.

PR #810, opened 2016-10-22
Document how to get the current time

knu:
This must be documented because it is a special feature implemented in the `date` filter, not something directly supported by the time parser.


pushrax:
Looks good, thanks. Keep in mind that if you use this on Shopify or Github pages, the value will be fixed until the template or input data is updated due to caching, but sometimes it's close enough if you're just trying to insert a copyright year or something like that.

Could you add a brief note about that to the doc page?

knu:
@pushrax Good point, I updated the example and description.  Feel free to fix wording before merging, thanks!

Issue #202, opened 2013-06-07
using regex for filter's argument in liquid for designer  


jojosati:
Could liquid parse some filter's argument as reqex instead of normal string ?

normal string
{{ 'foofoo' | replace:'foo','bar' }} #=> 'barbar'

regex pattern
{{'foofo' | replace:/foo.*/,'bar'}} #=> 'bar'

sathit

PR #812, opened 2016-10-26
allow split to accept numeric

richardmonette:
Attempt to coerce numeric values to string such that `'Fall 2016 Collection' | split: 2016` doesn't result in `wrong argument type Fixnum (expected Regexp)` The type check on `Numeric` is so that we don't convert, for example, a regex input to its `to_s` representation.

@dylanahsmith @pushrax

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -86,7 +86,7 @@ module Liquid
     #   <div class="summary">{{ post | split '//' | first }}</div>
     #
     def split(input, pattern)
-      input.to_s.split(pattern)
+      input.to_s.split(pattern.to_s)
     end
 
     def strip(input)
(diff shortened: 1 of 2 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -86,6 +86,7 @@ def truncatewords(input, words = 15, truncate_string = "...".freeze)
     #   <div class="summary">{{ post | split '//' | first }}</div>
     #
     def split(input, pattern)
+      pattern = pattern.to_s if pattern.is_a?(Numeric)
```
dylanahsmith: What about for non-string values other than Numeric?  Should we raise a Liquid::Error for that so that it doesn't result in an internal error?  Maybe we should have a Liquid::Utils.to_string method to make that pattern easy.

dylanahsmith: I didn't think we allowed a regex to be created from liquid

richardmonette: @pushrax my first thought was just to `to_s` the pattern, which is what we seem to do with most other filters, however this test https://github.com/Shopify/liquid/blob/master/test/integration/standard_filter_test.rb#L126 will fail then. @dylanahsmith if Liquid doesn't support creating a Regex anyways, maybe this test is bogus, we should just remove it and then use the `to_s` approach?

richardmonette: Looks like we've said Liquid regex is not supported, so I'd vote we just remove that test and to_s https://github.com/Shopify/liquid/issues/202#issuecomment-19112872

pushrax: Nice test – "thing works even though it's impossible to use it"

pushrax: 👍 to deleting that test.

PR #813, opened 2016-10-26
Update iteration.md

kainjow:
Fix liquid syntax error

Liquid Warning: Liquid syntax error (line 111): Unexpected character   in "i in (3..5) " in tags/iteration.md


kainjow:
Fixes the range section not displaying properly. Here's what it currently looks like at http://shopify.github.io/liquid/tags/iteration/

screen shot 2016-10-26 at 2 15 08 pm

pushrax:
Thanks!

PR #817, opened 2016-10-31
rename raw to raw liquid

richardmonette:
Straight rename of `raw` to `raw_liquid`.

Q: Why?
A: In Shopify core our definition of `Form < Liquid::Block` also includes `include ActionView::Helpers::TagHelper`, which in turn pulls in `require 'active_support/core_ext/string/output_safety'` https://github.com/rails/rails/blob/8f544bc447612924a50c37ead085a0ea4c217439/actionview/lib/action_view/helpers/tag_helper.rb#L1 Unfortunately, `OutputSafetyHelper` also defines `raw` which overwrites the definition https://github.com/rails/rails/blob/8f544bc447612924a50c37ead085a0ea4c217439/actionview/lib/action_view/helpers/output_safety_helper.rb#L16

```
(byebug) Form.ancestors
[Form, ActionView::Helpers::TagHelper, ActionView::Helpers::OutputSafetyHelper, ActionView::Helpers::CaptureHelper, Liquid::Block, Liquid::Tag, Liquid::ParserSwitching, NakayoshiFork, Object, Minitest::Expectations, Metaclass::ObjectMethods, Mocha::ObjectMethods, PP::ObjectMixin, Nori::CoreExt::Object, ERB::Util, ActiveSupport::Dependencies::Loadable, JSON::Ext::Generator::GeneratorMethods::Object, Kernel, BasicObject]
```

Defined first (correctly) in `Liquid::Tag` and the overwritten with the wrong code in `ActionView::Helpers::OutputSafetyHelper`

In order to be able to start using https://github.com/Shopify/liquid/pull/809 in Shopify production, we need to rename this method, so the conflict doesn't trigger some of the test suite to fail. 

@rafaelfranca @pushrax @dylanahsmith cc @Thibaut

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -93,10 +93,10 @@ module Liquid
         rescue MemoryError => e
           raise e
         rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
-          context.handle_error(e, token.line_number, token.raw)
+          context.handle_error(e, token.line_number, token.raw_liquid)
           output << nil
         rescue ::StandardError => e
-          output << context.handle_error(e, token.line_number, token.raw)
+          output << context.handle_error(e, token.line_number, token.raw_liquid)
         end
       end
 
--- a/lib/liquid/profiler.rb
+++ b/lib/liquid/profiler.rb
@@ -47,7 +47,7 @@ module Liquid
       attr_reader :code, :partial, :line_number, :children
 
       def initialize(node, partial)
-        @code        = node.respond_to?(:raw) ? node.raw : node
+        @code        = node.respond_to?(:raw_liquid) ? node.raw_liquid : node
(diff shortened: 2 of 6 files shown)


dylanahsmith:
Instead of doing

```ruby
class Form < Liquid::Block
  include ActionView::Helpers::TagHelper
```

for the form tag, why not use

```ruby
class Form < Liquid::Block
  module TagHelper
    extend ActionView::Helpers::TagHelper
  end
  private_constant :TagHelper
```

then prefix TagHelper method calls with TagHelper (e.g. replace `tag` with `TagHelper.tag` and `tag_content` with `TagHelper.content_tag`)

This way the code is clearer, we are less likely to encounter conflicts in the future and we don't need a breaking change to liquid.

richardmonette:
👍

PR #829, opened 2016-11-28
828 fix content variable error

### LINKED ISSUE #828: Include fails if strict_variables
When a template is rendered with `strict_variables: true` and the template includes a simple `include` like so:

```liquid
{% include "some_temploate" %}
```

a Liquid::UndefinedVariable is raised.

This is because of the "default name" feature where a variable is automatically assigned to the included template if the variable has the same name as the template.

Am I right that it should be possible in this case to look up a variable but now raise an error if not found and `:strict_variables` is set?
If so I will be happy to provide a PR.

lasseebert:
This fixes #828 

The test describes the wanted behaviour of not adding an error when including a template when using `strict_variables: true`

### DIFF
--- a/test/integration/tags/include_tag_test.rb
+++ b/test/integration/tags/include_tag_test.rb
@@ -235,4 +235,11 @@ class IncludeTagTest < Minitest::Test
 
     assert_template_result "Product: Draft 151cm ", "{% assign page = 'product' %}{% include page for foo %}", "foo" => { 'title' => 'Draft 151cm' }
   end
+
+  def test_including_with_strict_variables
+    template = Liquid::Template.parse("{% include 'simple' %}", error_mode: :warn)
+    template.render(nil, strict_variables: true)
+
+    assert_equal [], template.errors
+  end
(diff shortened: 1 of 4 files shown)


pushrax:
LGTM cc @dylanahsmith @fw42

test/integration/tags/include_tag_test.rb
```diff
@@ -235,4 +235,11 @@ def test_including_via_variable_value
 
     assert_template_result "Product: Draft 151cm ", "{% assign page = 'product' %}{% include page for foo %}", "foo" => { 'title' => 'Draft 151cm' }
   end
+
+  def test_including_with_strict_variables
+    template = Liquid::Template.parse("{% include 'simple' %}", error_mode: :warn)
```
fw42: Does this template ("simple") exist?

lasseebert: No, it does not need to exists to show the issue under test. In fact, AFAIR I added the template but removed it again before comitting, since it was not needed here.

PR #832, opened 2016-12-02
Drop support for ruby 2.0

dylanahsmith:
@fw42 & @pushrax please review

https://www.ruby-lang.org/en/downloads/ says ruby 2.0 is no longer maintained, so I don't think we should support it for liquid v4.

## Motivation

ruby 2.1 added Exception#cause which I would like to assume it available for a PR I'm going to open up to add a Liquid::InternalError that will wrap non-Liquid::Error exceptions.  This way we can attach the template_name and line_number to the error and still have the original exception available through the cause method for error reporting and logging.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 
 rvm:
-  - 2.0
   - 2.1
   - 2.2
   - ruby-head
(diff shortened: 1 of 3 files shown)


kainjow:
You'll probably want to add notes for macOS users on how to upgrade Ruby since Sierra still ships with 2.0.

dylanahsmith:
I don't think it is generally recommended to install gems into the system ruby anyways.  It can't be upgraded because parts of the system probably depend on that version.  For reference, rails requires ruby version `>= 2.2.2`.

PR #834, opened 2016-12-05
Add a filter to compute percentages

ashmaroli:
a filter to compute what percentage `num A` is of `num B`
  - returns a float
  - returns "Infinity" when supplied with `0`
  - can be rounded off to required decimal places by the *round* filter
  - returns a float with `.0` even when the modulo of `input` and `operand` is `0` && even if `round: 2` is used

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -332,6 +332,12 @@ module Liquid
       raise Liquid::ZeroDivisionError, e.message
     end
 
+    # percentage
+    def percentage_of(input, operand)
+      floating_product = apply_operation(input, 100.0, :*)
+      apply_operation(floating_product, operand, :/)
+    end
+
     def round(input, n = 0)
       result = Utils.to_number(input).round(Utils.to_number(n))
       result = result.to_f if result.is_a?(BigDecimal)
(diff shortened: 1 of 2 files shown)


kainjow:
Why not return 0 when input is 0?

pushrax:
Is this filter really needed when you can do `{{ x | times: 100.0 | divided_by: y }}`?

> Why not return 0 when input is 0?

That would be unexpected IMO.

kainjow:
>  assert_template_result "33", "{{ 40 | percentage_of: 120 | round }}"

Maybe it's just too early in the morning, but isn't 40% of 120 48?

fw42:
I don't have a strong opinion on this... Seems unnecessary to have a custom filter for it, but if there's a strong use-case then I'm ok with adding it.

ashmaroli:
> Is this filter really needed when you can do `{{ x | times: 100.0 | divided_by: y }}`?

I expected this question to come up sometime..
yes, they both return the same answer as that's exactly how this filter works internally..

This filter is useful for those who are not used to the syntax above and tend to calculate percentage by dividing a number first and then multiplying by 100.
This filter shortens the liquid block, especially if multiple math filters had already been called on the *`input`* and should a need remain to use further filters ahead (e.g `round`)

PR #835, opened 2016-12-05
change: Render an opaque internal error by default for non-Liquid::Error

dylanahsmith:
@fw42 & @pushrax, @richardmonette please review

## Problem

When a non-Liquid::Error exception is caught and rendered by liquid, it can cause sensitive information to be rendered.  I opened https://github.com/Shopify/liquid/pull/576 to solve this problem at Shopify, but it would be safer to have the default behaviour be to not show the message from these non-Liquid::Errors.

## Solution

When a non-Liquid::Error occurs during liquid rendering, it is replaced with a Liquid::InternalError exception which has the original exception available through its [cause](http://ruby-doc.org/core-2.2.4/Exception.html#method-i-cause) method.  The Liquid::InternalError has a message 'internal' so will show up as `Liquid error: internal` in the rendered output.  This Liquid::InternalError exception is what is pushed to the Liquid::Context#errors array.

The Liquid::Context#exception_handler's API has changed quite a bit since the last release of liquid.  It used to return a truthy value to re-raise the exception and a falsey value to render the error, however, re-raising the exception is a clearer and more flexible way to handle that behaviour. On master it has become more complicated to attach metadata to the error, which isn't necessary now that a Liquid::InternalError can be passed to it.  Instead, I just kept the feature of allowing the handler to return a value to render for the error, since nothing else is needed.  I changed `exception_handler` to `exception_renderer` to remove support for the old behaviour while avoiding a subtle breaking change.  I think the new name is also clearer.

The old behaviour can be acheived with the following exception_handler:

```ruby
context.exception_renderer = lambda do |exc|
  exc.is_a?(Liquid::InternalError) ? "Liquid error: #{exc.cause.message}" : exc
end
```

## Future Work

I think it would also be useful to be able to change the default template error handler.  That way it would be easier to provide a consistent way of handling exceptions across an app, even when there are multiple places where liquid rendering is done.  For instance, to simplify upgrading from liquid v3, the default exception_renderer could be overridden with the liquid v3 compatible one mentioned above.  This would also be an ideal place to report internal errors to an error reporting system like errbit or bugsnag if they are not re-raised.

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -13,7 +13,7 @@ module Liquid
   #   context['bob']  #=> nil  class Context
   class Context
     attr_reader :scopes, :errors, :registers, :environments, :resource_limits
-    attr_accessor :exception_handler, :template_name, :partial, :global_filter, :strict_variables, :strict_filters
+    attr_accessor :exception_renderer, :template_name, :partial, :global_filter, :strict_variables, :strict_filters
 
     def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_errors = false, resource_limits = nil)
       @environments     = [environments].flatten
@@ -28,7 +28,7 @@ module Liquid
       @this_stack_used = false
 
       if rethrow_errors
-        self.exception_handler = ->(e) { raise }
+        self.exception_renderer = ->(e) { raise }
       end
 
       @interrupts = []
@@ -74,32 +74,13 @@ module Liquid
     end
 
     def handle_error(e, line_number = nil, raw_token = nil)
-      if e.is_a?(Liquid::Error)
-        e.template_name ||= template_name
-        e.line_number ||= line_number
-      end
-
-      output = nil
-
-      if exception_handler
-        args = [e]
-        args << { line_number: line_number, raw_token: raw_token } if exception_handler.arity == 2
-        result = exception_handler.call(*args)
-        case result
-        when Exception
-          e = result
-          if e.is_a?(Liquid::Error)
-            e.template_name ||= template_name
-            e.line_number ||= line_number
-          end
-        when String
-          output = result
-        else
-          raise if result
-        end
-      end
+      e = internal_error unless e.is_a?(Liquid::Error)
+      e.template_name ||= template_name
+      e.line_number ||= line_number
       errors.push(e)
-      output || Liquid::Error.render(e)
+
+      e = exception_renderer.call(e) if exception_renderer
+      e.to_s
     end
 
     def invoke(method, *args)
@@ -222,6 +203,13 @@ module Liquid
     end
 
     private
+
+    def internal_error
+      # raise and catch to set backtrace and cause on exception
+      raise Liquid::InternalError, 'internal'
+    rescue Liquid::InternalError => exc
+      exc
+    end
 
     def squash_instance_assigns_with_environments
       @scopes.last.each_key do |k|
--- a/History.md
+++ b/History.md
@@ -3,6 +3,7 @@
 ## 4.0.0 / not yet released / branch "master"
 
 ### Changed
+* Render an opaque internal error by default for non-Liquid::Error (#835) [Dylan Thacker-Smith]
 * Ruby 2.0 support dropped (#832) [Dylan Thacker-Smith]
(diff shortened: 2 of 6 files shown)


richardmonette:
```
if @rethrow_errors
  c.exception_handler = ->(e) { raise }
end
```

This seems like a very specific override, if we are changing this around a bit right now anyways, could we also deprecate/remove this? Instead of having a parameter and a flag, perhaps we can just suggest using the parameter, passing raise.

dylanahsmith:
That override is specific because it is used for `Liquid::Template#render!`.  I don't think it makes sense to make that usage more complicated

richardmonette:
> dylan.smith e.g. if you re-used a template and called `render!` then `render` then `@rethrow_errors` would affect the second render

Apart from above, as discussed on Slack, 👍 from me

lib/liquid/context.rb
```diff
@@ -223,6 +204,13 @@ def lookup_and_evaluate(obj, key)
 
     private
 
+    def internal_error
+      # raise and catch to set backtrace and cause on exception
```
fw42: Is this because `return Liquid::InternaError.new("internal")` would not have a backtrace? Maybe something like `#set_backtrace(caller)` would be easier to understand? Also, shouldn't the backtrace of this exception be the same backtrace as the one of the original exception (`e` above)?

lib/liquid/context.rb
```diff
@@ -74,32 +74,13 @@ def pop_interrupt
...
-      if e.is_a?(Liquid::Error)
-        e.template_name ||= template_name
-        e.line_number ||= line_number
-      end
-
-      output = nil
-
-      if exception_handler
-        args = [e]
-        args << { line_number: line_number, raw_token: raw_token } if exception_handler.arity == 2
```
fw42: Hmm is this logic not necessary anymore? Was this only a hack to get the line number and now you are attaching it to the exception itself (since everything is a `Liquid::Error` or `Liquid::InternalError` now)?

lib/liquid/context.rb
```diff
@@ -74,32 +74,13 @@ def pop_interrupt
...
-        when Exception
-          e = result
-          if e.is_a?(Liquid::Error)
-            e.template_name ||= template_name
-            e.line_number ||= line_number
-          end
-        when String
-          output = result
-        else
-          raise if result
```
fw42: This logic (don't raise if handler returned false) is also gone now. Was it not important?

dylanahsmith: I think it is clearer to raise in the error_renderer than to return a boolean, so I wanted to cleanup this code by not supporting this legacy behavior.

lib/liquid/context.rb
dylanahsmith: I didn't see a way to set the cause on exceptions without raising an exception, and that is needed to preserve the original exceptions information.  I think previously setting the backtrack was used as a workaround for not being able to set the cause, and doing that as well could cause more confusion

lib/liquid/context.rb
pushrax: This removes the token from the error which was added for debugging, can we add that back?

dylanahsmith: I feel like `raw_token` was added outside of the error for the same reason that `line_number` was added outside of the error, because we couldn't attach that information to the error.  We actually already have a Liquid::Error#markup_context which looks like it serves the same purpose.

The other problem that we have is that we automatically render information that is on the Error, which prevents us from getting line number information without displaying it.  We could override that in the liquid_renderer, but we have made it awkward to get the liquid error message without the context (e.g. `error.to_s(false)`).  Once we deal with that issue we could actually have the line number as well as the token, without exposing them to the world.

pushrax: Ahh yeah `markup_context` is what we really want, 👍

mhenrixon:
Can I globally use the below code 👇 somehow?

```ruby
context.exception_renderer = lambda do |exc|
  exc.is_a?(Liquid::InternalError) ? "Liquid error: #{exc.cause.message}" : exc
end
```

While I appreciate the security, we must get the necessary feedback somewhere to avoid drawing blanks while debugging. 

We previously reverted a liquid upgrade because of lack of information on errors thrown. Is there something like a global error handler or so that I can use to present the cause or some such?

PR #839, opened 2016-12-16
Remove jQuery

admhlt:
This is not very DRY, but it lets us ditch jQuery on the Liquid docs page.

cc @Tetsuro


Tetsuro:
Super sweet @admhlt !! Yeah, let's get rid of 'er. 

My only comment is that since the variables are no longer jQuery variables, we can get rid of the `$` in the variable names. Ex: `$menuButton` -> `menuButton`

admhlt:
I figured that was an option, but then I forgot to actually remove them! I'll do that before shipping.

PR #840, opened 2016-12-16
Update release date for Liquid 4.0.0

admhlt:
4.0.0 was released the other day, so we should update the history doc.


pushrax:
Ah did this right around the same time you opened the PR, going to close.

PR #849, opened 2017-01-16
Fix internal liquid error when comparing hash with incompatible type

### LINKED ISSUE #842: Comparation condition bug on type check
Reproduce:

``` ruby
t = Liquid::Template.parse("{% if a > b %} Yes {% endif %}")
t.render!({a: {}, b: 2}.with_indifferent_access)
```

It can be fixed by adding `TypeError` besides `::ArgumentError` to the rescue block.

dylanahsmith:
Fixes #842

ruby 2.3 implements comparison operators on hashes, but these result in a TypeError when the types don't match.  Previously we did a respond_to? check for the operators which would cause `nil` to be returned, so along with the respond_to? check I checked for hash objects to keep compatibility with the previous behaviour.

### DIFF
--- a/lib/liquid/condition.rb
+++ b/lib/liquid/condition.rb
@@ -110,7 +110,7 @@ module Liquid
 
       if operation.respond_to?(:call)
         operation.call(self, left, right)
-      elsif left.respond_to?(operation) && right.respond_to?(operation)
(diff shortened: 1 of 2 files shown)


fw42:
cc @csfrancis

bogdan:
What about hash equal to hash?

PR #850, opened 2017-01-16
test: Equality comparison of two hashes

dylanahsmith:
@bogdan is this test what you wanted in your comment https://github.com/Shopify/liquid/pull/849#issuecomment-272956299


bogdan:
Yes, thank you.

PR #851, opened 2017-01-17
Allow benchmarks to benchmark render by itself

jerryliu55:
Currently the `benchmark.rb` under the `performance` directory runs 2 benchmarks:
* compile
* compile + render

This PR will allow it to run a benchmark on just the render.

### DIFF
--- a/performance/benchmark.rb
+++ b/performance/benchmark.rb
@@ -5,7 +5,7 @@ Liquid::Template.error_mode = ARGV.first.to_sym if ARGV.first
 profiler = ThemeRunner.new
 
 Benchmark.ips do |x|
-  x.time = 60
+  x.time = 10
   x.warmup = 5
 
   puts
@@ -13,5 +13,6 @@ Benchmark.ips do |x|
   puts
 
   x.report("parse:") { profiler.compile }
-  x.report("parse & run:") { profiler.run }
+  x.report("render:") { profiler.render }
+  x.report("parse & render:") { profiler.run }
 end
--- a/performance/theme_runner.rb
+++ b/performance/theme_runner.rb
@@ -21,53 +21,100 @@ class ThemeRunner
     end
   end
 
-  # Load all templates into memory, do this now so that
-  # we don't profile IO.
+  # Initialize a new liquid ThemeRunner instance
+  # Will load all templates into memory, do this now so that we don't profile IO.
   def initialize
     @tests = Dir[__dir__ + '/tests/**/*.liquid'].collect do |test|
       next if File.basename(test) == 'theme.liquid'
 
       theme_path = File.dirname(test) + '/theme.liquid'
-
-      [File.read(test), (File.file?(theme_path) ? File.read(theme_path) : nil), test]
+      {
+        liquid: File.read(test),
+        layout: (File.file?(theme_path) ? File.read(theme_path) : nil),
+        template_name: test
+      }
     end.compact
+
+    compile_all_tests
   end
 
+  # `compile` will test just the compilation portion of liquid without any templates
   def compile
-    # Dup assigns because will make some changes to them
-
-    @tests.each do |liquid, layout, template_name|
-      tmpl = Liquid::Template.new
-      tmpl.parse(liquid)
-      tmpl = Liquid::Template.new
-      tmpl.parse(layout)
+    @tests.each do |test_hash|
+      Liquid::Template.new.parse(test_hash[:liquid])
+      Liquid::Template.new.parse(test_hash[:layout])
     end
   end
 
+  # `run` is called to benchmark rendering and compiling at the same time
   def run
-    # Dup assigns because will make some changes to them
-    assigns = Database.tables.dup
-
-    @tests.each do |liquid, layout, template_name|
-      # Compute page_tempalte outside of profiler run, uninteresting to profiler
-      page_template = File.basename(template_name, File.extname(template_name))
+    each_test do |liquid, layout, assigns, page_template, template_name|
       compile_and_render(liquid, layout, assigns, page_template, template_name)
     end
   end
 
+  # `render` is called to benchmark just the render portion of liquid
+  def render
+    @compiled_tests.each do |test|
+      tmpl = test[:tmpl]
+      assigns = test[:assigns]
+      layout = test[:layout]
+
+      if layout
+        assigns['content_for_layout'] = tmpl.render!(assigns)
+        layout.render!(assigns)
+      else
+        tmpl.render!(assigns)
+      end
+    end
+  end
+
+  private
+
   def compile_and_render(template, layout, assigns, page_template, template_file)
+    compiled_test = compile_test(template, layout, assigns, page_template, template_file)
+    assigns['content_for_layout'] = compiled_test[:tmpl].render!(assigns)
+    compiled_test[:layout].render!(assigns) if layout
+  end
+
+  def compile_all_tests
+    @compiled_tests = []
+    each_test do |liquid, layout, assigns, page_template, template_name|
+      @compiled_tests << compile_test(liquid, layout, assigns, page_template, template_name)
+    end
+    @compiled_tests
+  end
+
+  def compile_test(template, layout, assigns, page_template, template_file)
+    tmpl = init_template(page_template, template_file)
+    parsed_template = tmpl.parse(template).dup
+
+    if layout
+      parsed_layout = tmpl.parse(layout)
+      { tmpl: parsed_template, assigns: assigns, layout: parsed_layout }
+    else
+      { tmpl: parsed_template, assigns: assigns }
+    end
+  end
+
+  # utility method with similar functionality needed in `compile_all_tests` and `run`
+  def each_test
+    # Dup assigns because will make some changes to them
+    assigns = Database.tables.dup
+
+    @tests.each do |test_hash|
+      # Compute page_template outside of profiler run, uninteresting to profiler
+      page_template = File.basename(test_hash[:template_name], File.extname(test_hash[:template_name]))
+      yield(test_hash[:liquid], test_hash[:layout], assigns, page_template, test_hash[:template_name])
+    end
+  end
+
+  # set up a new Liquid::Template object for use in `compile_and_render` and `compile_test`
+  def init_template(page_template, template_file)
     tmpl = Liquid::Template.new
     tmpl.assigns['page_title'] = 'Page title'
     tmpl.assigns['template'] = page_template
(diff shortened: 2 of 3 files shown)


performance/benchmark.rb
```diff
@@ -4,14 +4,19 @@
 Liquid::Template.error_mode = ARGV.first.to_sym if ARGV.first
 profiler = ThemeRunner.new
 
+# this profiler is for benchmarking render only
+profiler_render = ThemeRunner.new
+profiler_render.compile_tests
```
burke: ok, so I see what you're getting at here: there are two phases, parse/compile (A) and run/render (B). It does feel weird that our benchmarks here are {A, A+B}. It does make more sense to have {A, B} or {A, B, A+B}.

I don't see any reason this `profiler_render` has to be a separate variable. Could we just add `profiler.compile_tests` after line 5? Or better yet, just call `compile_tests` from `ThemeRunner#initialize`?

performance/theme_runner.rb
```diff
@@ -70,4 +68,55 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+    parsed_layout = tmpl.parse(layout)
+
+    if layout
+      { 'tmpl' => parsed_template, 'assigns' => assigns, 'layout' => parsed_layout }
+    else
+      { 'tmpl' => parsed_template, 'assigns' => assigns }
+    end
+  end
+
+  def benchmark_render
```
burke: I'd call this `render` to be consistent with the others.

performance/theme_runner.rb
```diff
@@ -70,4 +68,55 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
       content_for_layout
     end
   end
+
+  # the following 3 methods are for when we want to benchmark just the 'render'
+
+  # compile_tests should be called before the benchmark is run, and benchmark_render
+  # should be called to to benchmark just the render
+
+  def compile_tests
```
burke: Probably move this to the bottom and mark it `private`.

performance/theme_runner.rb
```diff
@@ -70,4 +68,55 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+
+    @compiled_tests = []
+    @tests.each do |liquid, layout, template_name|
+      # Compute page_tempalte outside of profiler run, uninteresting to profiler
+      page_template = File.basename(template_name, File.extname(template_name))
+      @compiled_tests << compile_and_save(liquid, layout, assigns, page_template, template_name)
+    end
+  end
+
+  def compile_and_save(template, layout, assigns, page_template, template_file)
```
burke: `private` here too

performance/theme_runner.rb
```diff
@@ -70,4 +68,55 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+    tmpl.assigns['template'] = page_template
+    tmpl.registers[:file_system] = ThemeRunner::FileSystem.new(File.dirname(template_file))
+
+    # in order to follow the order of parse template, render, parse layout, render
+    parsed_template = tmpl.parse(template).dup
+    tmpl.render!(assigns)
+    parsed_layout = tmpl.parse(layout)
+
+    if layout
+      { 'tmpl' => parsed_template, 'assigns' => assigns, 'layout' => parsed_layout }
```
burke: better to write this as:

```ruby
{ tmpl: parsed_template, assigns: assigns, layout: parsed_layout }
```

then access later as `test[:tmpl]`

performance/theme_runner.rb
```diff
@@ -70,4 +68,55 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+  def benchmark_render
+    @compiled_tests.each do |test|
+      # layout could be nil
+      tmpl = test['tmpl']
+      assigns = test['assigns']
+      layout = test['layout']
+
+      if layout
+        assigns['content_for_layout'] = tmpl.render!(assigns)
+        tmpl.render!(assigns)
```
burke: I think one or more of these is supposed to use `layout.render!`

jerryliu55:
@burke made the changes, please take a look

performance/benchmark.rb
```diff
@@ -5,13 +5,14 @@
 profiler = ThemeRunner.new
 
 Benchmark.ips do |x|
-  x.time = 60
+  x.time = 10
```
fw42: why change the time?

performance/theme_runner.rb
```diff
@@ -70,4 +70,57 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
   end
+
+  # the following 3 methods are for when we want to benchmark just the 'render'
+
+  # compile_tests is called in `initialize` to compile,
+  # render is called to benchmark just the render
+
+  def render
+    @compiled_tests.each do |test|
+      # layout could be nil
```
fw42: does this comment really add value?

performance/theme_runner.rb
```diff
@@ -70,4 +70,57 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+
+  private
+
+  def compile_tests
+    # Dup assigns because will make some changes to them
+    assigns = Database.tables.dup
+
+    @compiled_tests = []
+    @tests.each do |liquid, layout, template_name|
+      # Compute page_tempalte outside of profiler run, uninteresting to profiler
```
fw42: typo here ("tempalte" vs. "template")

performance/theme_runner.rb
```diff
@@ -70,4 +70,57 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+  end
+
+  private
+
+  def compile_tests
+    # Dup assigns because will make some changes to them
+    assigns = Database.tables.dup
+
+    @compiled_tests = []
+    @tests.each do |liquid, layout, template_name|
```
fw42: The following lines here look very similar to what the `run` method does. Can this be DRYed up somehow?

performance/theme_runner.rb
fw42: This also looks very similar to `compile_and_render`. Is the duplication really necessary?

performance/benchmark.rb
jerryliu55: the slowest benchmark iterates 150 times in 10 seconds so I thought that was representative, I can change it back if you want

performance/theme_runner.rb
jerryliu55: a version of compile that didn't render after compiling but instead saves the compilation results would be useful, so that's why I have `compile_and_save`. If we were to remove `compile_and_render`, that means changing the `run` benchmark to call `compile_all_tests` followed by `render` - would that be ok?

performance/theme_runner.rb
```diff
@@ -70,4 +61,66 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+  private
+
+  def compile_all_tests
+    @compiled_tests = []
+    each_test do |liquid, layout, assigns, page_template, template_name|
+      @compiled_tests << compile_and_save(liquid, layout, assigns, page_template, template_name)
+    end
+  end
+
+  def compile_and_save(template, layout, assigns, page_template, template_file)
```
jules2689: I'd name this `compile_test`

performance/theme_runner.rb
```diff
@@ -70,4 +61,66 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+    tmpl = init_template(page_template, template_file)
+    parsed_template = tmpl.parse(template).dup
+    parsed_layout = tmpl.parse(layout)
```
jules2689: move this to the if block

performance/theme_runner.rb
```diff
@@ -70,4 +61,66 @@ def compile_and_render(template, layout, assigns, page_template, template_file)
...
+    if layout
+      { tmpl: parsed_template, assigns: assigns, layout: parsed_layout }
+    else
+      { tmpl: parsed_template, assigns: assigns }
+    end
+  end
+
+  # utility method with similar functionality needed in `compile_all_tests` and `run`
+  def each_test
+    # Dup assigns because will make some changes to them
```
jules2689: This sentence is hard to read, can you clarify its meaning?

performance/theme_runner.rb
```diff
@@ -45,22 +45,13 @@ def compile
...
     end
   end
 
   def compile_and_render(template, layout, assigns, page_template, template_file)
-    tmpl = Liquid::Template.new
-    tmpl.assigns['page_title'] = 'Page title'
-    tmpl.assigns['template'] = page_template
-    tmpl.registers[:file_system] = ThemeRunner::FileSystem.new(File.dirname(template_file))
-
+    tmpl = init_template(page_template, template_file)
```
jules2689: This could be
```ruby
  def compile_and_render(template, layout, assigns, page_template, template_file)
    compiled_test = compile(template, layout, assigns, page_template, template_file)

    if layout
      assigns['content_for_layout'] = compiled_test[:tmpl].render!(assigns)
      compiled_test[:layout].render!(assigns)
    else
      compiled_test[:tmpl]
    end
  end
```
removing the duplication flo mentioned

performance/benchmark.rb
burke: I played with it a bit and it looks like 60s doesn't really generate more meaningful data than 10s, so +1 for this change IMO.

performance/theme_runner.rb
```diff
@@ -21,53 +21,104 @@ def read_template_file(template_path)
...
+        liquid: File.read(test),
+        layout: (File.file?(theme_path) ? File.read(theme_path) : nil),
+        template_name: test
+      }
     end.compact
+
+    compile_all_tests
   end
 
+  # <code>compile</code> will test just the compilation portion of liquid without any templates
```
burke: Unless this is style throughout the repo, use backtick or `+` to delimit code.

pushrax:
> ok, so I see what you're getting at here: there are two phases, parse/compile (A) and run/render (B). It does feel weird that our benchmarks here are {A, A+B}. It does make more sense to have {A, B} or {A, B, A+B}.

When a template uses `{% include %}` then when you call `render` the included file will be parsed. So {B} is really {A+B} anyway in that case. The templates we have in the benchmark suite right now don't use include though.

jules2689:
Are you suggesting adding a template that does that? @pushrax
  
  
  

  
  
>

burke:
you liquid people ok with merging this?

jules2689:
Flo was fine with it if we were when I talked to him fwiw

jules2689:
This should be fine. It doesn't affect functionality, and we can iterate more. So :+1: for merging now :)

PR #860, opened 2017-02-07
Avoid calling line_number on String node when rescuing a render error.

### LINKED ISSUE #853: NoMethodError: undefined method `line_number'
Hi,

While integrating unit tests to my project, I'm confronted to this error : 

`NoMethodError: undefined method line_number for "hi ":String`

I use README test :

```
@template = Liquid::Template.parse("hi {{name}}") # Parses and compiles the template
@template.render('name' => 'tobi')
```

I use liquid 4.0.0 and I use `Minitest::Test` and include Liquid into my tests. Rails version 4.2.5.1

Thank you.

dylanahsmith:
Fixes #853

## Problem

The backtrace in that issue shows ```NoMethodError: undefined method `line_number' for "hi ":String``` is happening on on line [lib/liquid/block_body.rb:99](https://github.com/Shopify/liquid/blob/v4.0.0/lib/liquid/block_body.rb#L99).  That means that something is causing a StandardError exception when rendering a String node, then the rescue block is calling `token.line_number` on the String token which causes a NoMethodError that hides the original exception.

## Solution

If the token is a `String`, then we don't know the line number, so I just passed `nil` as the line_number to Context#handle_error.

The third argument to Context#handle_error wasn't being used, so I removed it.

I'm not sure how to write a regression test for this.  I will need more information from @Apokly about what caused the original exception.  @Apokly could you try using this liquid branch and see what exception you end up getting, since that should be the exception that caused the rendering of the string node to fail.

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -107,7 +107,7 @@ module Liquid
     private
 
     def render_node(node, context)
-      node_output = (node.respond_to?(:render) ? node.render(context) : node)
+      node_output = node.is_a?(String) ? node : node.render(context)
       node_output = node_output.is_a?(Array) ? node_output.join : node_output.to_s
 
       context.resource_limits.render_length += node_output.length
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -93,10 +93,11 @@ module Liquid
         rescue MemoryError => e
           raise e
         rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
-          context.handle_error(e, token.line_number, token.raw)
+          context.handle_error(e, token.line_number)
           output << nil
         rescue ::StandardError => e
-          output << context.handle_error(e, token.line_number, token.raw)
+          line_number = token.is_a?(String) ? nil : token.line_number
+          output << context.handle_error(e, line_number)
         end
       end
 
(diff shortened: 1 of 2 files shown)


Apokly:
Thanks for PR. It's now working perfectly.
Btw, I had to use `render` instead of `render!` to render my template.
Here is the error with `error!` :

`... is not an ActiveModel-compatible object. It must implement :to_partial_path`

dylanahsmith:
It looks like a rails render method got called during liquid rendering.  There is a [Liquid::BlockBody#render_node](https://github.com/Shopify/liquid/blob/v4.0.0/lib/liquid/block_body.rb#L109) line that does `node.respond_to?(:render) ? node.render(context) : node`, but node should either be a String, Liquid::Variable or Liquid::Tag.  I'm guessing that something defined String#render and the implementation uses rails to render the string.

richardmonette:
Was mostly an experiment in terms of getting more info for working backwards on liquid errors that trickle in, feel free to remove if need be.

dylanahsmith:
> did you have plans for raw_token?

https://github.com/Shopify/liquid/pull/809 introduced it but that PR was mostly replaced by https://github.com/Shopify/liquid/pull/835, but I accidentally missed that raw_token argument

> In addition to this change, should render_node be changed to check that the type is Liquid::Variable or Liquid::Tag instead of checking for the existence of render?

Sure.  I'll change it to just check if the node is a String.

PR #864, opened 2017-02-20
fix  #861: duplicate inclusion condition logic error of Liquid::Strainer.add_filter method

### LINKED ISSUE #861: duplicate inclusion condition logic error of Liquid::Strainer.add_filter  method
In `Liquid::Strainer `class,  method `add_filter` has logic to avoid duplicate inclusion of the same filter module. But its judgement condition seems be wrong.

[ current codes](https://github.com/Shopify/liquid/blob/master/lib/liquid/strainer.rb#L30)
```ruby
def self.add_filter(filter)
      raise ArgumentError, "Expected module but got: #{filter.class}" unless filter.is_a?(Module)
      unless self.class.include?(filter)
        invokable_non_public_methods = (filter.private_instance_methods + filter.protected_instance_methods).select { |m| invokable?(m) }
        if invokable_non_public_methods.any?
          raise MethodOverrideError, "Filter overrides registered public methods as non public: #{invokable_non_public_methods.join(', ')}"
        else
          send(:include, filter)
          @filter_methods.merge(filter.public_instance_methods.map(&:to_s))
        end
      end
    end
```

In the 3th line ` unless self.class.include?(filter)>` , since method `add_filter` is a class method, the result of  ` self.class` is   not `Strainer` (or Strainer child class)  but `Class` , which is not we expect.  

[test case](https://github.com/chenxianyu2015/liquid/blob/aa79c33dda95b3b7ab2812e30db191a113763516/test/unit/strainer_unit_test.rb#L149)
```ruby
  def test_add_filter_returns_nil_when_a_filter_module_was_added
    a = Module.new
    strainer = Context.new.strainer
    result1 = strainer.class.add_filter(a)
    result2 = strainer.class.add_filter(a)
    assert_kind_of Set, result1
    assert_kind_of NilClass, result2
  end
```

[fix:](https://github.com/chenxianyu2015/liquid/blob/aa79c33dda95b3b7ab2812e30db191a113763516/lib/liquid/strainer.rb#L30)
```ruby
    def self.add_filter(filter)
      raise ArgumentError, "Expected module but got: #{filter.class}" unless filter.is_a?(Module)
      unless self.include?(filter)
        invokable_non_public_methods = (filter.private_instance_methods + filter.protected_instance_methods).select { |m| invokable?(m) }
        if invokable_non_public_methods.any?
          raise MethodOverrideError, "Filter overrides registered public methods as non public: #{invokable_non_public_methods.join(', ')}"
        else
          send(:include, filter)
          @filter_methods.merge(filter.public_instance_methods.map(&:to_s))
        end
      end
    end
```

chenxianyu2015:
fix  #861: duplicate inclusion condition logic error of Liquid::Strainer.add_filter method

### DIFF
--- a/test/unit/strainer_unit_test.rb
+++ b/test/unit/strainer_unit_test.rb
@@ -145,4 +145,13 @@ class StrainerUnitTest < Minitest::Test
     Strainer.global_filter(LateAddedFilter)
     assert_equal 'filtered', Strainer.create(nil).invoke('late_added_filter', 'input')
   end
+
+  def test_add_filter_returns_nil_when_a_filter_module_was_added
+    a = Module.new
+    strainer = Context.new.strainer
+    result1 = strainer.class.add_filter(a)
+    result2 = strainer.class.add_filter(a)
+    assert_kind_of Set, result1
+    assert_kind_of NilClass, result2
+  end
 end # StrainerTest
--- a/test/unit/strainer_unit_test.rb
+++ b/test/unit/strainer_unit_test.rb
@@ -146,12 +146,19 @@ class StrainerUnitTest < Minitest::Test
     assert_equal 'filtered', Strainer.create(nil).invoke('late_added_filter', 'input')
   end
 
-  def test_add_filter_returns_nil_when_a_filter_module_was_added
-    a = Module.new
+  def test_add_filter_does_not_include_already_included_module
+    mod = Module.new do
+      class << self
+        attr_accessor :include_count
+        def included(mod)
+          self.include_count += 1
+        end
+      end
+      self.include_count = 0
+    end
     strainer = Context.new.strainer
-    result1 = strainer.class.add_filter(a)
-    result2 = strainer.class.add_filter(a)
-    assert_kind_of Set, result1
-    assert_kind_of NilClass, result2
+    strainer.class.add_filter(mod)
+    strainer.class.add_filter(mod)
+    assert_equal 1, mod.include_count
   end
 end # StrainerTest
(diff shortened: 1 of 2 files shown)


test/unit/strainer_unit_test.rb
```diff
@@ -145,4 +145,13 @@ def test_global_filter_clears_cache
     Strainer.global_filter(LateAddedFilter)
     assert_equal 'filtered', Strainer.create(nil).invoke('late_added_filter', 'input')
   end
+
+  def test_add_filter_returns_nil_when_a_filter_module_was_added
+    a = Module.new
+    strainer = Context.new.strainer
+    result1 = strainer.class.add_filter(a)
+    result2 = strainer.class.add_filter(a)
```
dylanahsmith: I don't think the `add_filter` method intended to return anything.  It just happened to use Set#merge on the last line to mutate the set of filter methods.  Perhaps we should be returning `nil` explicitly for public methods so that gem users don't start depending on the return value in cases like this.

What was the undesirable effect of the bug that caused you to want to fix it?  Or did you just happen to notice the bug while reading the code?

dylanahsmith: For example, if the problem is the included hook is getting called twice for the module, we could test for the bug like this:

```ruby
  def test_add_filter_does_not_include_already_included_module
    mod = Module.new do
      class << self
        attr_accessor :include_count
        def included(mod)
          self.include_count += 1
        end
      end
      self.include_count = 0
    end
    strainer = Context.new.strainer
    strainer.class.add_filter(mod)
    strainer.class.add_filter(mod)
    assert_equal 1, mod.include_count
  end
```

chenxianyu2015: @dylanahsmith  Yes, you are right. I just happened to notice the bug while reading the code.  Your test case is more clearer than mine and your comment is great. I'm going to close this PR then make a PR with your test case. How about that?

dylanahsmith: I don't think you need to open another PR.  You could just change the test in this PR.

fw42:
Thanks for your contribution!

PR #869, opened 2017-03-13
Add globals for strict_variables & strict_filters

maximebedard:
Add the possibility to configure strict_variables and strict_filters globally. This makes it easy to switch the behaviour in tests, yet preserve the lax validation in production. (useful when you want to ensure default templates uses appropriate variables or filters).

Thoughts?

@tjoyal @fw42

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ For standard use you can just pass it the content of a file and call render with
 
 Setting the error mode of Liquid lets you specify how strictly you want your templates to be interpreted.
 Normally the parser is very lax and will accept almost anything without error. Unfortunately this can make
-it very hard to debug and can lead to unexpected behaviour. 
+it very hard to debug and can lead to unexpected behaviour.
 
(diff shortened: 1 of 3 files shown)


tjoyal:
I would try to avoid allowing to do the same thing in 2 ways. Also, passing the option to render would override the proposed "default" behaviour.

The same can be achieved from an application by making it an environment config(eg.: prod/dev/tests) then passing that config to the `render` method `options` parameter.

maximebedard:
My only thing for this is mainly around testing, it makes it a lot simpler. E.g.

```rb
def setup
  @old = Liquid::Template.strict_variables
  Liquid::Template.strict_variables = true
end

def teardown
  Liquid::Template.strict_variables = @old
end

def test_something
  # test something that fail variable lookup raises
end

# production code stays unchanged
```

This way there is no need to add a configuration setting in `(dev|test|prod).rb`.

maximebedard:
Closing, I've used params in `#render` instead.

PR #874, opened 2017-03-20
Fix Liquid/C BlockBody loading issue

fw42:
Fixes https://github.com/Shopify/liquid-c/issues/32.

Note that this PR is against the `3-0-stable` branch, not against `master`. The problem doesn't exist on the `master` branch.

cc @wjdp


lib/liquid.rb
```diff
@@ -57,6 +57,7 @@ module Liquid
 require 'liquid/parser_switching'
 require 'liquid/tag'
 require 'liquid/block'
+require 'liquid/block_body'
```
dylanahsmith: Why is block_body.rb even in this branch?  Liquid::BlockBody isn't used by any of the code in this branch.

fw42:
Good point, I totally missed that

fw42:
> we should instead update the version constraint to require liquid versions ~> 4.0

You mean Liquid 3 should depend on Liquid-C 4?

Shouldn't we rather remove the BlockBody dependency from Liquid-C version 3?

dylanahsmith:
Oh, I thought you meant the problem was with using the latest version of liquid-c with an old version of liquid.  Ya, it looks like we meant to have the liquid-c 3.x be compatible with liquid 3.x and it looks like that isn't the case

PR #877, opened 2017-03-28
Title Case filter

dyspop:
I understand this may be a good candidate for "Features that can easily be implemented on top of Liquid (for example as a custom filter or custom filesystem)" but I think it is also a "Feature that is likely to be useful to the majority of Liquid users".

```
> title_case('bob JONES mcAllister')
 => "Bob Jones McAllister" 
```
```
{% assign name = "bob JONES mcAllister" %}
{{ name | title_case }}
Bob Jones McAllister
```
Please review! If it's rejected for technical reasons I'll gladly comply and resubmit. If you find this to be rejected because it's not a useful filter generally, please let me know. Thanks.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -361,6 +361,20 @@ module Liquid
       end
     end
 
+    # capitalize first character only unless input is either all or none caps already
+    def title_case(input)
+      inputs = input.split(' ')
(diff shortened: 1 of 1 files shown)


dyspop:
Just reject. The submitted filter returns an error. I'll revise and resubmit.

PR #890, opened 2017-05-09
Fix stack level too deep error

ph3t:
Right now liquid doesn't have any limits for the recursion depth of conditions. Using a large amount of `and` operators will cause an uncatchable exception that depending on the configuration could stop the execution of the application using liquid. This could be an issue if users are allowed to provide or edit liquid templates.

To fix the issue I have added a maximum depth that will raise an exception if reached.

This issue only is reproducible with `and` operators.

@EiNSTeiN- @tjoyal

### DIFF
--- a/lib/liquid/condition.rb
+++ b/lib/liquid/condition.rb
@@ -7,6 +7,7 @@ module Liquid
   #   c.evaluate #=> true
   #
   class Condition #:nodoc:
+    @@depth = 0
     @@operators = {
       '=='.freeze => ->(cond, left, right) {  cond.send(:equal_variables, left, right) },
       '!='.freeze => ->(cond, left, right) { !cond.send(:equal_variables, left, right) },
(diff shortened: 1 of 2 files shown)


EiNSTeiN-:
@tjoyal can you recommend someone else to review this who knows liquid?

lib/liquid/condition.rb
```diff
@@ -47,6 +48,11 @@ def evaluate(context = Context.new)
       when :or
         result || @child_condition.evaluate(context)
       when :and
+        @@depth += 1
```
dylanahsmith: Don't store this in global state.  Instead, use the context to keep state local to this liquid rendering

lib/liquid/condition.rb
```diff
@@ -47,6 +48,11 @@ def evaluate(context = Context.new)
...
+        if @@depth >= 500
```
dylanahsmith: This should be a class variable so the library user has some control over the value.

lib/liquid/condition.rb
```diff
@@ -47,6 +48,11 @@ def evaluate(context = Context.new)
...
+          @@depth = 0
```
dylanahsmith: This is the only place where the depth is reduced, which means that we would get an error every 500 times an liquid `if` condition is evaluated.  Instead the depth needs to be incremented and decremented around the rendering a block.

tjoyal:
cc @simon-shopify :trollface:

PR #891, opened 2017-05-09
Use a loop to evaluate binary comparisions to avoid recursion

dylanahsmith:
Fixes #890

## Problem

Running the following script results in a SystemStackError due to the recursive method calls used to evaluate conditions

```ruby
require 'liquid'

code = "{% if true "
code << "and true " * 50_000
code << "%}rendered{% endif %}"
puts Liquid::Template.parse(code).render
```

## Solution

Use a loop to evaluate the conditions, where local variables keep track of the current state, including a `condition` variable that is used instead of `self` to evaluate child conditions.

### DIFF
--- a/lib/liquid/condition.rb
+++ b/lib/liquid/condition.rb
@@ -41,16 +41,22 @@ module Liquid
     end
 
     def evaluate(context = Context.new)
-      result = interpret_condition(left, right, operator, context)
+      condition = self
+      result = nil
+      loop do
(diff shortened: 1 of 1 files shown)


fw42:
I guess it's a bit tricky to write a regression test for this?

dylanahsmith:
The tricky part of writing the regression test would be determining the size of the condition chain needed to cause a SystemStackError with the recursive version of the code.  Even after getting that right it would end up being a slow test in our test suite.

PR #898, opened 2017-05-25
CGI powered standard filters to handle non string inputs

tjoyal:
## Problem

Some standard filters to not properly handle unexpected argument.

```ruby
Liquid::Template.parse("{{ 1 | url_encode }}").render
```

```
NoMethodError: undefined method `encoding' for 1:Fixnum
    /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/cgi/util.rb:7:in `escape'
    /Users/tjoyal/projects/liquid/lib/liquid/standardfilters.rb:45:in `url_encode'
```

## Solution

I'm usually aim to raise exceptions when the received parameters is not of the proper type(eg.: `Liquid::StandardError`), but most of the standard filters do explicit `to_s` before processing the input (for the same reason we also do an explicit check for `nil`?).

I ended up adding explicit `to_s`. 

Thoughts?
Is this going in the right direction or should all non string parameters raise an exception?
Why is `nil` treated differently here compared to other filters? Seems to be a behaviour split introduced by https://github.com/Shopify/liquid/commit/704937bc00b4472200196158c7a7f4ef43c18714#diff-844406140a08cdeddd843cfbdf0222d6L45

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -33,7 +33,7 @@ module Liquid
     end
 
     def escape(input)
-      CGI.escapeHTML(input).untaint unless input.nil?
+      CGI.escapeHTML(input.to_s).untaint unless input.nil?
     end
     alias_method :h, :escape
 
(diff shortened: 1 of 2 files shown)


dylanahsmith:
> Why is nil treated differently here compared to other filters? Seems to be a behaviour split introduced by 704937b#diff-844406140a08cdeddd843cfbdf0222d6L45

That commit didn't cause a behaviour split, previously it would `rescue input` which would return `nil` if the input was `nil`.

Issue #293, opened 2013-12-20
It went crazy when included partial share the same name with the hash key of an array parameter passed in.


IrvinFan:
I have a partial liquid as shown below in /tmp/_part.liquid:

```
In part: {% for d in part %} {{ d }} {% endfor %}
```

Set up liquid file system so that 'include' can find it:

```
> Liquid::Template.file_system = Liquid::LocalFileSystem.new('/tmp')
```

Now render a liquid string that include the partial and pass an array in using key 'part':

```
> Liquid::Template.parse("Hello: {% include 'part' %}").render('part' => ['one', 'two'])
 => "Hello: In part:  one In part:  two "
```

It looks like the partial was invoked twice and each time only one element was passed in.
If I change the hash key of the parameter to something different from the partial liquid name:

```
In part: {% for d in new_name %} {{ d }} {% endfor %}
```

everything works fine:

```
> Liquid::Template.parse("Hello: {% include 'part' %}").render('new_name' => ['one', 'two'])
 => "Hello: In part:  one  two "
```

Why was that? name conflict?

shopmike:
Because Liquid has always been this way this is considered standard behavior that can not be removed. To accomplish the intended outcome the `{% include 'tags' tags: blog.all_tags %}` syntax should be used

PR #901, opened 2017-06-14
Make 'include with' different from 'include for'

fastjames:
The include tag currently accepts the 'with' and 'for' qualifiers when a
local variable is passed in. The documentation at
http://www.rubydoc.info/github/Shopify/liquid/Liquid/Include
implies that these different terms cause different behavior, but this
has never been the case. Consequently, when a variable exists in the
parent template with the same name as the partial it is not possible to
pass an array to that liquid partial without the partial being rendered
once for each element in the same-named array.

This change allows the caller to pass an array in using the 'with'
keyword and force the partial to be rendered once. The default behavior
for same-named variables, and the behavior with the for keyword, remain
unchanged.

### DIFF
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -14,7 +14,7 @@ module Liquid
   #   {% include 'product' for products %}
   #
   class Include < Tag
-    Syntax = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?/o
+    Syntax = /(#{QuotedFragment}+)(\s+(with|for)\s+(#{QuotedFragment}+))?/o
 
     def initialize(tag_name, markup, options)
       super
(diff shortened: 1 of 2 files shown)


fastjames:
Thanks for the label! Does this mean I misread the documentation? As far as I can tell, there was no way to do what I wanted (pass an array in as a local variable to a partial without rendering the partial once per element).

fastjames:
This would also resolve the question in https://github.com/Shopify/liquid/issues/293.

dylanahsmith:
> Thanks for the label! Does this mean I misread the documentation?

This means that this change could break existing liquid templates that rely on the current behaviour, regardless of what the documentation implies.  E.g. if a liquid template uses `{% include 'product' with products %}` to render the partial for each product in products, then it would work right now and break from your change.

shopmike:
Because Liquid has always been this way this is considered standard behavior that can not be removed. To accomplish the intended outcome in #293 the `{% include 'tags' tags: blog.all_tags %}` syntax should be used

PR #909, opened 2017-07-03
added attr_accessors for collection and variable names in for tag

renevanpelt:
If you want to know which variables are present in a template before rendering, you'll need this.


pushrax:
I'd feel more comfortable with this using `attr_reader` instead

renevanpelt:
Good point

PR #914, opened 2017-07-31
raise ZeroDivisionError when dividing by 0.0

elfassy:
# Why
```ruby
[1] pry(main)> 1/0
ZeroDivisionError: divided by 0
from (pry):1:in `/'
[2] pry(main)> 1/0.0
=> Infinity
[3] pry(main)> (1/0.0).is_a?(Numeric)
=> true
[4] pry(main)> -1.0/0
=> -Infinity
```
There's no good reason to surface the float Infinity in liquid. All divisions by zero should raise the `ZeroDivisionError`

# What
- raise `Liquid::ZeroDivisionError` when making a division by 0 (or 0.0)
- updated the tests now that `FloatDomainError` are only possible with NaN (example `0/0`)

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -321,15 +321,13 @@ module Liquid
 
     # division
     def divided_by(input, operand)
+      raise Liquid::ZeroDivisionError, 'divided by 0' if operand == 0
       apply_operation(input, operand, :/)
-    rescue ::ZeroDivisionError => e
-      raise Liquid::ZeroDivisionError, e.message
     end
 
     def modulo(input, operand)
+      raise Liquid::ZeroDivisionError, 'divided by 0' if operand == 0
       apply_operation(input, operand, :%)
-    rescue ::ZeroDivisionError => e
-      raise Liquid::ZeroDivisionError, e.message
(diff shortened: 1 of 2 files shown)


Thibaut:
This makes sense to me, but we may run into backward compatibility issues.

elfassy:
@Thibaut what do you have in mind? The only thing I can think of is having that exception instead of displaying `Inf USD` (and I don't see why anyone would want to display infinity). 
Moreover, any template that passes infinity to a custom filter has a chance of blowing up if that helper doesn't handle this edge case itself, example: `{{ 1.0 | divided_by: 0 | money_format }}`

elfassy:
After speaking with @Thibaut looks like this change will be a breaking change for some people. For instance
```
{% capture percentage %}{{ item.compare_at_price_max | minus: item.price | times: 100.0 | divided_by: item.compare_at_price_max | money_without_currency | times: 100 | remove: '.0' }}%{% endcapture %}
```
would be quivalent to `{{ 'Inf' | times: 100}} #=> 0` which makes sense as the value of a discount if there is no `compare_at_price_max`. This PR could be merged in a major release bump.

PR #915, opened 2017-07-31
Add to_power as standard filter

ericcornelissen:
A simple filter to computer the input to a certain power, it
- will compute any number to any power.
- will correctly handle a negative base.
- will return 0 if the input is not a number.
- will return a float when the power is less then 1 (i.e. a root), **would like some feedback on this**.

Although this is a filter that can be implemented with ease when needed, the filter is so basic and small that I think it doesn't hurt to add it to the core. Additionally, [in some scenarios](https://github.com/simple-icons/simple-icons/issues/480#issuecomment-319086273) it will not be possible to add custom filters, in which case something as basic as this would be nice to have available 🙂

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -319,6 +319,10 @@ module Liquid
       apply_operation(input, operand, :*)
     end
 
+    def to_power(input, operand)
+      apply_operation(input, operand, :**)
+    end
+
     # division
     def divided_by(input, operand)
       apply_operation(input, operand, :/)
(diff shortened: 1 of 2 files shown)


pushrax:
Implementation LGTM.

I noted a method in https://github.com/simple-icons/simple-icons/issues/480#issuecomment-319189883 for your use-case that avoids an exponential, so until this is also needed by others (I've never seen it requested before) I'd prefer to keep API surface area down to make porting Liquid easier.

devdreamsolution (approved):

PR #916, opened 2017-08-09
Add support for iterating ranges with float boundaries.

simon-shopify:
The meaning I gave is to start at the lower bound and step by one. This has the
advantage of having (n..m) and ((n.to_f)..(m.to_f)) behave the same. Another
thing we could do that have this same property is to cast bounds to integer.
This would do the right thing if there is a floating point rounding error, but I
feel it is more surprising for cases such as (1.5..8).

Note that it is not possible to create ranges with floating point boundaries
directly in Liquid at the moment. However, since there is no distinction between
Ruby and Liquid ranges, a drop can introduce such a value by returning it.

### DIFF
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -129,7 +129,7 @@ module Liquid
       end
 
       collection = context.evaluate(@collection_name)
-      collection = collection.to_a if collection.is_a?(Range)
+      collection = collection.step(1).to_a if collection.is_a?(Range)
 
       limit = context.evaluate(@limit)
       to = limit ? limit.to_i + from : nil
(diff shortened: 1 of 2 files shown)


pushrax:
Why does it make any sense to allow this?

simon-shopify:
> Why does it make any sense to allow this?

Ah yes I think I was unclear on this point. What I'm doing is fixing an exception that happened in production. I had to choose between
* giving a proper meaning to `for` with floats, or
* gracefully failing.

I went for the first one for these reasons.
* I think the behaviour is reasonable when you are iterating over natural numbers that happen to be stored in floats. It can happen if the number comes from Javascript? Or as a result of a division?
* How do you explain to a non-programmer that why there is such distinction between integers and floats. How do you make it clear how to fix it?

pushrax:
Hmm, division is perhaps the most compelling reason why this would happen, can we handle that more directly by rounding if the number is an epsilon away from an integer?

simon-shopify:
> Hmm, division is perhaps the most compelling reason why this would happen, can we handle that more directly by rounding if the number is an epsilon away from an integer?

An raise a proper `LiquidError` if it's not?

pushrax:
Yeah, seems reasonable

devdreamsolution (approved):

PR #921, opened 2017-08-18
Do not raise when variable is defined

pascalbetz:
Variables which are defined but have a nil value should not raise.

See #749

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -73,9 +73,6 @@ Style/Documentation:
 Style/ClassAndModuleChildren:
   Enabled: false
 
-Style/TrailingComma:
-  Enabled: false
-
 Style/IndentHash:
   EnforcedStyle: consistent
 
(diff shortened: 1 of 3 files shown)


fw42:
Can you fix the style violations?

pascalbetz:
Fixed the ones that I introduced. The others should be fixed in another PR, I guess.

fw42:
ugh sorry, I wasn't aware that things are failing on master too

fw42:
Hm it's not failing for me locally. Not sure why it's failing on travis.

.rubocop.yml
```diff
@@ -73,9 +73,6 @@ Style/Documentation:
 Style/ClassAndModuleChildren:
   Enabled: false
 
-Style/TrailingComma:
-  Enabled: false
```
dylanahsmith: The style violations in existing code are because of this change to the rubocop style, which shouldn't be needed for this PR.

pascalbetz:
Hmm, picked up other rubycop with me, fixed.

pascalbetz:
@dylanahsmith @fw42 thanks, that was rather fast from PR to acceptance 😄

PR #929, opened 2017-09-11
Prefer faster `while true` over `loop do`

ashmaroli:
`Kernel#loop` is about more than **`2x slower`** than `while true`
Even slower with Ruby 2.4
 
Benchmark: https://github.com/JuanitoFatas/fast-ruby#loop-vs-while-true-code

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -34,7 +34,13 @@ Lint/UnusedBlockArgument:
 Lint/EndAlignment:
   AlignWith: variable
 
+Lint/LiteralInCondition:
+  Enabled: false
+
 Lint/UnusedMethodArgument:
+  Enabled: false
+
+Style/InfiniteLoop:
   Enabled: false
 
 Style/SingleLineBlockParams:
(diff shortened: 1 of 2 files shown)


fw42:
lolwut

fw42:
Do you see any noticeable difference in the Liquid benchmarks?

pushrax:
Is this because of the block?

ashmaroli:
> Is this because of the block?

@pushrax It probably is, according to this [answer](https://stackoverflow.com/a/45070639/7681659) on Stack Overflow

@fw42, is there a way that I can reliably benchmark `master` and this PR's branch simultaneously?

pushrax:
The standard way is just to check out each branch and run the benchmark (maybe do this twice to warm fs caches). Would be neat to have an automatic comparison tool though.

carsonreinke:
I tried this (because I was so curious) and there did not appear any noticeable difference in the benchmarks, sometimes master would win and sometimes this PR would win.

ashmaroli:
> sometimes master would win and sometimes this PR would win.

sadly that is indeed the case..
and yet when I run the benchmark from `fast-ruby`, `Kernel#loop` is *always* the slower one...

pushrax:
It's probably just peanuts compared to the cost of the rest of the code, no real reason not to merge it though I suppose

PR #930, opened 2017-09-13
Fix sort and sort_natural on sorting with non-string and nil values

er1:
If any value being sorted on is a not a string, the filter fails with a liquid error

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -121,17 +121,23 @@ module Liquid
     def sort(input, property = nil)
       ary = InputIterator.new(input)
       if property.nil?
-        ary.sort
+        ary.sort do |a, b|
+          if !a.nil? && !b.nil?
+            a <=> b
+          else
+            a.nil? ? 1 : -1
+          end
+        end
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
-      elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
+      elsif ary.all? { |el| el.respond_to?(:[]) }
         ary.sort do |a, b|
           a = a[property]
           b = b[property]
-          if a && b
+          if !a.nil? && !b.nil?
             a <=> b
           else
-            a ? -1 : 1
+            a.nil? ? 1 : -1
           end
         end
       end
@@ -143,17 +149,23 @@ module Liquid
       ary = InputIterator.new(input)
 
       if property.nil?
-        ary.sort { |a, b| a.to_s.casecmp(b.to_s) }
+        ary.sort do |a, b|
+          if !a.nil? && !b.nil?
+            a.to_s.casecmp(b.to_s)
+          else
+            a.nil? ? 1 : -1
+          end
+        end
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
-      elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
+      elsif ary.all? { |el| el.respond_to?(:[]) }
         ary.sort do |a, b|
           a = a[property]
           b = b[property]
-          if a && b
+          if !a.nil? && !b.nil?
             a.to_s.casecmp(b.to_s)
           else
-            a ? -1 : 1
+            a.nil? ? 1 : -1
           end
         end
       end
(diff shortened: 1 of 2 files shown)


fw42:
I'm curious what your use case is for this. When does this error come up for you?

er1:
One example is when one product has barcodes on their variants and the other does not, this code fails.
```
{{ product1.variants | concat: product2.variants | sort_natural: 'barcode' }}
```
I'll be honest though, I went in trying to break things.

lib/liquid/standardfilters.rb
```diff
@@ -143,11 +143,19 @@ def sort_natural(input, property = nil)
...
+        ary.sort { |a, b| a.to_s.casecmp(b.to_s) }
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
       elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
-        ary.sort { |a, b| a[property].casecmp(b[property]) }
+        ary.sort do |a, b|
+          a = a[property]
+          b = b[property]
+          if a && b
+            a[property].to_s.casecmp(b[property].to_s)
```
pushrax: Hmm I'm curious why there's no failing test... this is equivalent to `a[property][property].to_s`

lib/liquid/standardfilters.rb
```diff
@@ -143,11 +143,19 @@ def sort_natural(input, property = nil)
...
-        ary.sort { |a, b| a.casecmp(b) }
+        ary.sort { |a, b| a.to_s.casecmp(b.to_s) }
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
       elsif ary.first.respond_to?(:[]) && !ary.first[property].nil?
-        ary.sort { |a, b| a[property].casecmp(b[property]) }
+        ary.sort do |a, b|
+          a = a[property]
+          b = b[property]
+          if a && b
```
pushrax: Would it make sense remove the conditional here and `to_s` nil as well?

lib/liquid/standardfilters.rb
er1: There is no test coverage for that it seems.

lib/liquid/standardfilters.rb
er1: The conditional pushes nil to the end rather than i beginning similar to how `sort` behaves.

dylanahsmith: That's inconsistent with the `if property.nil?` branch

lib/liquid/standardfilters.rb
er1: I added test coverage for this and fixed that issue

er1:
The existing code contains the following which only checks the first element in the array being sorted for properties.
```
```
This leads to some odd and unexpected behaviour. I gather this was done for speed reasons. I would like to resolve this but I do not know what the intended behaviour should be given the performance constraints of liquid.

The output to the following statements should generally behave the same (at least for the provided test data)
```
{{ product.variants | sort: 'barcode' | map: 'barcode' | json }}
{{ product.variants | sort_natural: 'barcode' | map: 'barcode' | json }}
{{ product.variants | map: 'barcode' | sort | json }}
{{ product.variants | map: 'barcode' | sort_natural | json }}
{{ product.variants | reverse | sort: 'barcode' | map: 'barcode' | json }}
{{ product.variants | reverse | sort_natural: 'barcode' | map: 'barcode' | json }}
{{ product.variants | reverse | map: 'barcode' | sort | json }}
{{ product.variants | reverse | map: 'barcode' | sort_natural | json }}
```
for:
```
product = {
  "variants" => [
    { "barcode" => "Alpha1" },
    { "barcode" => "Alpha2" },
    { "barcode" => "Alpha3" },
    { "barcode" => nil },
    { "barcode" => nil },
    { "barcode" => nil }
  ]
}
```
produces:
```
["Alpha1","Alpha2","Alpha3",null,null,null]
Liquid error: internal
Liquid error: comparison of String with nil failed
Liquid error: internal
[]
[]
Liquid error: comparison of NilClass with String failed
Liquid error: internal 
```

tannermares:
Is this waiting on anything? Just ran into this bug when trying to sort by a property when the first item had a nil value. Would love to not have to monkey patch!

PR #954, opened 2017-11-28
Add max and min filters

nithinbekal:
This will introduce `max` and `min` filters. An example use case:

```liquid
{%- assign width = 250 -%}

{%- if image.width < width -%}
  {%- assign width = image.width -%}
{%- endif -%}
```

The above can be replaced with:

```liquid
{%- assign width = image.width | max: 250 -%}
```

If invalid inputs are given (eg. `"foo"`), they would default to 0 (eg. `{{ "foo" | max: 5 }` returns `0`, and `{{ "foo" | min:5 }}` returns `5`), which is consistent with other filters that work on numbers (ceil, floor etc), but I haven't added test cases for those.

@Thibaut

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -355,6 +355,22 @@ module Liquid
       raise Liquid::FloatDomainError, e.message
     end
 
+    def max(input, n)
+      max_value = Utils.to_number(n)
+
+      result = Utils.to_number(input)
+      result = max_value if result > max_value
+      result.is_a?(BigDecimal) ? result.to_f : result
+    end
+
+    def min(input, n)
+      min_value = Utils.to_number(n)
+
+      result = Utils.to_number(input)
+      result = min_value if result < min_value
+      result.is_a?(BigDecimal) ? result.to_f : result
+    end
+
     def default(input, default_value = ''.freeze)
       if !input || input.respond_to?(:empty?) && input.empty?
         default_value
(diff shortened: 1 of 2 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -355,6 +355,16 @@ def floor(input)
       raise Liquid::FloatDomainError, e.message
     end
 
+    def max(input, n)
+      result = [input, n].map { |number| Utils.to_number(number) }.min
```
pushrax: Allocating an array just for this is not great, we're not running on something fancy like [TruffleRuby](https://github.com/graalvm/truffleruby) that can optimize that allocation away.

nithinbekal:
@pushrax Updated the methods to use conditions instead of allocating arrays, and added additional tests for number-like things.

test/integration/standard_filter_test.rb
```diff
@@ -498,6 +498,26 @@ def test_floor
...
   end
 
+  def test_max
+    assert_template_result "4", "{{ 5 | max:4 }}"
+    assert_template_result "5", "{{ 5 | max:5 }}"
+    assert_template_result "5", "{{ 5 | max:6 }}"
+
+    assert_template_result "4.5", "{{ 4.5 | max:5 }}"
+    assert_template_result "5", "{{ width | max:5 }}", 'width' => NumberLikeThing.new(6)
+    assert_template_result "4", "{{ width | max:5 }}", 'width' => NumberLikeThing.new(4)
```
Thibaut: Could add a test for `{{ 5 | max: width }}`

nithinbekal:
Added tests for `{{ 5 | max: width }}`, as suggested by Thibaut.

PR #957, opened 2017-12-04
Add docs for at_least and at_most filters

nithinbekal:
These filters were added in #954.

@adamhollett cc @Thibaut


nithinbekal:
The names of the filters, `max` and `min` were confusing, and were changed to `at_most` and `at_least`. I've updated the docs to reflect this.

adamhollett:
This is fine to merge 👍 

Tests will fail on this branch because there are no tests for the Liquid website 😅

PR #958, opened 2017-12-06
Rename min/max filters for clarity

Thibaut:
`min` and `max` were added in https://github.com/Shopify/liquid/pull/954. The original implementation behaved as "`min` = limit a number to a minimum value" and "`max` = limit a number to a maximum value".

This makes sense in the context of Liquid, since a filter is generally interpreted as a verb/action, but is confusing in the context of programming languages in general, which tend to have `min` / `max` functions that return the minimum/maximum value passed as argument.

After realizing this, I flipped the behavior in 282d42f98d03d8bdedc1f4964c46be13ba0d98f4, before realizing that it's not that simple 😛 

Given the confusion (I've asked multiple people and got different answers), I propose we rename the filters for clarity.

@nithinbekal @pushrax @davidcornu @fw42 @dylanahsmith thoughts?

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -353,19 +353,19 @@ module Liquid
       raise Liquid::FloatDomainError, e.message
     end
 
-    def max(input, n)
-      max_value = Utils.to_number(n)
-
-      result = Utils.to_number(input)
-      result = max_value if max_value > result
-      result.is_a?(BigDecimal) ? result.to_f : result
-    end
-
-    def min(input, n)
+    def at_least(input, n)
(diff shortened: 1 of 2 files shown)


davidcornu:
Of all the discussed options I think `at_most` and `at_least` make the most sense.

fw42:
Are we not concerned that this breaks backwards compatibility?

Thibaut:
@fw42 this was added yesterday. It hasn't been released yet.

Arafatk:
Yes, Thanks.       
This seems better.

nithinbekal:
👍 This seems much better.

Issue #945, opened 2017-10-19
Page title has preceding hyphen


keyfer:
Loading https://shopify.github.io/liquid/, the page title has a preceding hyphen.

![https://screenshot.click/19-41-uj7tp-4vzli.jpg](https://screenshot.click/19-41-uj7tp-4vzli.jpg)

pushrax:
@admhlt when you have a sec

admhlt:
I remember this happened when I tried to create a workaround for a github-pages plugin being wonky. I'll see if I can fix it. Thanks!

Issue #223, opened 2013-07-12
if condition on empty string


etienneroudeix:
Hello, I was just wanting to test if my_variable was an empty string.

I'm using Liquid through Jekyll if it can help.

I tried :

{% if my_variable == '' %}
  I'm not empty
{% endif %}

but no way.

In fact I have to do like this 

{% if my_variable == %}
  I'm not empty
{% endif %}

It sound ugly to me so I let you know.

By the way it's the same behaviour using case condition.

Thank you for your job on this rocking project.

apla:
none of this really works in jekyll 2.5.4

```
{{ if page.xxx and page.xxx != "" and page.xxx != nil }}
<h1>crap!</h1>
{{ endif }}
```

parkr:
@apla Have you tried the Jekyll 3 beta? Jekyll 2 is still using Liquid 2.

rickydazla:
Is there a / what is the difference between `nil`, `blank`, and `empty`? Are they interchangeable? In Shopify Liquid is one preferable or are there situations where different ones should be used?

onsitediary-blog-de:
Github pages still use Jekyll 2.4.0. Would be great, if this issue would be fixed.

onsitediary-blog-de:
In meantime I use following workaround:
{% capture site_info %}X{{ site.form-info-website }}{% endcapture %}
{% if site_info == "X" %}
  <h1>form-info-website is not set</h1>
{% endif %}

yoshyosh:
What is the way to do this in Jekyll 2?

chrisvogt:
@yoshyosh, you can check for undefined variables like so:

```ruby
{% if !site_info %}
  {% include static-info.html %}
{% endif %}
```

resurge:
I'm having a similar (or the same actually) issue.
But in my case my variable is an array.

None of the checks above are working. Although I feel just a simple
```
{% if my_array %}
... do something with array contents ...
{% endif %}
```
should work.

7blink:
I have been using 
{% if page.testVariable.size != 0 %}

OilSlick:
I'm running Jekyll 3.0.1 and had to implement the "capture" workaround proposed by onsitediary-blog-de on April 12.  I tried other solutions mentioned in this thread with no luck.

edwinwright:
[...]

PR #959, opened 2017-12-06
Removing redundant hyphen from page title

Arafatk:
Refer to #945    
@pushrax @adamhollett


pushrax:
The hyphen should be there when there is a page title. Ironically this issue is relevant https://github.com/Shopify/liquid/issues/223.

pushrax:
It should be possible to fix the extra hyphen though, maybe with `unless page.title.blank?`

PR #960, opened 2017-12-06
Update README.md

Arafatk:
@pushrax  Maybe add a Circle CI Badge too like [here](https://github.com/Arafatk/liquid).

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 * [Version history](History.md)
 * [Liquid documentation from Shopify](http://docs.shopify.com/themes/liquid-basics)
 * [Liquid Wiki at GitHub](https://github.com/Shopify/liquid/wiki)
+* [Liquid Cheet Sheet](http://cheat.markdunkley.com/)
 * [Website](http://liquidmarkup.org/)
 
 ## Introduction


README.md
```diff
@@ -7,6 +7,7 @@
 * [Version history](History.md)
 * [Liquid documentation from Shopify](http://docs.shopify.com/themes/liquid-basics)
 * [Liquid Wiki at GitHub](https://github.com/Shopify/liquid/wiki)
+* [Liquid Cheet Sheet](http://cheat.markdunkley.com/)
```
dylanahsmith: Most of that is Shopify specific, which could cause confusion in the context of liquid being used as a gem.

dylanahsmith: I guess the same could be said about the objects documented in the "Liquid documentation from Shopify" link, but at least that one is clearer from the link text that it is referring to shopify's liquid documentation.

Arafatk: @dylanahsmith Yup!!    
Well that kinda makes sense.     
Anyways, you can add the Circleci badge if that's okay.    
Do you mind if I close this pr?

dylanahsmith: I don't even know why we have both CircleCI and TravisCI.

I don't mind if you close the PR.

PR #962, opened 2017-12-07
Minor Doc fix

Arafatk:


pushrax:
This documentation is a little lacking, other possible types are `Range`, `Drop`, `Symbol`, `Time`, and anything else with a `#to_liquid` method. But at least this PR makes the docs self-consistent 😛

Arafatk:
@pushrax  Yes, I thought the same thing but perhaps this change seems appropriate here.        
Thanks for the merge.

PR #972, opened 2018-01-24
Add a regression test for a liquid-c trim mode bug

### LINKED ISSUE #971: Whitespace filter trim whitespace inside string
Hi,

It seems that the whitespace control trims whitespace where it shouldn't within Shopify system:

```liquid
{%- capture bgset -%}{{ 'test' }} [--phone]{%- endcapture -%}
```

Result:

test[--phone]

Expected result (space after test and literal should be preserved):

test [--phone]

Thanks!

dylanahsmith:
This is a regression test to make sure https://github.com/Shopify/liquid-c/pull/39 fixes #971

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,6 @@ group :test do
   gem 'rubocop', '~> 0.49.0'
 
   platform :mri do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'bd53db95de3d44d631e7c5a267c3d934e66107dd'
+    gem 'liquid-c', github: 'Shopify/liquid-c', ref: '9168659de45d6d576fce30c735f857e597fa26f6'
   end
 end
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -12,6 +12,6 @@ group :test do
   gem 'rubocop', '~> 0.49.0'
 
   platform :mri do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'bd53db95de3d44d631e7c5a267c3d934e66107dd'
+    gem 'liquid-c', github: 'Shopify/liquid-c', branch: 'fix-rtrim-followed-by-tag'
```
dylanahsmith: I will have to wait until https://github.com/Shopify/liquid-c/pull/39 is merged before updating back to the latest commit on liquid-c's master branch

PR #977, opened 2018-02-02
standardfilters: add regex replace/remove filters

xu-cheng:
This commit adds the following new filters:
 * regex_replace
 * regex_replace_first
 * regex_remove
 * regex_remove_first

They are similar to the existing replace/replace_first/remove/remove_first filters.
Except, they treat the search string parameter as regular expression.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -204,9 +204,19 @@ module Liquid
       input.to_s.gsub(string.to_s, replacement.to_s)
     end
 
+    # Replace occurrences of a string which matches the regular expression with another
+    def regex_replace(input, regex, replacement = ''.freeze)
+      input.to_s.gsub(Regexp.new(regex.to_s), replacement.to_s)
+    end
+
     # Replace the first occurrences of a string with another
     def replace_first(input, string, replacement = ''.freeze)
       input.to_s.sub(string.to_s, replacement.to_s)
+    end
+
+    # Replace the first occurrences of a string which matches the regular expression with another
+    def regex_replace_first(input, regex, replacement = ''.freeze)
+      input.to_s.sub(Regexp.new(regex.to_s), replacement.to_s)
     end
 
     # remove a substring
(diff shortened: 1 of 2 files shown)


fw42:
Running user-supplied regular expressions can be a bit of a security concern. I'm not sure we want to add this. @clayton-shopify, wdyt?

xu-cheng:
I don't believe regular expression can cause security problem, since it is only a type-3 language under Chomsky hierarchy. Google search only shows that regex could become problem as a vector for DoS attack in some naive regex engine.  However, the implementation of the ruby regex engine is not naive. And there are many other approaches to achieve DoS attack in liquid (e.g. nested for loop).

clayton-shopify:
@xu-cheng Most regular expression engines have extensions (e.g. backreferences) which prevent the expression from being converted to a finite automaton. As a result, checking whether an input matches an expression can take exponential time.

Reference: https://perl.plover.com/NPC/

This means that evaluating user-supplied regexes generally opens up denial of service attacks.

clayton-shopify:
The problem is also described in the documentation for `Regexp`:

http://ruby-doc.org/core-2.5.0/Regexp.html#class-Regexp-label-Performance

xu-cheng:
@clayton-shopify Thanks for the information. If the DoS is the only attack concern here (which I think it is), may I suggest a simple safety guard like this:

```ruby
def regex_replace(input, regex, replacement = ''.freeze)
  Timeout::timeout(5) do
    input.to_s.gsub(Regexp.new(regex.to_s), replacement.to_s)
  rescue
    input.to_s
  end
end
```

After, regex is a very powerful tool which can certainly benefit users.

fw42:
`Timeout` in Ruby comes with it's own set of [problems](http://blog.headius.com/2008/02/rubys-threadraise-threadkill-timeoutrb.html). I'd prefer to not merge this at all, it doesn't seem like a very common feature request among Shopify theme developers.

dawez:
Actually I would love to see this feature implemented.

kwent:
Shouldn't be a concern with https://blog.saeloun.com/2022/08/09/ruby-introduces-regexp-timeout.html now

PR #981, opened 2018-02-11
[CI] Test against Ruby 2.5 version

nicolasleger:


pushrax:
Does `ruby-head` currently fail? I'd prefer to keep it out of `allow_failures` unless there's a good reason.

nicolasleger:
`ruby-head` is unstable so sometimes it can broke, sometimes not.
It still runs tests against it but without blocking development.
It is listed in `allow_failures` for `rails` gem for example https://github.com/rails/rails/blob/master/.travis.yml

pushrax:
Fair enough. I was thinking we should try to keep up to date with ruby development, but there's probably not enough delta on this repository for that to even be worth it.

PR #988, opened 2018-03-13
Assign regexps to constants

ashmaroli:
This is to avoid creating multiple instances of those regexps each time they're encountered within a given Class instance

### DIFF
--- a/lib/liquid/expression.rb
+++ b/lib/liquid/expression.rb
@@ -21,20 +21,24 @@ module Liquid
       'empty'.freeze => MethodLiteral.new(:empty?, '').freeze
     }
 
+    SINGLE_QUOTED_STRING = /\A'(.*)'\z/m
+    DOUBLE_QUOTED_STRING = /\A"(.*)"\z/m
+    INTEGERS_REGEX       = /\A(-?\d+)\z/
+    FLOATS_REGEX         = /\A(-?\d+\.\d+)\z/
+    RANGES_REGEX         = /\A\((\S+)\.\.(\S+)\)\z/
+
     def self.parse(markup)
       if LITERALS.key?(markup)
         LITERALS[markup]
       else
         case markup
-        when /\A'(.*)'\z/m # Single quoted strings
+        when SINGLE_QUOTED_STRING, DOUBLE_QUOTED_STRING
           $1
-        when /\A"(.*)"\z/m # Double quoted strings
-          $1
-        when /\A(-?\d+)\z/ # Integer and floats
+        when INTEGERS_REGEX
           $1.to_i
-        when /\A\((\S+)\.\.(\S+)\)\z/ # Ranges
+        when RANGES_REGEX
           RangeLookup.parse($1, $2)
-        when /\A(-?\d[\d\.]+)\z/ # Floats
+        when FLOATS_REGEX
           $1.to_f
         else
           VariableLookup.parse(markup)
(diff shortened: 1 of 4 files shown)


lib/liquid/expression.rb
```diff
@@ -21,20 +21,24 @@ def to_liquid
       'empty'.freeze => MethodLiteral.new(:empty?, '').freeze
     }
 
+    SINGLE_QUOTED_STRING = /\A'(.*)'\z/m
+    DOUBLE_QUOTED_STRING = /\A"(.*)"\z/m
+    INTEGERS_REGEX       = /\A(-?\d+)\z/
+    FLOATS_REGEX         = /\A(-?\d+\.\d+)\z/
```
pushrax: So, I know this was weird before, allowing `1.2.3` as a float, but I'm not sure it's 100% safe for us to change it without some testing against our production templates. Can we keep it weird for now?

Liquid-C's strict parsing [will raise an error](https://github.com/Shopify/liquid-c/blob/958052d08fb857f16a6799dbe5edeea092b54a67/ext/liquid_c/parser.c#L131-L158) in that case FWIW, though in Shopify we swallow that error and run the lax parser.

ashmaroli: I've reverted back to the "weird" regex.
Funny how it also matches `1..10` as a `Float`.. :slightly_smiling_face:

pushrax:
Thanks!

dylanahsmith:
> This is to avoid creating multiple instances of those regexps each time they're encountered within a given Class instance

Ruby pre-allocates the regexps so that it doesn't create instances each time they are encountered.  This is similar to how newer ruby versions pre-allocate frozen string literals.

ashmaroli:
> Ruby pre-allocates the regexps so that it doesn't create instances each time they are encountered.

I did not know that.. Since what ruby version is this behavior observed?

dylanahsmith:
I don't remember when that wasn't the case, since we have used object allocation profiling through stackprof to reduce memory allocations.  I just tested on the minimum ruby version liquid supports (ruby 2.1) and found that it doesn't allocate new regexp objects

```ruby
irb> p = -> { /\A\s*\z/ }
irb> p.call.object_id == p.call.object_id
=> true
```

I often use the following function that I keep in my ~/.irbrc to check if something allocates objects

```
def count_allocations
  require 'stackprof'
  profile = StackProf.run(mode: :object) do
    yield
  end
  profile[:samples]
end
```

ashmaroli:
> I don't remember when that wasn't the case..

So why wasn't *Class-constants* with regexp values not done away with in `v4.0`? Just curious..

ashmaroli:
Running the following yields contradictory results:
```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

puts "foo bar".object_id
puts "foo bar".object_id
puts "foo bar".object_id

puts ""

puts %r!\A\w+\z!.object_id
puts %r!\A\w+\z!.object_id
puts %r!\A\w+\z!.object_id
```
```
19189092
19189092
19189092

19190880
19190556
19190244
```

dylanahsmith:
Right, it isn't quite like symbols or frozen string literals in that it only avoid runtime allocations of those regexps.  It does however allocate a separate one for each line regexp literal, which is why I used a proc to return the same regexp object.

dylanahsmith:
> So why wasn't Class-constants with regexp values not done away with in v4.0? Just curious..

Using named constants can still help for readability and to avoid inconsistencies.

pushrax:
Yes, I still prefer this style.

PR #990, opened 2018-03-15
Replace String#=~(regex) w/ faster methods

ashmaroli:
This is a micro-optimization that may not make a difference in the grand scheme of things.. :grin: 

Here's a crude benchmark script in case you're curious
```ruby
require 'benchmark/ips'

WhitespaceOrNothing = /\A\s*\z/

def match_regex(input)
  !!(input =~ WhitespaceOrNothing)
end

def strip_empty(input)
  input.strip.empty?
end

[
  "",
  "  ",
  "\n\n",
  "\n \n \n"
].each do |input|
  if match_regex(input) == strip_empty(input)
    Benchmark.ips do |x|
      x.report("'#{input}'-regex") { match_regex(input) }
      x.report("'#{input}'-native") { strip_empty(input) }
      x.compare!
    end
  end
end
```

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -2,7 +2,6 @@ module Liquid
   class BlockBody
     FullToken = /\A#{TagStart}#{WhitespaceControl}?\s*(\w+)\s*(.*?)#{WhitespaceControl}?#{TagEnd}\z/om
     ContentOfVariable = /\A#{VariableStart}#{WhitespaceControl}?(.*?)#{WhitespaceControl}?#{VariableEnd}\z/om
-    WhitespaceOrNothing = /\A\s*\z/
     TAGSTART = "{%".freeze
     VARSTART = "{{".freeze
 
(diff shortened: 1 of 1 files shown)


lib/liquid/block_body.rb
```diff
@@ -44,7 +43,7 @@ def parse(tokenizer, parse_context)
           end
           parse_context.trim_whitespace = false
           @nodelist << token
-          @blank &&= !!(token =~ WhitespaceOrNothing)
+          @blank &&= token.strip.empty?
```
dylanahsmith: `strip` allocates a new string, so I wouldn't expect it to help much.  Instead, we could leverage String#match? if available, which doesn't allocate any objects.  We could even use refinements to make that more natural to use (e.g. `/\A\s*\z/.match?(token)`).

ashmaroli: While `String#match?` is fast, its only available with Ruby 2.4 and above,
I'm not sure how using refinements would be better, so feel free to overwrite this branch or replace with another PR..

PR #991, opened 2018-03-15
Backport String#match? to reduce object allocations on ruby 2.4+

dylanahsmith:
Closes #990
cc @ashmaroli

We can use refinements to backport ruby features without affecting the code using the library.

I haven't yet benchmarked the difference, so just threw this together in response to https://github.com/Shopify/liquid/pull/990#discussion_r174843621

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -1,4 +1,6 @@
 module Liquid
+  using RubyBackports
+
   class BlockBody
     FullToken = /\A#{TagStart}#{WhitespaceControl}?\s*(\w+)\s*(.*?)#{WhitespaceControl}?#{TagEnd}\z/om
     ContentOfVariable = /\A#{VariableStart}#{WhitespaceControl}?(.*?)#{WhitespaceControl}?#{VariableEnd}\z/om
@@ -44,7 +46,7 @@ module Liquid
           end
           parse_context.trim_whitespace = false
           @nodelist << token
-          @blank &&= !!(token =~ WhitespaceOrNothing)
+          @blank &&= token.match?(WhitespaceOrNothing)
         end
         parse_context.line_number = tokenizer.line_number
       end
--- a/lib/liquid/ruby_backports.rb
+++ b/lib/liquid/ruby_backports.rb
@@ -1,0 +1,13 @@
+module Liquid
+  module RubyBackports
+    Liquid.private_constant :RubyBackports
+
+    if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.4')
+      refine String do
+        def match?(regexp)
+          !!(self =~ regexp)
+        end
+      end
+    end
+  end
+end
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -45,6 +45,7 @@ module Liquid
 end
 
 require "liquid/version"
+require "liquid/ruby_backports"
 require 'liquid/lexer'
 require 'liquid/parser'
 require 'liquid/i18n'


lib/liquid/block_body.rb
```diff
@@ -1,4 +1,6 @@
 module Liquid
+  using RubyBackports
```
pushrax: Ruby 2.1 and 2.2 docs say: "You may only activate refinements at top-level, not inside any class, module or method scope." 2.3 says "You may activate refinements at top-level, and inside classes and modules."

May as well put this line at the top?

dylanahsmith: Sure.  I guess it just wasn't scoped when I tested on ruby 2.1, but I also don't need it to be scoped.

dylanahsmith: Oh right, it does make it awkward to make the Liquid::RubyBackport a private constant

pushrax: Well, it does pass on CI, not sure if the docs are just being defensive or if there's something else at play.

dylanahsmith: Did some testing and found that the following works on ruby 2.1 but not in ruby 2.3 because ruby 2.3 scopes the refinement

```ruby
module Liquid
  using RubyBackports
end

module Liquid
  class BlockBody
```

I think the nested refinement is cleaner than having `using Liquid.const_get(:RubyBackports)` at the top-level or making Liquid::RubyBackports public.

pushrax: 👍 so you can activate a refinement in a module, but it will be still globally scoped.

ashmaroli:
While the branch does run successfully on Ruby 2.3 locally, I'm not able to benchmark it...

lib/liquid/ruby_backports.rb
```diff
@@ -0,0 +1,13 @@
+module Liquid
+  module RubyBackports
+    Liquid.private_constant :RubyBackports
+
+    if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.4')
```
ashmaroli: Benchmarks show that `RbConfig::CONFIG.values_at('MAJOR', 'MINOR').join.to_i < 24` is faster..
```ruby
Benchmark.ips do |x|
  x.report("rbconfig") { RbConfig::CONFIG.values_at('MAJOR', 'MINOR').join.to_i < 24 }
  x.report("gemver") { Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.4') }
  x.compare!
end
```

pushrax: Why does this matter? It's run once at load time.

ashmaroli: yes, but since Refinements need to be repeated on every *file* for the constituent code to be *refined*, I figured having multiple instances of `Gem::Version` for no other particular reason to be counter-productive..

dylanahsmith: This is happening outside of the `refine` block so it should only get executed once. This also means the refinement would be empty and have no effect on ruby 2.4+.

pushrax:
If we end up needing multiple backports the gem could make sense, for just the one this refinement seems fine. PR LGTM, @dylanahsmith want to merge?

dylanahsmith:
The backports shouldn't affect the application code, which is why I used refinements.  I believe the `backports` gem monkey patches to add missing features, which would affect the application.

I just haven't merged this since I hadn't gotten around to benchmarking the effect of ruby refinements on performance.

shopmike:
Backports can be removed from this due to https://github.com/Shopify/liquid/pull/1131

shopmike:
Resolved in https://github.com/Shopify/liquid/pull/1165

PR #1000, opened 2018-03-28
add filters, adjust namespacing, add requires to initializer

gatesporter8:
1. Add filters from filters folder of avant-templates.
2. Change namespacing (nest filters under liquid namespace, "Template.register_filter" instead of "Liquid::Template.reqister_filter".
3. Require new filter files to liquid.rb initializer.

### DIFF
--- a/lib/liquid/array.rb
+++ b/lib/liquid/array.rb
@@ -1,0 +1,15 @@
+module Liquid
+  module Array
+    def empty(input)
+      input.empty?
+    end
+
+    def count(input)
+      input.count
+    end
+
+    alias_method :length, :count
+  end
+
+  Template.register_filter(Array)
+end
--- a/lib/liquid/date.rb
+++ b/lib/liquid/date.rb
@@ -1,0 +1,25 @@
+require 'utils'
+module Liquid
+  module Date
+    def date(input, format)
+      format = format.to_s
+      return input if format.empty?
+
+      if input == 'now'
+        company = @context['company'] || {}
+        time_zone = company['time_zone']
+        return Time.current.strftime(format) unless time_zone
+
+        Time.use_zone(time_zone) do
+          Time.current.strftime(format)
+        end
+      else
+        return input unless date = ::Utils.to_date(input)
+
+        date.strftime(format)
+      end
+    end
+  end
+
+  Template.register_filter(Date)
+end
--- a/lib/liquid/mixed.rb
+++ b/lib/liquid/mixed.rb
@@ -1,0 +1,13 @@
+module Liquid
+  module Mixed
+    def present(input)
+      input.present?
+    end
+
+    def blank(input)
+      input.blank?
+    end
+  end
+
+  Template.register_filter(Mixed)
+end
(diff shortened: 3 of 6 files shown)


lib/liquid/date.rb
```diff
@@ -0,0 +1,25 @@
+require 'utils'
+module Liquid
+  module Date
+    def date(input, format)
```
dylanahsmith, line 4: There is already a [date filter](https://github.com/Shopify/liquid/blob/e9b649b3455c63859f1b865a8684607d6ff5b050/lib/liquid/standardfilters.rb#L332)

lib/liquid/array.rb
```diff
@@ -0,0 +1,15 @@
+module Liquid
+  module Array
+    def empty(input)
+      input.empty?
+    end
+
+    def count(input)
```
dylanahsmith, line 7: Why not use the [`size` filter](https://github.com/Shopify/liquid/blob/e9b649b3455c63859f1b865a8684607d6ff5b050/lib/liquid/standardfilters.rb#L24)?

lib/liquid/mixed.rb
```diff
@@ -0,0 +1,13 @@
+module Liquid
+  module Mixed
+    def present(input)
+      input.present?
+    end
+
+    def blank(input)
+      input.blank?
```
dylanahsmith, line 8: These methods aren't defined in ruby, so this is implicitly dependent on active support.

lib/liquid/number.rb
```diff
@@ -0,0 +1,42 @@
+module Liquid
+  module Number
+    def precision(input, *args)
+      sigdigs = args[0] || 2
+      sprintf("%.#{sigdigs}f", input.to_s.gsub(/[^0-9.]/, ''))
+    end
+
+    def ordinalize(input)
+      input.ordinalize
```
dylanahsmith, line 9: This is another implicit dependency on active support

lib/liquid/number.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    # Given a number/float/string, convert it to a comma-separate, two-decimal-
+    # place number. Make sure to maintain any currency symbols as well.
+    # e.g. "$-250100" => "$-250,100.00"; 1234.5 => "1,234.50"
+    # @param [Mixed] input - the value to filter.
+    # @return [String] the money-fied result.
+    def money(input)
```
dylanahsmith, line 17: This filter is quite hard coded to a specific money format

lib/liquid/string.rb
```diff
@@ -0,0 +1,13 @@
+module Liquid
+  module String
+    def titleize(input)
+      input.titleize
+    end
```
dylanahsmith, line 5: This is another implicit active support dependency.

lib/liquid/string.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    def pluralize(num, singular, plural)
+      num == 1 ? singular : plural
```
dylanahsmith, line 8: This assumption won't be valid when dealing with other languages

PR #1005, opened 2018-04-16
Refactored and optimized rendering

christopheraue:
Measures:
1) A while loop is faster than iterating with #each.
2) Check string tokens first. String tokens are far more frequent
   than interrupt tokens. If a token is a string token, checking
   for an interrupt token can be avoided.
3) String tokens just map to themselves and don't need the special
   treatment of `BlockBody#render_node` (except the resource limit
   check).

Benchmark
=========

$ bundle exec rake benchmark:run

Before
------
```
Run 1)
              parse:     41.630  (± 0.0%) i/s -    420.000  in  10.089309s
             render:     75.962  (± 3.9%) i/s -    763.000  in  10.066823s
     parse & render:     25.497  (± 0.0%) i/s -    256.000  in  10.040862s

Run 2)
              parse:     42.130  (± 0.0%) i/s -    424.000  in  10.064738s
             render:     77.003  (± 1.3%) i/s -    777.000  in  10.093524s
     parse & render:     25.739  (± 0.0%) i/s -    258.000  in  10.024581s

Run 3)
              parse:     41.976  (± 2.4%) i/s -    420.000  in  10.021406s
             render:     76.184  (± 1.3%) i/s -    763.000  in  10.018104s
     parse & render:     25.641  (± 0.0%) i/s -    258.000  in  10.062549s
```

After
-----
```
Run 1)
              parse:     42.198  (± 0.0%) i/s -    424.000  in  10.048737s
             render:     79.495  (± 2.5%) i/s -    798.000  in  10.042610s
     parse & render:     25.874  (± 3.9%) i/s -    260.000  in  10.053336s

Run 2)
              parse:     41.961  (± 0.0%) i/s -    420.000  in  10.009858s
             render:     78.623  (± 1.3%) i/s -    791.000  in  10.064043s
     parse & render:     25.927  (± 0.0%) i/s -    260.000  in  10.028472s

Run 3)
              parse:     42.321  (± 2.4%) i/s -    424.000  in  10.021159s
             render:     79.468  (± 2.5%) i/s -    798.000  in  10.048127s
     parse & render:     26.065  (± 0.0%) i/s -    262.000  in  10.053814s
```

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -70,33 +70,27 @@ module Liquid
       output = []
       context.resource_limits.render_score += @nodelist.length
 
-      @nodelist.each do |token|
-        # Break out if we have any unhanded interrupts.
-        break if context.interrupt?
-
-        begin
+      idx = 0
+      while node = @nodelist[idx]
+        case node
+        when String
+          check_resources(context, node)
+          output << node
+        when Variable
+          render_node_to_output(node, output, context)
+        when Block
+          render_node_to_output(node, output, context, node.blank?)
+          break if context.interrupt? # might have happened in a for-block
+        when Continue, Break
           # If we get an Interrupt that means the block must stop processing. An
           # Interrupt is any command that stops block execution such as {% break %}
           # or {% continue %}
-          if token.is_a?(Continue) || token.is_a?(Break)
-            context.push_interrupt(token.interrupt)
-            break
-          end
-
-          node_output = render_node(token, context)
-
-          unless token.is_a?(Block) && token.blank?
-            output << node_output
-          end
-        rescue MemoryError => e
-          raise e
-        rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
-          context.handle_error(e, token.line_number)
-          output << nil
-        rescue ::StandardError => e
-          line_number = token.is_a?(String) ? nil : token.line_number
-          output << context.handle_error(e, line_number)
+          context.push_interrupt(node.interrupt)
+          break
+        else # Other non-Block tags
+          render_node_to_output(node, output, context)
         end
+        idx += 1
       end
 
       output.join
@@ -104,15 +98,25 @@ module Liquid
 
     private
 
-    def render_node(node, context)
-      node_output = node.is_a?(String) ? node : node.render(context)
+    def render_node_to_output(node, output, context, skip_output = false)
+      node_output = node.render(context)
       node_output = node_output.is_a?(Array) ? node_output.join : node_output.to_s
+      check_resources(context, node_output)
+      output << node_output unless skip_output
+    rescue MemoryError => e
+      raise e
+    rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
+      context.handle_error(e, node.line_number)
+      output << nil
+    rescue ::StandardError => e
+      line_number = node.is_a?(String) ? nil : node.line_number
+      output << context.handle_error(e, line_number)
+    end
 
+    def check_resources(context, node_output)
       context.resource_limits.render_length += node_output.length
-      if context.resource_limits.reached?
-        raise MemoryError.new("Memory limits exceeded".freeze)
-      end
-      node_output
+      return unless context.resource_limits.reached?
+      raise MemoryError.new("Memory limits exceeded".freeze)
     end
 
     def create_variable(token, parse_context)
--- a/lib/liquid/profiler/hooks.rb
+++ b/lib/liquid/profiler/hooks.rb
@@ -1,13 +1,13 @@
 module Liquid
   class BlockBody
-    def render_node_with_profiling(node, context)
+    def render_node_with_profiling(node, output, context, skip_output = false)
       Profiler.profile_node_render(node) do
-        render_node_without_profiling(node, context)
+        render_node_without_profiling(node, output, context, skip_output)
       end
     end
 
-    alias_method :render_node_without_profiling, :render_node
-    alias_method :render_node, :render_node_with_profiling
+    alias_method :render_node_without_profiling, :render_node_to_output
+    alias_method :render_node_to_output, :render_node_with_profiling
   end
 
   class Include < Tag


lib/liquid/block_body.rb
```diff
@@ -70,49 +70,57 @@ def render(context)
       output = []
       context.resource_limits.render_score += @nodelist.length
 
-      @nodelist.each do |token|
-        # Break out if we have any unhanded interrupts.
+      idx = 0
+      while node = @nodelist[idx]
+        render_node_to_output(node, output, context)
         break if context.interrupt?
```
fw42: Hm. Not sure I understand why this `break` was at the beginning of the loop body before but now it's at the end of the loop body. Could you explain?

lib/liquid/block_body.rb
```diff
@@ -70,49 +70,57 @@ def render(context)
...
 
       output.join
     end
 
     private
 
+    def render_node_to_output(node, output, context)
+      case node
+      when String
+        node_output = node
```
fw42: Is this extra variable necessary? Why not
```ruby
check_resources(context, node)
output << node
```
?

lib/liquid/block_body.rb
```diff
@@ -70,49 +70,57 @@ def render(context)
...
+        check_resources(context, node_output)
+        output << node_output
+      when Continue, Break
+        # If we get an Interrupt that means the block must stop processing. An
+        # Interrupt is any command that stops block execution such as {% break %}
+        # or {% continue %}
+        context.push_interrupt(node.interrupt)
+      when Block
+        node_output = render_node(node, context)
+        output << node_output unless node.blank?
```
fw42: If `node.blank?` is true, did we actually have to call `render_node`? Or could we check that before?

fw42:
Thanks for your contributions!

lib/liquid/block_body.rb
christopheraue: If an interrupt happens during `render_node_to_output` it seems unnecessary to wait for the next iteration of the loop to break out of it.

lib/liquid/block_body.rb
christopheraue: It's noI necessary. I probably left it there because of the symmetry with the other branches. Should I change it?

lib/liquid/block_body.rb
christopheraue: `render_node` may have side effects for block tags. At least the test suite lights up, if `blank?` it's checked before.

lib/liquid/block_body.rb
```diff
@@ -70,49 +70,57 @@ def render(context)
       output = []
       context.resource_limits.render_score += @nodelist.length
 
-      @nodelist.each do |token|
-        # Break out if we have any unhanded interrupts.
+      idx = 0
+      while node = @nodelist[idx]
```
dylanahsmith: > A while loop is faster than iterating with #each.

This statement surprised me, so I created this microbenchmark

```ruby
require 'benchmark'

N = 100000
LIST = (0..1000).to_a

Benchmark.bmbm do |x|
  x.report("while") do
    N.times do
      LIST.each do |item|
      end
    end
  end
  x.report("each") do
    N.times do
      idx = 0
      while item = LIST[idx]
        idx += 1
      end
    end
  end
end
```

the results were

```
while   3.470000   0.010000   3.480000 (  3.480269)
each    2.230000   0.000000   2.230000 (  2.232083)
```

so I think this change might have made things worse, even if the overall benchmark improved from the other changes in this PR.

lib/liquid/block_body.rb
dylanahsmith: We could avoid this check entirely if `token.is_a?(String)` and we didn't extract code out into the `render_node_to_output` method.  Perhaps we could instead move the exception handling into `render_node` to simplify this method without losing the ability to call `break` inside the case statement.

christopheraue: That's a good point. I'll try that.

lib/liquid/block_body.rb
christopheraue: When running your benchmark I get the same result.

But when running the liquid benchmark `while` appears to be faster than `each`. Here are three runs using `each`:

```
Run 1)
              parse:     41.466  (± 2.4%) i/s -    416.000  in  10.039060s
             render:     77.517  (± 3.9%) i/s -    777.000  in  10.036251s
     parse & render:     25.457  (± 3.9%) i/s -    256.000  in  10.060507s

Run 2)
              parse:     41.188  (± 0.0%) i/s -    412.000  in  10.004006s
             render:     77.819  (± 2.6%) i/s -    784.000  in  10.081885s
     parse & render:     25.772  (± 3.9%) i/s -    258.000  in  10.015315s

Run 3)
              parse:     41.479  (± 0.0%) i/s -    416.000  in  10.030267s
             render:     76.804  (± 6.5%) i/s -    770.000  in  10.071492s
     parse & render:     24.713  (± 4.0%) i/s -    248.000  in  10.059831s
```

Maybe one of you can check that too.

lib/liquid/block_body.rb
```diff
@@ -70,49 +70,57 @@ def render(context)
...
+      when String
+        node_output = node
+        check_resources(context, node_output)
+        output << node_output
+      when Continue, Break
+        # If we get an Interrupt that means the block must stop processing. An
+        # Interrupt is any command that stops block execution such as {% break %}
+        # or {% continue %}
+        context.push_interrupt(node.interrupt)
+      when Block
```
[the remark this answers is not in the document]
christopheraue: This was more "refactor" than "optimize". It's a little less DRY, but I find it cleaner and easier to read. It comes down to personal taste, I guess.

lib/liquid/block_body.rb
dylanahsmith: Interesting.  I tried with only the while loop change it did improve performance on the benchmark.

christopheraue:
Taking all comments into account I updated the code. The `case` statement and `break` are now directly inside the loop again. This already increased performance. I then also moved the handling of `Variable` and `Block` tokens to the top of the `case` statement. That gave the performance another push.

These are the new results:
```
Run 1)
              parse:     42.283  (± 0.0%) i/s -    424.000  in  10.028306s
             render:     83.158  (± 2.4%) i/s -    832.000  in  10.009201s
     parse & render:     26.417  (± 0.0%) i/s -    266.000  in  10.069718s

Run 2)
              parse:     41.159  (± 4.9%) i/s -    412.000  in  10.031297s
             render:     81.591  (± 3.7%) i/s -    816.000  in  10.018225s
     parse & render:     25.924  (± 3.9%) i/s -    260.000  in  10.035653s

Run 3)
              parse:     42.418  (± 2.4%) i/s -    424.000  in  10.003100s
             render:     84.183  (± 2.4%) i/s -    847.000  in  10.069781s
     parse & render:     26.726  (± 0.0%) i/s -    268.000  in  10.029857s
```

PR #1011, opened 2018-06-06
add note about how date is parsed

afeld:
Was previously unclear what formats would be accepted. Thanks!

### DIFF
--- a/filters/date.md
+++ b/filters/date.md
@@ -3,7 +3,7 @@ title: date
 description: Liquid filter that prints and formats dates.
 ---
 
-Converts a timestamp into another date format. The format for this syntax is the same as [`strftime`](http://strftime.net).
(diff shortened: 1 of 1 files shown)


filters/date.md
```diff
@@ -3,7 +3,7 @@ title: date
 description: Liquid filter that prints and formats dates.
 ---
 
-Converts a timestamp into another date format. The format for this syntax is the same as [`strftime`](http://strftime.net).
+Converts a timestamp into another date format. The format for this syntax is the same as [`strftime`](http://strftime.net). The input is parsed using Ruby's [`Time.parse`](https://ruby-doc.org/stdlib/libdoc/time/rdoc/Time.html#method-c-parse), so will accept any format accepted there.
```
adamhollett: How about:

> The input uses the same format as Ruby's [`Time.parse`](https://ruby-doc.org/stdlib/libdoc/time/rdoc/Time.html#method-c-parse).

adamhollett:
Thanks very much!

PR #1025, opened 2018-09-10
Liquid::ParseTreeVisitor

singpolyma-shopify:
Currently, there is no way to traverse the parse tree of a liquid document without using `instance_variable_get` and a lot of knowledge of library internals, which is quite brittle.

This change enables writing code to traverse the tree for reasons other than rendering (such as static analysis).

### DIFF
--- a/lib/liquid/parse_tree_visitor.rb
+++ b/lib/liquid/parse_tree_visitor.rb
@@ -17,8 +17,8 @@ module Liquid
 
     def add_callback_for(*classes, &block)
       callback = block
-      callback = ->(node, _) { block.call(node) } if block.arity.abs == 1
-      callback = ->(_, _) { block.call } if block.arity.zero?
+      callback = ->(node, _) { yield node } if block.arity.abs == 1
+      callback = ->(_, _) { yield } if block.arity.zero?
       classes.each { |klass| @callbacks[klass] = callback }
       self
     end
--- a/lib/liquid/parse_tree_visitor.rb
+++ b/lib/liquid/parse_tree_visitor.rb
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 module Liquid
-  class Traversal
+  class ParseTreeVisitor
     def self.for(node, callbacks = Hash.new(proc {}))
-      if defined?(node.class::Traversal)
-        node.class::Traversal
+      if defined?(node.class::ParseTreeVisitor)
+        node.class::ParseTreeVisitor
       else
         self
       end.new(node, callbacks)
@@ -23,12 +23,12 @@ module Liquid
       self
     end
 
-    def traverse(context = nil)
+    def visit(context = nil)
       children.map do |node|
         item, new_context = @callbacks[node.class].call(node, context)
         [
           item,
-          Traversal.for(node, @callbacks).traverse(new_context || context)
+          ParseTreeVisitor.for(node, @callbacks).visit(new_context || context)
         ]
       end
     end
--- a/lib/liquid/tags/assign.rb
+++ b/lib/liquid/tags/assign.rb
@@ -47,6 +47,12 @@ module Liquid
         1
       end
     end
+
+    class Traversal < Liquid::Traversal
+      def children
+        [@node.from]
+      end
+    end
   end
 
   Template.register_tag('assign'.freeze, Assign)
--- a/lib/liquid/tags/assign.rb
+++ b/lib/liquid/tags/assign.rb
@@ -10,6 +10,8 @@ module Liquid
   class Assign < Tag
     Syntax = /(#{VariableSignature}+)\s*=\s*(.*)\s*/om
 
+    attr_reader :to, :from
+
     def initialize(tag_name, markup, options)
       super
       if markup =~ Syntax
--- a/lib/liquid/tags/assign.rb
+++ b/lib/liquid/tags/assign.rb
@@ -48,7 +48,7 @@ module Liquid
       end
     end
 
-    class Traversal < Liquid::Traversal
+    class ParseTreeVisitor < Liquid::ParseTreeVisitor
       def children
         [@node.from]
       end
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -191,6 +191,12 @@ module Liquid
     def render_else(context)
       @else_block ? @else_block.render(context) : ''.freeze
     end
+
+    class Traversal < Liquid::Traversal
+      def children
+        (super + [@node.limit, @node.from, @node.collection_name]).compact
+      end
+    end
   end
 
   Template.register_tag('for'.freeze, For)
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -192,7 +192,7 @@ module Liquid
       @else_block ? @else_block.render(context) : ''.freeze
     end
 
-    class Traversal < Liquid::Traversal
+    class ParseTreeVisitor < Liquid::ParseTreeVisitor
       def children
         (super + [@node.limit, @node.from, @node.collection_name]).compact
       end
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -46,8 +46,7 @@ module Liquid
   class For < Block
     Syntax = /\A(#{VariableSegment}+)\s+in\s+(#{QuotedFragment}+)\s*(reversed)?/o
 
-    attr_reader :collection_name
-    attr_reader :variable_name
+    attr_reader :collection_name, :variable_name, :limit, :from
 
     def initialize(tag_name, markup, options)
       super
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -16,6 +16,8 @@ module Liquid
   class Include < Tag
     Syntax = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?/o
 
+    attr_reader :template_name_expr, :variable_name_expr, :attributes
+
     def initialize(tag_name, markup, options)
       super
 
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -109,6 +109,15 @@ module Liquid
 
       file_system.read_template_file(context.evaluate(@template_name_expr))
     end
+
+    class Traversal < Liquid::Traversal
+      def children
+        [
+          @node.template_name_expr,
+          @node.variable_name_expr
+        ] + @node.attributes.values
+      end
+    end
   end
 
   Template.register_tag('include'.freeze, Include)
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -110,7 +110,7 @@ module Liquid
       file_system.read_template_file(context.evaluate(@template_name_expr))
     end
 
-    class Traversal < Liquid::Traversal
+    class ParseTreeVisitor < Liquid::ParseTreeVisitor
       def children
         [
           @node.template_name_expr,
--- a/lib/liquid/traversal.rb
+++ b/lib/liquid/traversal.rb
@@ -15,7 +15,7 @@ module Liquid
       @callbacks = callbacks
     end
 
-    def callback_for(*classes, &block)
+    def add_callback_for(*classes, &block)
       callback = block
       callback = ->(node, _) { block.call(node) } if block.arity.abs == 1
       callback = ->(_, _) { block.call } if block.arity.zero?
--- a/lib/liquid/traversal.rb
+++ b/lib/liquid/traversal.rb
@@ -1,0 +1,118 @@
+# frozen_string_literal: true
+
+module Liquid
+  class Traversal
+    def self.for(node, callbacks = Hash.new(proc {}))
+      kase = CASES.find { |(klass, _)| node.is_a?(klass) }&.last
+      (kase || self).new(node, callbacks)
+    end
+
+    def initialize(node, callbacks)
+      @node = node
+      @callbacks = callbacks
+    end
+
+    def callback_for(*classes, &block)
+      callback = block
+      callback = ->(node, _) { block.call(node) } if block.arity.abs == 1
+      callback = ->(_, _) { block.call } if block.arity.zero?
+      classes.each { |klass| @callbacks[klass] = callback }
+      self
+    end
+
+    def traverse(context = nil)
+      children.map do |node|
+        item, new_context = @callbacks[node.class].call(node, context)
+        [
+          item,
+          Traversal.for(node, @callbacks).traverse(new_context || context)
+        ]
+      end
+    end
+
+    protected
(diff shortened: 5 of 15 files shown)


fw42:
What is this needed for?

singpolyma-shopify:
@fw42 https://shopify.slack.com/archives/C07F3LRPF/p1536350533000100

singpolyma-shopify:
@pushrax Wouldn't such a visitor still need a way to actually get at the nodes, however?  Or would you suggest building one like that that knows about the instance variable names, etc?

singpolyma-shopify:
https://github.com/rmosolgo/graphql-ruby/blob/132e274c04e45c9fd17dbd4abb992cd0115ea22a/lib/graphql/language/visitor.rb#L47

The example here uses a `node.children` method, which is basically what I'm trying to add here.

pushrax:
> Wouldn't such a visitor still need a way to actually get at the nodes, however?

Yeah, we definitely need a way to get at the children. It doesn't necessarily need to be a public API though, and it could be better-defined. In the context of `nodelist`, a node is a String or a thing with a `render` method (Tag, BlockBody, Variable). The thing that makes me uncomfortable about having a public holistic `children` API as proposed is type ambiguity. If we introduce this, it needs to not collide in terminology, and/or we need to redefine "node" in a specific way. For the sake of argument, wlog, let's try to figure out the definition of "node" we need.

Perusing some types:

- Tag: base class for tags. Has an empty `nodelist` method, has a `render` method.
  - Block: inherits Tag, these are tags that have a start and an end and some stuff in between. They contain >0 BlockBody instances and delegate `nodelist` to them.
    - If, For, etc. inherit Block
- BlockBody: contains the contents between the start and end of a Block. Has a `render` method.
- Variable: contains a name of a type Expression.parse can return and >=0 filters, which each have >=0 arguments of a type Expression.parse can return. `Variable` does not inherit from `Tag`, it's a special thing, but it does still have a `render` method.
- VariableLookup: some access to a drop, which can have recursive references to other VariableLookups and other types returned from Expression.parse. This is the closest thing to an expression node we have, but is not currently considered a node at all.
- Condition: compares a "left" value of any type to a "right" value of any type, so long as those types respond to the comparator method. In practise, these types come from Expression.parse. Contains a BlockBody "attachment" that should be rendered when the condition is met. However, a condition doesn't itself have a nodelist, and it's not a node at the moment; it doesn't have a `render` method. The parent of the condition (an If, Case, or Unless) maps the attachments for its own nodelist and for rendering. If Condition becomes a node, this probably wouldn't make sense.

So, what _is_ a node in this new world? I guess it's a sum type of all these classes and their members. Expression.parse can return strings, integers, ranges, floats, and VariableLookups, so these must all be node types as well. I suppose there is nothing intrinsically wrong with that, but it's hard to reason about what you'll get when you call `children` with such a complicated array of possibilities.

Your choice to reject strings in the implementation of `BlockBody#each_node` is a good example of a how this type system is confusing. A string is probably a valid node type, but it's not obvious at all why without going through how Expression.parse is used. I can't even say it's completely clear to me if you got it wrong or right, because a string in that context might really be a different type than a string in the Expression.parse context.

The code that depends on such a `children` method is likely to be strongly coupled to the Liquid version, perhaps brittle. This is why I suggested the visitor approach: you declare you want to know about e.g. VariableLookups and that's all you are given. This somewhat hides the complexity of the node sum type from the caller.

A refactor to create a proper Node class that is inherited globally and provides a consistent API is a potential way to alleviate some complexity.

singpolyma-shopify:
@pushrax How would you suggest making it non-public?  Have the `Liquid::Traversal` (or `Liquid::Visitor` or whoever we call it) use `send` to bypass access control and get the children?  Or giving the knowledge of what instance variables to pull, etc, to the visitor class?

singpolyma-shopify:
Hmm, so, probably having all the subnodes be readable anyway (via public attr_readers) is desirable -- I've run into a few I want in my code outside of the traversal.  So maybe the right thing to do is expose all of these, and then the logic of which things are children can live in the visitor but we don't need to use `instance_variable_get` or similar hacks anymore in that case.

Thoughts?

pushrax:
That sounds quite reasonable to me. Having a Ruby-visible public `children` method that is documented as internal is okay with me as well.

singpolyma-shopify:
Implemented, cleaned up, and added tests

singpolyma-shopify:
Not sure what's up with the policial failures... they don't match local rubocop run.  Is policial not using the config from the repo?

singpolyma-shopify:
Ok, it passes Policial here now.  Local rubocop failures in files I didn't modify persist, but they seem to be present on master

lib/liquid/traversal.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+
+module Liquid
+  class Traversal
```
pushrax: Visitor is a name that carries more useful connotation for me here. Is it the wrong name for a reason I'm missing?

lib/liquid/traversal.rb
```diff
@@ -0,0 +1,120 @@
...
+    def self.for(node, callbacks = Hash.new(proc {}))
+      kase = CASES.find { |(klass, _)| node.is_a?(klass) }&.last
```
pushrax: `CASES` isn't used for anything else, maybe it should be inlined to make the code clearer?

```rb
traversal_class = case node
when Liquid::Assign then Assign
when Liquid::Case then Case
...
```

lib/liquid/traversal.rb
```diff
@@ -0,0 +1,120 @@
...
+      def children
+        [@node.name] + @node.filters.flatten
+      end
+    end
+
+    class VariableLookup < Traversal
+      def children
+        @node.lookups
+      end
+    end
```
pushrax: I'm not sure this is better than colocating these definitions alongside their corresponding classes. Perhaps the convention could be:

```rb
# variable_lookup.rb
module Liquid
  class VariableLookup
    class Traversal < Liquid::Traversal
      def children
        @node.lookups
      end
    end
  end
end
```

or do away with this abstraction altogether and put a children method on VariableLookup etc directly. If the latter, picking the method name is the hardest problem, given the rant about type consistency I posted earlier. Perhaps something to the effect of `children_of_all_types`, maybe phrased more eloquently. Then the new public accessors can be removed as well. Reflecting back, I think my dislike of the method added in the original PR was:

- the method name was ambiguous
- I'd rather have consumers use the visitor pattern to access the data

Having the method at all wasn't the issue for me.

lib/liquid/traversal.rb
```diff
@@ -0,0 +1,120 @@
...
+      kase = CASES.find { |(klass, _)| node.is_a?(klass) }&.last
+      (kase || self).new(node, callbacks)
+    end
+
+    def initialize(node, callbacks)
+      @node = node
+      @callbacks = callbacks
+    end
+
+    def callback_for(*classes, &block)
```
pushrax: I'm not super happy with this name, but can't think of any great alternatives. Some options I considered:

- `add_callback`
- `add_callback_for`
- `add_visitor`
- `add_visitor_for`
- `on_visit`
- `[]=`

WDYT?

lib/liquid/traversal.rb
singpolyma-shopify: I felt like since this is just one possible instance of the visitor pattern, taking the name of the whole pattern for this case was over-broad.  But happy to call it `Visitor` if that makes more sense to you.

lib/liquid/traversal.rb
singpolyma-shopify: Doing it as an inline `case` expression fails policial because there are too many cases.  Doing it with data also makes it easier to extend this to a registry-style model in the future (maybe for custom liquid tags to add to or similar?)

lib/liquid/traversal.rb
pushrax: `AllTypesVisitor`? I'm basically looking for the ability to have a decent idea what this does without reading the implementation.

singpolyma-shopify:
They individual classes do have test coverage from the integration tests -- I did it that way because setting up the tree by parsing a Liquid snippet seemed much cleaner than doing manual object-unit setup.

I chose the full Visitor pattern instead of a private-ish method for a couple reasons: (1) the idea of a private-ish method is very bothersome, nothing but docs to even suggest it is not meant to be used directly (2) no pollution of the core-LIquid namespaces with methods that are only used by a different module

lib/liquid/parse_tree_visitor.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def initialize(node, callbacks)
+      @node = node
+      @callbacks = callbacks
+    end
+
+    def add_callback_for(*classes, &block)
+      cb = block
+      cb = ->(node, _) { block[node] } if block.arity.abs == 1
+      cb = ->(_, _) { block[] } if block.arity.zero?
```
fw42: nitpick but I find the `[]` notation unnecessarily magical. Can we use `call()`?

lib/liquid/parse_tree_visitor.rb
```diff
@@ -0,0 +1,44 @@
...
+      end.new(node, callbacks)
+    end
+
+    def initialize(node, callbacks)
+      @node = node
+      @callbacks = callbacks
+    end
+
+    def add_callback_for(*classes, &block)
+      cb = block
```
fw42: can we call this `callback` instead of `cb`?

lib/liquid/parse_tree_visitor.rb
```diff
@@ -0,0 +1,44 @@
...
+      cb = ->(node, _) { block[node] } if block.arity.abs == 1
```
fw42: why is the `abs` needed?

lib/liquid/parse_tree_visitor.rb
```diff
@@ -0,0 +1,44 @@
...
+      cb = ->(_, _) { block[] } if block.arity.zero?
+      classes.each { |klass| @callbacks[klass] = cb }
+      self
+    end
+
+    def visit(context = nil)
+      children.map do |node|
+        item, new_context = @callbacks[node.class][node, context]
```
fw42: same here, `[node, context]` makes this unnecessarily cryptic imho. What's wrong with `call(node, context)`?

lib/liquid/parse_tree_visitor.rb
```diff
@@ -0,0 +1,44 @@
...
+        [
+          item,
+          ParseTreeVisitor.for(node, @callbacks).visit(
+            new_context.nil? ? context : new_context
```
fw42: nitpick: maybe `new_context || context` seems a bit simpler?

lib/liquid/tags/assign.rb
```diff
@@ -45,6 +47,12 @@ def assign_score_of(val)
         1
       end
     end
+
+    class ParseTreeVisitor < Liquid::ParseTreeVisitor
+      def children
+        [@node.from]
```
fw42: why is `@node.to` not also considered a child?

lib/liquid/tags/for.rb
```diff
@@ -46,8 +46,8 @@ module Liquid
   class For < Block
     Syntax = /\A(#{VariableSegment}+)\s+in\s+(#{QuotedFragment}+)\s*(reversed)?/o
 
-    attr_reader :collection_name
-    attr_reader :variable_name
+    attr_reader :collection_name, :variable_name, :limit, :from,
+      :for_block, :else_block
```
fw42: why did `for_block` and `else_block` need to be accessible? doesn't seem like you are using that anywhere

lib/liquid/tags/include.rb
```diff
@@ -107,6 +109,15 @@ def read_template_from_file_system(context)
 
       file_system.read_template_file(context.evaluate(@template_name_expr))
     end
+
+    class ParseTreeVisitor < Liquid::ParseTreeVisitor
+      def children
+        [
+          @node.template_name_expr,
+          @node.variable_name_expr
+        ] + @node.attributes.values
```
fw42: why only the `values`?

test/integration/parse_tree_visitor_test.rb
```diff
@@ -0,0 +1,231 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class ParseTreeVisitorTest < Minitest::Test
+  include Liquid
+
+  def traversal(template)
```
fw42: nitpick: I think helpers like this are conventionally put at the end of the test and made `private`

test/integration/parse_tree_visitor_test.rb
```diff
@@ -0,0 +1,231 @@
...
+    ParseTreeVisitor
+      .for(Template.parse(template).root)
+      .add_callback_for(VariableLookup, &:name)
+      .visit.flatten.compact
```
fw42: Why the `visit.flatten.compact` here? by making this a part of the test, I might accidentally break the API that you wanted without your tests covering it. For example I could change the code to always add a `nil` to the return value of every `visit` and that would most likely break consumers of this new API, but none of your tests would cover that. Same for the `flatten` (I could add some garbage array nesting and your tests wouldn't catch it).

I think you should add at least one or two tests or so to explicitly assert that the array structure is what you wanted. If you *always* want this to be flat and compact then maybe `#visit` itself should do that (rather than the caller)?

lib/liquid/parse_tree_visitor.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Liquid
+  class ParseTreeVisitor
+    def self.for(node, callbacks = Hash.new(proc {}))
+      if defined?(node.class::ParseTreeVisitor)
```
fw42: I feel like this subclassing pattern seems overkill. Why not just have each class define a `children` method (then you also wouldn't have to make so many accessors public) and then only have one `ParseTreeVisitor` class (rather than one subclass per tag)?

lib/liquid/parse_tree_visitor.rb
singpolyma-shopify: when there are optional arguments, ruby returns a negative number for arity

lib/liquid/tags/assign.rb
singpolyma-shopify: `@node.to` is just the name of the variable being assigned, it's not an expression

lib/liquid/tags/include.rb
singpolyma-shopify: the keys in the hash are the name of attributes to be assigned, not expressions

lib/liquid/parse_tree_visitor.rb
singpolyma-shopify: As discussed above, using a full visitor pattern instead of a method that has to be public but that we don't want anyone to call seemed cleaner, and also allows us to keep traversal-related things out of the main namespace.

If you feel very strongly, I can change this back again.

singpolyma-shopify:
@fw42 Sorry about that -- somehow didn't push all my changes but thought I had.  Made those other changes you requested now.

PR #1026, opened 2018-09-13
Add `where` filter to standard filters

samdoiron:
Users of Liquid will often wish to filter an array to only those items that match a certain criteria. For example, showing "pinned" messages at the top of a list.

Example usage:

`{{ comments | where: "pinned" | first }}`

or

`{{ products | where: "category", "kitchen" }}`

* Add where filter to standard filters
* Add tests for new where functionality

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -151,6 +151,20 @@ module Liquid
       end
     end
 
+    # Filter the elements of an array to those with a certain property value.
+    # By default the target is any truthy value.
+    def where(input, property, target_value = nil)
+      ary = InputIterator.new(input)
+
+      if ary.empty?
+        []
+      elsif ary.first.respond_to?(:[]) && target_value.nil?
+        ary.where_present(property)
+      elsif ary.first.respond_to?(:[])
+        ary.where(property, target_value)
+      end
+    end
+
     # Remove duplicate elements from an array
     # provide optional property with which to determine uniqueness
     def uniq(input, property = nil)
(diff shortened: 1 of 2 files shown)


fw42:
Worth noting that Jekyll, which uses Liquid, already has similar functionality [here](https://github.com/jekyll/jekyll/blob/4271495fcaad4073b2375b76089041acd16ffb49/lib/jekyll/filters.rb#L169). I suspect that merging this PR might break Jekyll in some way maybe? Or maybe their implementation will just shadow ours. cc @parkr

samdoiron:
Quick look through the filter registration system seems to suggest that its a most-recent-wins situation, so this shouldn't break those who have already defined a filter with this name.

pushrax:
Without a passed target value, perhaps the most sensible behaviour is to target any truthy value, i.e. not false or nil.

samdoiron:
Thats a good idea, it would be useful to eg. find all the items that have an associated value. I'll make that change.

---

Also: working on pulling down Jekyll to see if pointing it at this version will break things.

samdoiron:
Pointed Jekyll at this branch and all the tests run green.

adamhollett:
From what I know about the Jekyll core team, @oe or @pathawks might be the right people to get eyes on this.

pathawks:
I believe Jekyll already includes its own version of some Liquid filters (`sort` is one, there may be others); I do not think this is a problem. :+1:

lib/liquid/standardfilters.rb
```diff
@@ -151,6 +151,20 @@ def sort_natural(input, property = nil)
       end
     end
 
+    # Filter the elements of an array to those with a certain property value.
+    # by default the target value is assumed to be `true`.
```
Thibaut: Nit: capitalize "by" since it's the first word of a sentence. Remove "assumed to be". Change "true" to "truthy".

PR #1001, opened 2018-03-29
Spring cleanup

DirtyF:
👋 

I use this💧 documentation a lot, so I thought it was fair to give it a little spring cleanup:

- [x] Fix redirects for basics, filters and tags menu sections
- [x] Specify website language
- [x] Use Jekyll's URL filters
- [x] Avoid link loop on homepage
- [x] update dependencies
- [x] update Travis config

😘

PR #1031, opened 2018-09-21
Fix redirects

DirtyF:
Redirections were working locally in #1001 but turns out, they're not fix online.

As Jekyll documentation links to these URLs, it's a bummer for users.

## Problem

- https://shopify.github.io/liquid/basics/ currently returns 404
- https://shopify.github.io/liquid/filters/ currently returns 404
- https://shopify.github.io/liquid/tags/ currently returns 404

### Fixes

 - Use `redirect_from` instead of `redirect_to` 
 - Set URL to GitHub Pages website as liquidmarkup.org returns a 301


adamhollett:
Thanks again :)

PR #1032, opened 2018-09-27
Single regexp for strip_html

printercu:
Hi!

Benchmark shows almost 2x improvement.

<details>
<summary>Benchmark source</summary>

```ruby
STRIP_HTML = Regexp.union(
  /<script.*?<\/script>/m,
  /<!--.*?-->/m,
  /<style.*?<\/style>/m,
  /<.*?>/m,
)

def strip_html_old(input)
  empty = ''.freeze
  input.to_s.gsub(/<script.*?<\/script>/m, empty).gsub(/<!--.*?-->/m, empty).gsub(/<style.*?<\/style>/m, empty).gsub(/<.*?>/m, empty)
end

def strip_html_new(input)
  input.to_s.gsub(STRIP_HTML, ''.freeze)
end

require 'benchmark/ips'        

string = <<~HTML
  <div>test</div>
  <div id='test'>test</div>
  <script type='text/javascript'>document.write('some stuff');</script>
  <style type='text/css'>foo bar</style>
  <div\nclass='multiline'>test</div>
  <!-- foo bar \n test -->test
HTML

Benchmark.ips do |x|
  x.report('old') { strip_html_old(string) }
  x.report('new') { strip_html_new(string) }
  x.compare!
end
```
</details>

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -11,6 +11,12 @@ module Liquid
       "'".freeze => '&#39;'.freeze
     }
     HTML_ESCAPE_ONCE_REGEXP = /["><']|&(?!([a-zA-Z]+|(#\d+));)/
+    STRIP_HTML = Regexp.union(
+      /<script.*?<\/script>/m,
+      /<!--.*?-->/m,
+      /<style.*?<\/style>/m,
+      /<.*?>/m,
+    )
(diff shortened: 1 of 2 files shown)


pushrax:
I modified this to avoid changing behaviour while keeping a lot of the speedup.

PR #1046, opened 2018-10-19
Make builds green

davidcornu:
Just realized master builds are failing (example: https://travis-ci.org/Shopify/liquid/jobs/443702626) because of RuboCop violations. This also means we're not actually running tests in master as the default Rake task runs `:rubocop` first. 

https://github.com/Shopify/liquid/blob/ca9e75db5329b864982218f70b9f3bab9a8ddc05/Rakefile#L6

## Changes

- Our `.rubocop.yml` ignores a couple of different rules for trailing commas but these introduced in a later version so RuboCop still complains about them
- With the trailing commas ignored, RuboCop was still complaining about the perceived complexity of a couple of methods in `StandardFilters`

    ```
    lib/liquid/standardfilters.rb:121:5: C: Perceived complexity for sort is too high. [13/11]
        def sort(input, property = nil)
        ^^^
    lib/liquid/standardfilters.rb:148:5: C: Perceived complexity for sort_natural is too high. [13/11]
        def sort_natural(input, property = nil)
        ^^^
    ```

    Rather than whitelist these or change the rule I removed some duplication and added an early return for empty inputs which solved the problem and cleans things up nicely.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -71,10 +71,7 @@ Style/Documentation:
 Style/ClassAndModuleChildren:
   Enabled: false
 
-Style/TrailingCommaInArrayLiteral:
-  Enabled: false
-
-Style/TrailingCommaInHashLiteral:
+Style/TrailingCommaInLiteral:
   Enabled: false
 
 Layout/IndentHash:
@@ -125,6 +122,6 @@ Style/TrivialAccessors:
 Style/WordArray:
   Enabled: false
 
-Naming/MethodName:
+Style/MethodName:
   Exclude:
     - 'example/server/liquid_servlet.rb'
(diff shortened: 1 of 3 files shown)


davidcornu:
> Build #716204 terminated with an error.
> 
> ```
> The `Style/TrailingCommaInLiteral` cop no longer exists. Please use `Style/TrailingCommaInArrayLiteral` and/or `Style/TrailingCommaInHashLiteral` instead.
> (obsolete configuration found in .rubocop.yml20181019-55-1wvyxx5, please update it)
> ```

It would seem Policial is running a different version of RuboCop than the one we're declaring in the Gemfile. Not sure what the best course of action is.

PR #1049, opened 2018-11-08
 add fact that to_str doesn't do anything anymore for drops to the history.md

hampei:
When I wrote my first drop, to_s didn't change the way a drop class rendered itself, but to_str did. 
In the new version it's the other way around, but I can't find anything on this in issues or commits.

### DIFF
--- a/History.md
+++ b/History.md
@@ -22,6 +22,7 @@
 * Liquid::Template.file_system's read_template_file method is no longer passed the context. (#441) [James Reid-Smith]
 * Remove `liquid_methods` (See https://github.com/Shopify/liquid/pull/568 for replacement)
 * Liquid::Template.register_filter raises when the module overrides registered public methods as private or protected (#705) [Gaurav Chande]
+* overriding to_str on a drop doesn't do anything anymore, override to_s instead.
 
 ### Fixed
 
--- a/History.md
+++ b/History.md
@@ -12,7 +12,7 @@
 * Add url_decode filter to invert url_encode (#645) [Larry Archer]
 * Add global_filter to apply a filter to all output (#610) [Loren Hale]
 * Add compact filter (#600) [Carson Reinke]
-* Rename deprecated "has_key?" and "has_interrupt?" methods (#593) [Florian Weingarten]
+* Rename deprecated "has_key?" and "has_interrupt?" methods to "key?" and "interrupt?" (#593) [Florian Weingarten]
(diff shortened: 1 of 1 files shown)


pushrax:
This is due to https://github.com/Shopify/liquid/pull/492/commits/9f7e6011109e0f979043f89542f5bcebea51e407. I'm fairly sure breaking `to_str` for drops was not intended. We could solve this by having the default `Liquid::Drop#to_s` implementation be something like:

```diff
- self.class.to_s
+ respond_to?(:to_str) ? to_str : self.class.to_s
```

Or change [`render_node_to_output`](https://github.com/Shopify/liquid/blob/2b3c81cfd05ef23869d88cde03faafa1d393ac5b/lib/liquid/block_body.rb#L103):

```diff
- node_output = node_output.is_a?(Array) ? node_output.join : node_output.to_s
+ node_output = if node_output.is_a?(Array)
+   node_output.join
+ elsif node_output.respond_to?(:to_str)
+   node_output.to_str
+ else
+   node_output.to_s
+ end
```

or just state the breaking change. WDYT @dylanahsmith @fw42? `to_str` seems like a reasonable thing to support.

dylanahsmith:
I'm onboard with changing the default `Liquid::Drop#to_s` method to support overriding `to_str`.  However, I would avoid the respond_to check by doing this

```diff
-    def to_s
+    def to_str
       self.class.name
     end

+    def to_s
+      to_str
+    end
+
```

pushrax:
Very good point, that is the more sensible way.

hampei:
@dylanahsmith All of my drops now actually look exactly like this, since only defining to_s broke some of our filters. 

So shall I make a new PR for the other changes I made to the history.md? 

Feel free to just copy them into master, saves me work.

History.md
```diff
@@ -22,6 +23,7 @@
 * Liquid::Template.file_system's read_template_file method is no longer passed the context. (#441) [James Reid-Smith]
 * Remove `liquid_methods` (See https://github.com/Shopify/liquid/pull/568 for replacement)
 * Liquid::Template.register_filter raises when the module overrides registered public methods as private or protected (#705) [Gaurav Chande]
+* Overriding to_str on a drop doesn't do anything anymore, override to_s instead
```
dylanahsmith: I don't think it is common to document unintentional changes in the History.md.  I think we could just document that it is fixed in a follow-up PR for the next version.  Otherwise, LGTM

devdreamsolution (approved):

PR #1059, opened 2019-01-16
Apply error-checking to sort, sort_natural, where, uniq, map, compact filter(s)

itsGarland:
Some types (for example `Array`s and `Integer`s) cannot be indexed using strings. Previously this would raise a TypeError when attempted. Now we catch that error and return `nil` instead. Original issue was with map but the same issue was presented in many of the other filters. I also inlined the where implementation because the `rescue` block was no longer necessary in the class method. 

* Created get_property method that raises ArgumentError if property hash call raises TypeError

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -128,8 +128,10 @@ module Liquid
           nil_safe_compare(a, b)
         end
       elsif ary.all? { |el| el.respond_to?(:[]) }
-        ary.sort do |a, b|
-          nil_safe_compare(a[property], b[property])
+        begin
+          ary.sort { |a, b| nil_safe_compare(a[property], b[property]) }
+        rescue TypeError
+          raise_property_error(property)
         end
       end
     end
@@ -146,8 +148,10 @@ module Liquid
           nil_safe_casecmp(a, b)
         end
       elsif ary.all? { |el| el.respond_to?(:[]) }
-        ary.sort do |a, b|
-          nil_safe_casecmp(a[property], b[property])
+        begin
+          ary.sort { |a, b| nil_safe_casecmp(a[property], b[property]) }
+        rescue TypeError
+          raise_property_error(property)
         end
       end
     end
@@ -160,9 +164,17 @@ module Liquid
       if ary.empty?
         []
       elsif ary.first.respond_to?(:[]) && target_value.nil?
-        ary.where_present(property)
+        begin
+          ary.select { |item| item[property] }
+        rescue TypeError
+          raise_property_error(property)
+        end
       elsif ary.first.respond_to?(:[])
-        ary.where(property, target_value)
+        begin
+          ary.select { |item| item[property] == target_value }
+        rescue TypeError
+          raise_property_error(property)
+        end
       end
     end
 
@@ -176,7 +188,11 @@ module Liquid
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
       elsif ary.first.respond_to?(:[])
-        ary.uniq{ |a| a[property] }
+        begin
+          ary.uniq { |a| a[property] }
+        rescue TypeError
+          raise_property_error(property)
+        end
       end
     end
 
@@ -198,6 +214,8 @@ module Liquid
           r.is_a?(Proc) ? r.call : r
         end
       end
+    rescue TypeError
+      raise_property_error(property)
     end
 
     # Remove nils within an array
@@ -210,7 +228,11 @@ module Liquid
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
       elsif ary.first.respond_to?(:[])
-        ary.reject{ |a| a[property].nil? }
+        begin
+          ary.reject { |a| a[property].nil? }
+        rescue TypeError
+          raise_property_error(property)
+        end
       end
     end
 
@@ -394,6 +416,10 @@ module Liquid
 
     private
 
+    def raise_property_error(property)
+      raise Liquid::ArgumentError.new("cannot select the property '#{property}'")
+    end
+
     def apply_operation(input, operand, operation)
       result = Utils.to_number(input).send(operation, Utils.to_number(operand))
       result.is_a?(BigDecimal) ? result.to_f : result
@@ -459,24 +485,6 @@ module Liquid
         @input.each do |e|
           yield(e.respond_to?(:to_liquid) ? e.to_liquid : e)
         end
-      end
-
-      def where(property, target_value)
-        select do |item|
-          item[property] == target_value
-        end
-      rescue TypeError
-        # Cannot index with the given property type (eg. indexing integers with strings
-        # which are only allowed to be indexed by other integers).
-        raise ArgumentError.new("cannot select the property `#{property}`")
-      end
-
-      def where_present(property)
-        select { |item| item[property] }
-      rescue TypeError
-        # Cannot index with the given property type (eg. indexing integers with strings
-        # which are only allowed to be indexed by other integers).
-        raise ArgumentError.new("cannot select the property `#{property}`")
       end
     end
   end
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -268,8 +268,32 @@ class StandardFiltersTest < Minitest::Test
     assert_equal [], @filters.sort([], "a")
   end
 
+  def test_sort_invalid_property
+    foo = [
+      [1],
+      [2],
+      [3]
+    ]
+
+    assert_raises Liquid::ArgumentError do
+      @filters.sort(foo, "bar")
+    end
+  end
+
   def test_sort_natural_empty_array
     assert_equal [], @filters.sort_natural([], "a")
+  end
+
+  def test_sort_natural_invalid_property
+    foo = [
+      [1],
+      [2],
+      [3]
+    ]
+
+    assert_raises Liquid::ArgumentError do
+      @filters.sort_natural(foo, "bar")
+    end
   end
 
   def test_legacy_sort_hash
@@ -295,8 +319,32 @@ class StandardFiltersTest < Minitest::Test
     assert_equal [], @filters.uniq([], "a")
   end
 
+  def test_uniq_invalid_property
+    foo = [
+      [1],
+      [2],
+      [3]
+    ]
+
+    assert_raises Liquid::ArgumentError do
+      @filters.uniq(foo, "bar")
+    end
+  end
+
   def test_compact_empty_array
     assert_equal [], @filters.compact([], "a")
+  end
+
+  def test_compact_invalid_property
+    foo = [
+      [1],
+      [2],
+      [3]
+    ]
+
+    assert_raises Liquid::ArgumentError do
+      @filters.compact(foo, "bar")
+    end
   end
 
   def test_reverse
@@ -362,6 +410,29 @@ class StandardFiltersTest < Minitest::Test
 
   def test_map_works_on_enumerables
     assert_template_result "123", '{{ foo | map: "foo" }}', "foo" => TestEnumerable.new
+  end
+
+  def test_map_returns_empty_on_2d_input_array
+    foo = [
+      [1],
+      [2],
+      [3]
+    ]
+
+    assert_raises Liquid::ArgumentError do
+      @filters.map(foo, "bar")
+    end
+  end
+
+  def test_map_returns_empty_with_no_property
+    foo = [
+      [1],
+      [2],
+      [3]
+    ]
+    assert_raises Liquid::ArgumentError do
+      @filters.map(foo, nil)
+    end
   end
 
   def test_sort_works_on_enumerables


lib/liquid/standardfilters.rb
```diff
@@ -194,8 +194,12 @@ def map(input, property)
         if property == "to_liquid".freeze
           e
         elsif e.respond_to?(:[])
-          r = e[property]
-          r.is_a?(Proc) ? r.call : r
+          begin
+            r = e[property]
+            r.is_a?(Proc) ? r.call : r
```
Thibaut: This line should not be part of the `begin` block, or we may catch other errors unrelated to the `e[property]` method call.

lib/liquid/standardfilters.rb
```diff
@@ -194,8 +194,12 @@ def map(input, property)
...
+          rescue TypeError
```
Thibaut: Hmm I wonder if there's another way to catch this that doesn't involve a `rescue`.

lib/liquid/standardfilters.rb
```diff
@@ -194,8 +194,12 @@ def map(input, property)
...
+            nil
```
Thibaut: I don't think we should silently swallow the error. I'd raise `Liquid::ArgumentError`.

lib/liquid/standardfilters.rb
itsGarland: I discussed this with Sam, and `e[property]` must be called in order to verify its own existence. Otherwise, checking if `e` has attribute `property` beforehand requires knowledge of `property` type (e.g.; if `e` is an `Array` and `property` is a `string`, then by calling `e.[](property)` we pass in a string argument, which causes an error).

Thibaut:
@pushrax good point. Let's add a wrapper method. Something like this:
```ruby
get_property(item, property)

def get_property(item, property)
  item[property]
rescue TypeError 
  # Cannot index with the given property type (eg. indexing integers with strings 
  # which are only allowed to be indexed by other integers).
  raise ArgumentError.new("cannot select the property `#{property}`") 
end
```

Thoughts?

itsGarland:
@pushrax You bring up a good point how often this occurs. I can write a method like `property_exists?(input, property)` which I can put in some of the methods that do not handle invalid `property` inputs. However, I notice there are a lot of other methods that accept user input and do not throw errors (e.g.; passing `"abc"` as input into a `sort` filter does not cause an issue because it is wrapped in a `Inputiterator` object). So my question then is, should I standardize all methods to either throw/return the object back(or nil), or should I just throw those with user input for `property`?

pushrax:
Any input that can trigger a ruby (non-liquid) exception should be changed to a liquid exception, anything that would previously return nil should continue to return nil unless there's a great reason to change it.

itsGarland:
some notes I just wanted to make:
- There is a lack of standardizing between how certain functions return. That is fine as we would not want to make a breaking change by throwing ArgumentError in all circumstances.
- There is an assumption in some filters that the first element is of the same type (or acts like) the rest of the elements in the input array; therefore assuming that the element can respond to `[]` implies the remaining elements do, too.

lib/liquid/standardfilters.rb
```diff
@@ -188,15 +188,11 @@ def reverse(input)
...
-          e
-        elsif e.respond_to?(:[])
-          r = e[property]
-          r.is_a?(Proc) ? r.call : r
-        end
+      ary = InputIterator.new(input).map { |e| e.is_a?(Proc) ? e.call : e }
+      if property == "to_liquid".freeze
+        ary
+      elsif ary.first.respond_to?(:[])
+        ary.map { |e| (e = get_property(e, property)).is_a?(Proc) ? e.call : e }
```
Thibaut: Could we implement this without a double `map`? This change is making the method slower.

lib/liquid/standardfilters.rb
```diff
@@ -160,9 +160,9 @@ def where(input, property, target_value = nil)
       if ary.empty?
         []
       elsif ary.first.respond_to?(:[]) && target_value.nil?
-        ary.where_present(property)
+        ary.select { |item| get_property(item, property) }
```
samdoiron: Could we move this logic into `where` / `where_present` instead of here?

lib/liquid/standardfilters.rb
itsGarland: @Thibaut I was trying to mirror the format of the other methods. Is the time complexity that much different because it still runs linearly right?

pushrax: The bigger problem here is that this allocates an extra array now. This should be done without the extra array.

itsGarland:
I've updated my PR to leave `map` filter as is. Any other comments @pushrax  @Thibaut ?

lib/liquid/standardfilters.rb
pushrax: 100%, these methods are near the bottom of the file on `InputIterator`, not part of stdlib.

pushrax: As-is, this PR deletes those methods, which I think is fine now that `get_property` contains the logic.

lib/liquid/standardfilters.rb
```diff
@@ -394,6 +394,12 @@ def default(input, default_value = ''.freeze)
 
     private
 
+    def get_property(input, property)
+      input[property]
+    rescue
```
pushrax: Is there a reason you changed this from explicitly rescuing `TypeError`?

itsGarland: I think I just forgot. Thanks for catching that 👍

itsGarland:
@pushrax with the exception of `map`, 

```ruby
# Remove nils within an array
# provide optional property with which to check for nil
def compact(input, property = nil)

  if property.nil?
    ary.compact
    []
  elsif ary.first.respond_to?(:[])
    applesauce(Proc.new { ary.reject{ |a| a[property].nil? } }, property)
  end
end

def applesauce(proc, property)
  proc.call
rescue TypeError
  raise Liquid::ArgumentError.new("Cannot select the property '#{property}'")
end
```
how does this sound?

itsGarland:
@pushrax when you get time please tell me what you think

pushrax:
With a benchmark 👍 to that idea.

itsGarland:
@pushrax I've made the change. Please review when you're available.

lib/liquid/standardfilters.rb
```diff
@@ -394,6 +393,12 @@ def default(input, default_value = ''.freeze)
 
     private
 
+    def safe_property_check(property)
+      yield
+    rescue TypeError
+      raise Liquid::ArgumentError.new("Cannot select the property '#{property}'")
```
pushrax: nit: we interpolate errors as `Liquid error: <error message>`, and this reads more nicely with a lowercase error message.

pushrax:
```
jli@tomato ~/s/g/S/liquid git:master ❯❯❯ bx ruby add_rescue.rb
   no safe_property_check     1.264M i/s -      1.283M times in 1.015032s (791.43ns/i)
outer safe_property_check     1.238M i/s -      1.241M times in 1.001754s (807.45ns/i)
inner safe_property_check   868.640k i/s -    897.890k times in 1.033673s (1.15μs/i)
   no safe_property_check     1.317M i/s -      3.791M times in 2.879269s (759.58ns/i)
outer safe_property_check     1.229M i/s -      3.715M times in 3.023683s (813.82ns/i)
inner safe_property_check   872.386k i/s -      2.606M times in 2.987118s (1.15μs/i)

Comparison:
   no safe_property_check:   1316517.8 i/s
outer safe_property_check:   1228770.3 i/s - 1.07x  slower
inner safe_property_check:    872386.0 i/s - 1.51x  slower
```

```rb
require 'benchmark_driver'

Benchmark.driver do |x|
  x.prelude <<~RUBY
    array = ["a"] * 10
    property = "foo"

    def safe_property_check(property)
      yield
    rescue TypeError
      raise
    end
  RUBY

  x.report 'no safe_property_check', <<~RUBY
    array.map { |x| x }
  RUBY

  x.report 'outer safe_property_check', <<~RUBY
    safe_property_check(property) do
      array.map { |x| x }
    end
  RUBY

  x.report 'inner safe_property_check', <<~RUBY
    array.map do |x|
      safe_property_check(property) do
        x
      end
    end
  RUBY
end
```

itsGarland:
I played a bit with benchmark and inlining the rescue behaviour preserves current runtime:
```
   no safe_property_check     1.348M i/s -      1.396M times in 1.035210s (741.60ns/i)
outer safe_property_check     1.247M i/s -      1.268M times in 1.017346s (802.15ns/i)
inner safe_property_check   789.318k i/s -    820.575k times in 1.039600s (1.27μs/i)
   no safe_property_check     1.379M i/s -      4.045M times in 2.933226s (725.09ns/i)
outer safe_property_check     1.282M i/s -      3.740M times in 2.916987s (779.96ns/i)
inner safe_property_check   797.638k i/s -      2.368M times in 2.968707s (1.25μs/i)

Comparison:
   no safe_property_check:   1379136.1 i/s 
outer safe_property_check:   1282121.2 i/s - 1.08x  slower
inner safe_property_check:    797638.2 i/s - 1.73x  slower

OTT-Garland:test-folder itsgarland$ vi test-file.rb 
OTT-Garland:test-folder itsgarland$ ruby test-file.rb 
   no safe_property_check     1.340M i/s -      1.414M times in 1.054674s (746.14ns/i)
outer safe_property_check     1.270M i/s -      1.272M times in 1.001760s (787.54ns/i)
            inline rescue     1.405M i/s -      1.408M times in 1.002228s (711.58ns/i)
inner safe_property_check   795.751k i/s -    823.020k times in 1.034268s (1.26μs/i)
   no safe_property_check     1.381M i/s -      4.021M times in 2.912349s (724.34ns/i)
outer safe_property_check     1.297M i/s -      3.809M times in 2.935918s (770.72ns/i)
            inline rescue     1.380M i/s -      4.216M times in 3.055957s (724.86ns/i)
inner safe_property_check   799.202k i/s -      2.387M times in 2.987046s (1.25μs/i)

Comparison:
   no safe_property_check:   1380564.3 i/s 
            inline rescue:   1379582.2 i/s - 1.00x  slower
outer safe_property_check:   1297488.2 i/s - 1.06x  slower
inner safe_property_check:    799202.0 i/s - 1.73x  slower
```
of course that literally comes back to the original issue of duplicating logic everywhere. However since it is slightly slower using a method, would it just make sense to inline the `rescue` blocks anyway? From what I can tell, `e[property]` is only used as a return value, for comparison, or for type checking. I think then, there would unlikely be a future change in error handling (let alone functionality)...so the duplicated logic won't change that often. What are your thoughts?

pushrax:
Adding ~50ns overhead is probably okay for most real cases here, but it's not really much more code to do something like

```rb
def sort(input, property)
  ...
  elsif ary.all? { |el| el.respond_to?(:[]) }
    begin
      ary.sort { |a, b| nil_safe_compare(a[property], b[property]) }
    rescue TypeError
      raise_property_error(property)
    end
  end
end
```

Usually I wouldn't care so much, but I wouldn't be surprised if some Liquid templates use these filters in tight loops.

itsGarland:
@pushrax I rewrote it with just inlining the rescue blocks just to see what it looks like. It's messy but it does the job better since it performs faster..

does this look better or did I misunderstand your previous comment?

pushrax:
You could abstract the `raise` itself into a method, since that's the rare case.

itsGarland:
@pushrax 

```
Comparison:
                            no check:   1469819.0 i/s 
                        inline raise:   1466718.7 i/s - 1.00x  slower
inline begin block but extract raise:   1455266.1 i/s - 1.01x  slower
```
This is a lot better than before 👍.

lib/liquid/standardfilters.rb
```diff
@@ -176,7 +188,11 @@ def uniq(input, property = nil)
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
       elsif ary.first.respond_to?(:[])
-        ary.uniq{ |a| a[property] }
+        begin
+          ary.uniq{ |a| a[property] }
```
Thibaut: Space before `{`

lib/liquid/standardfilters.rb
```diff
@@ -210,7 +231,11 @@ def compact(input, property = nil)
       elsif ary.empty? # The next two cases assume a non-empty array.
         []
       elsif ary.first.respond_to?(:[])
-        ary.reject{ |a| a[property].nil? }
+        begin
+          ary.reject{ |a| a[property].nil? }
```
Thibaut: Space before `{`

lib/liquid/standardfilters.rb
Thibaut: 👍 ^

lib/liquid/standardfilters.rb
```diff
@@ -460,26 +489,9 @@ def each
...
-        # Cannot index with the given property type (eg. indexing integers with strings
-        # which are only allowed to be indexed by other integers).
-        raise ArgumentError.new("cannot select the property `#{property}`")
-      end
     end
   end
 
   Template.register_filter(StandardFilters)
 end
+
```
Thibaut: Remove ^

test/integration/standard_filter_test.rb
```diff
@@ -696,3 +767,4 @@ def with_timezone(tz)
     ENV['TZ'] = old_tz
   end
 end # StandardFiltersTest
+
```
Thibaut: Remove ^

lib/liquid/standardfilters.rb
```diff
@@ -198,6 +214,8 @@ def map(input, property)
           r.is_a?(Proc) ? r.call : r
         end
       end
+    rescue TypeError
+      raise_property_error(property)
```
itsGarland: In my previous iterations I stored the InputIterator instance in a variable to isolate the begin/rescue block from the initialization in order to only catch `TypeError` from the mapping functionality. However, after looking at the InputIterator `initialize` method, there doesn't seem to be a cause of TypeError to being raised from a user writing in Liquid.

lib/liquid/standardfilters.rb
```diff
@@ -394,6 +416,10 @@ def default(input, default_value = ''.freeze)
 
     private
 
+    def raise_property_error(property)
+      raise Liquid::ArgumentError.new("Liquid error: cannot select the property '#{property}'")
```
Thibaut: What @pushrax meant is that the `Liquid error:` prefix is added automatically further up the stack inside Shopify. This should be:
```ruby
raise Liquid::ArgumentError.new("cannot select the property '#{property}'")
```

pushrax: Ah good catch...

itsGarland:
eZ Pz.

PR #1062, opened 2019-02-03
Fix simple RuboCop offenses and update TODO file

ashmaroli:
*The RuboCop version used remains the same*


ashmaroli:
*I've signed the CLA now.*

ashmaroli:
TBH, I don't see the point of the `ode-review/policial` check on PRs because this repository has locked to a older version of RuboCop, whereas Policial checks use a more recent RuboCop version.

pushrax:
Yeah, the double RuboCop is unnecessary. The right solution is to remove RuboCop from the CI run and let Policial check, and update our ruleset based on the base style guide at https://github.com/Shopify/ruby-style-guide

ashmaroli:
@pushrax What I meant to say is that the current RuboCop versions and Policial versions are at odds with each other.
Since you feel that Policial checks need to take priority, then the right thing to do is to update the RuboCop config in sync with the version used by Policial. (Because current Policial failure is mainly because of config mismatch.)
I haven't used Policial ever. So don't know if you can configure Policial to use the same version locked to in this repository's `Gemfile` / `Gemfile.lock`... (like Hound CI)..

ashmaroli:
Additionally, https://github.com/Shopify/ruby-style-guide seems to be a private repository that I cannot access.

pushrax:
Hm, I guess the repository is private but the pages endpoint is not https://shopify.github.io/ruby-style-guide/ and https://shopify.github.io/ruby-style-guide/rubocop.yml

I like Policial since it nicely integrates with Github, making review faster. You’re exactly right, the config needs to change, since Policial will look at .rubocop.yml in the repo root.

I don’t mean to suggest you need to do this, my last comment was mostly thinking out loud.

pushrax:
It’s hilarious that we have “Avoid bugs.“ in the style guide.

PR #1066, opened 2019-02-24
Autoload Liquid modules and classes

ashmaroli:
To minimize allocations from simply calling `require 'liquid'` in a program.
For example, there is no need to allocate memory for loading `Liquid::ForloopDrop` *until* a `Liquid::For` instance is being rendered.

Files that define multiple modules / classes are loaded immediately.


ashmaroli:
Ah! I hadn't considered the Shopify environment and interactions there..

PR #1070, opened 2019-03-07
Validate the character encoding in url_decode.

clayton-shopify:
The `url_decode` filter can produce arbitrary byte sequences, including ones that are not valid for the current character encoding. Applications using Liquid most likely expect its output to be validly encoded. Therefore, it seems useful to raise an error in cases where `url_decode` produces an invalid byte sequence.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -52,7 +52,12 @@ module Liquid
     end
 
     def url_decode(input)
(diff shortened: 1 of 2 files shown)


clayton-shopify:
> Can merge without fixing Rubocop

OK, I removed that commit.

pushrax:
I mainly meant not to worry about making the build green, but that works too :)

PR #1071, opened 2019-03-11
Fix large for-loop collection

stayhero:
When a large Range is given in a for-loop (e.g. 1..100000000000000), it may crash the Ruby process or at least eats a lot of memory and takes a significant amount of time to render the template.

I added the following testcase to template_test.rb:

```
    t = Template.parse("{% for a in (1..100000000000000) %} {% for a in (1..10) %} foo {% endfor %} {% endfor %}")
    t.resource_limits.render_score_limit = 50
    assert_equal "Liquid error: Memory limits exceeded", t.render
    assert t.resource_limits.reached?
```

The render_score_limit does not help here because it's more of a Ruby performance problem in for.rb, where the `collection = collection.to_a if collection.is_a?(Range)` happens. 

A fix may be to check if the given Range is larger than the given render_score_limit. I'm actually not sure if there may be a better fix (e.g. checking the resource limits somewhere else) but at least it works for the given test case.

### DIFF
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -1,5 +1,4 @@
 module Liquid
-
   # "For" iterates over an array or collection.
   # Several useful variables are available to you within the loop.
   #
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -1,4 +1,5 @@
 module Liquid
+
   # "For" iterates over an array or collection.
   # Several useful variables are available to you within the loop.
   #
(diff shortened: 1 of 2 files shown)


stayhero:
BTW: You can checkout this branch to show the behavior for a large given collection without my fix:

https://github.com/stayhero/liquid/tree/large_forloop_failing_test

stayhero:
@pushrax Anything missing to merge the PR?

fw42:
@pushrax: ping

devdreamsolution (approved):

lib/liquid/tags/for.rb
```diff
@@ -128,7 +128,14 @@ def collection_segment(context)
       end
 
       collection = context.evaluate(@collection_name)
-      collection = collection.to_a if collection.is_a?(Range)
+      if collection.is_a?(Range)
+        score_limit = context.resource_limits.render_score_limit
+        if score_limit && score_limit < collection.size
+          context.resource_limits.render_score += collection.size
+          raise MemoryError.new("Memory limits exceeded".freeze)
```
ianks, line 135: The general idea here is correct, but this will cause a subtle behavior difference. The semantics are that rendering should continue _until_ the render score is reached - meaning the first `0..RENDER_SCORE` items should be rendered before the memory error.

Instead, we should do something like:

```ruby
if score_limit && score_limit < collection.size
  collection = collection.begin..score_limit
end
```

If the render score is too high still, we may need to adjust `Liquid::Utils.slice_collection` so that it doesn't accidentally `to_a` to range.

PR #1073, opened 2019-03-13
defer hash allocation until needed for unparsed_args

richardmonette:
### Summary

Depending on whether there are any `unparsed_args` args we may be performing unnecessary hash allocations in the `parse_filter_expressions` method. This may seem trivial, but given the frequency with which this method is called, this may actually be putting extra strain on the GC.

### Performance

I did not want to propose this potential optimization unless I had some numbers to back it up, so I did a simple CPU benchmark and checked on the GC stats to see if allocations were improved. (I removed the test I was using to run the performance check in the last commit of this branch, but you can check the second commit if you'd like to see my performance test setup.)

```bash
pre alloc w kwargs  0.880000   0.000000   0.880000 (  0.892699)
allocated_objects: 2600059
deferred alloc w kwargs  0.890000   0.010000   0.900000 (  0.889087)
allocated_objects: 2600050
pre alloc wo kwargs  0.030000   0.000000   0.030000 (  0.032060)
allocated_objects: 500050
deferred alloc wo kwargs  0.030000   0.000000   0.030000 (  0.027086)
allocated_objects: 400049
pre alloc w arg  0.230000   0.000000   0.230000 (  0.239780)
allocated_objects: 1100050
deferred alloc w arg  0.240000   0.000000   0.240000 (  0.233269)
allocated_objects: 1000050
```

Performance-wise, the change is mostly negligible, with the suggestion of some improvement in the case of no `unparsed_args`. 😐

However, on the GC/allocation side, again in the case of no `unparsed_args`, we clearly are making less allocations. 😄

### Conclusion

Given that this change should have zero user facing implications, I think it might be worth while.

### DIFF
--- a/lib/liquid/variable.rb
+++ b/lib/liquid/variable.rb
@@ -95,16 +95,17 @@ module Liquid
 
     def parse_filter_expressions(filter_name, unparsed_args)
       filter_args = []
-      keyword_args = {}
+      keyword_args = nil
       unparsed_args.each do |a|
         if matches = a.match(JustTagAttributes)
+          keyword_args ||= {}
           keyword_args[matches[1]] = Expression.parse(matches[2])
         else
           filter_args << Expression.parse(a)
         end
       end
       result = [filter_name, filter_args]
-      result << keyword_args unless keyword_args.empty?
+      result << keyword_args if keyword_args
       result
     end
 


lib/liquid/variable.rb
```diff
@@ -94,17 +94,15 @@ def render(context)
...
+          result << {} unless result[2]
+          result[2][matches[1]] = Expression.parse(matches[2])
         else
-          filter_args << Expression.parse(a)
+          result[1] << Expression.parse(a)
         end
       end
-      result = [filter_name, filter_args]
-      result << keyword_args unless keyword_args.empty?
       result
```
pushrax: I'd find this a bit easier to read with locals for `filter_args` and `keyword_args` and the last line allocate the array of `[filter_name, filter_args, keyword_args]`. It's fine to have the nil 3rd element, performance would be nearly the same or slightly better ([no memory cost on MRI since the array is 3 elements or less](https://gist.github.com/pushrax/bd2d3bb99c46ed0570cfbb7d5e82d742#file-memalloc-rb-L21-L31), and locals are faster to access than array members). The point is just readability though.

lib/liquid/variable.rb
```diff
@@ -95,17 +95,16 @@ def render(context)
...
+          keyword_args = {} unless keyword_args
           keyword_args[matches[1]] = Expression.parse(matches[2])
         else
           filter_args << Expression.parse(a)
         end
       end
-      result = [filter_name, filter_args]
-      result << keyword_args unless keyword_args.empty?
-      result
+      [filter_name, filter_args, keyword_args].compact
```
Thibaut: Isn't this part slower than before?

richardmonette: Replaced `compact` with the old `if` approach

```
pre alloc w kwargs  0.840000   0.010000   0.850000 (  0.842502)
allocated_objects: 2600059
deferred alloc w kwargs  0.800000   0.000000   0.800000 (  0.804708)
allocated_objects: 2600050
pre alloc wo kwargs  0.030000   0.000000   0.030000 (  0.028275)
allocated_objects: 500050
deferred alloc wo kwargs  0.030000   0.000000   0.030000 (  0.030198)
allocated_objects: 400049
pre alloc w arg  0.220000   0.010000   0.230000 (  0.224994)
allocated_objects: 1100050
deferred alloc w arg  0.220000   0.000000   0.220000 (  0.216734)
allocated_objects: 1000050
```

pushrax: We don't need `if` or `compact`, let's just let the value be nil. Either way it's a 40 byte object without extra heap.

richardmonette: > We don't need if or compact, let's just let the value be nil.

```
  1) Failure:
VariableUnitTest#test_lax_filter_argument_parsing [/Users/richardmonette/src/github.com/Shopify/liquid/test/unit/variable_unit_test.rb:203]:
--- expected
+++ actual
@@ -1 +1 @@
-[["pluralize", ["comment", "comments"]]]
+[["pluralize", ["comment", "comments"], nil]]


  2) Failure:
VariableUnitTest#test_string_to_filter [/Users/richardmonette/src/github.com/Shopify/liquid/test/unit/variable_unit_test.rb:150]:
Expected: [["image", ["med"]]]
  Actual: [["image", ["med"], nil]]

  3) Failure:
VariableUnitTest#test_filters_without_whitespace [/Users/richardmonette/src/github.com/Shopify/liquid/test/unit/variable_unit_test.rb:130]:
--- expected
+++ actual
@@ -1 +1 @@
-[["textileze", []], ["paragraph", []]]
+[["textileze", [], nil], ["paragraph", [], nil]]


  4) Failure:
VariableUnitTest#test_filter_with_date_parameter [/Users/richardmonette/src/github.com/Shopify/liquid/test/unit/variable_unit_test.rb:124]:
Expected: [["date", ["%m/%d/%Y"]]]
  Actual: [["date", ["%m/%d/%Y"], nil]]

  5) Failure:
VariableUnitTest#test_symbol [/Users/richardmonette/src/github.com/Shopify/liquid/test/unit/variable_unit_test.rb:144]:
Expected: [["image", ["med"]]]
  Actual: [["image", ["med"], nil]]

  6) Failure:
VariableUnitTest#test_filters [/Users/richardmonette/src/github.com/Shopify/liquid/test/unit/variable_unit_test.rb:82]:
Expected: [["textileze", []]]
  Actual: [["textileze", [], nil]]
```

We have tests which assume the result array 2 element, without `nil`. 

I guess we could change that, but I was hoping to avoid modifying any assumptions made by our test suite for this minor perf optimization.

pushrax: This is all internal stuff anyway, it's mostly about clarity and I would just change the tests. Like you say doesn't really matter too much either way, so feel free to do this or not and merge without further review.

PR #1077, opened 2019-03-18
update Rubocop for trailing comma styles

richardmonette:
Quick fix for:

```
The `Style/TrailingCommaInLiteral` cop no longer exists. Please use `Style/TrailingCommaInArrayLiteral` and/or `Style/TrailingCommaInHashLiteral` instead.
(obsolete configuration found in .rubocop.yml20190318-51-xlf5g6, please update it)
```

Since this was set as disabled, assuming we want it disabled in both cases.


ashmaroli:
@richardmonette While your fix appeases Policial (it uses a newer version of RuboCop), note that the Gemfile in this repository locks to an older version of RuboCop:
https://github.com/Shopify/liquid/blob/f2f467bdbc248e7bf26b4f5e552cfbce6052f811/Gemfile#L14-L15
and therefore Travis builds now fail with RuboCop flagging the offenses (since both `Style/TrailingCommaInArrayLiteral` and `Style/TrailingCommaInHashLiteral` are unknown to the RuboCop program we're using.)

So, the decision should ideally be to either configure Policial to use the RuboCop version locked to in the repository or to update RuboCop version used in the repository to one that has `Style/TrailingCommaInArrayLiteral` and `Style/TrailingCommaInHashLiteral` implemented.

richardmonette:
I went with 0.53.0 https://github.com/Shopify/liquid/pull/1077/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R15 for two reasons:

1. That's the version at which the change in question happened https://github.com/rubocop-hq/rubocop/issues/5404
2. When I tried the very latest, 0.65.0, I was getting new Rubocop warnings about only supporting Ruby >= 2.2, which means an additional bump for that (we're at 2.1 currently.) Bumping Ruby _feels_ like a bigger change, so I wanted to try and avoid doing that for just fixing a simple linting error.

PR #1080, opened 2019-03-24
Memoize Liquid::Tag#raw

ashmaroli:
This change is avoid unnecessary String allocations.

## Theory

- Since interpolated strings can't be frozen, they generate a new object for each call
- Not every tag is initialized with a `markup`. For example, `{% comment %}`. In such cases, the string resulting from `Tag#raw` has a trailing whitespace (`"comment "`) which has to be stripped away which may or may not result in a new string object based on the method used.

Memoizing the resulting object resolves the above two issues.

### DIFF
--- a/lib/liquid/tag.rb
+++ b/lib/liquid/tag.rb
@@ -25,7 +25,7 @@ module Liquid
     end
 
     def raw
-      "#{@tag_name} #{@markup}"
+      @raw ||= "#{@tag_name} #{@markup}".strip
     end
 
     def name


ashmaroli:
Unfortunately, this will be a breaking-change for those who are using the `#raw` method chained with `String#split!`. For example, `{% #{raw.split!} %}` will now return `{%  %}` :disappointed: 

So this can't be merged in until `liquid-5.0.0`......

PR #1084, opened 2019-04-01
Allow method literals to provide useful values

er1:
This allows the `empty` literal to act as an empty array when being used outside of comparisons.
It would make empty array constructions such as `{% assign x = '' | split: '' %}` more readable as `{% assign x = empty %}`
This should not impact existing uses since empty array will get coerced into empty string if and where it is being used.


ashmaroli:
@er1 This looks like a breaking-change to me. Can you test if the empty array can be *filled* later on..?
```
{% assign dropship = empty %}
{% if product.price < store.sale_price %}
  {{ dropship | push: product }}
{% endif %}
```

er1:
@ashmaroli It would depend on how that filter works. I am assuming that values are immutable and that the standard filters coerce to arrays when they need arrays before performing operations. This would be in line with how enumerable drops behave.

test/integration/variable_test.rb
```diff
@@ -34,6 +34,33 @@ def test_using_blank_as_variable_name
...
+  end
+
+  def test_empty_can_be_filled
+    template = Template.parse("{% assign foo = empty %}{{ foo | concat: bar }}")
+    assert_equal('Y', template.render!('bar' => ['Y']))
+    template = Template.parse("{% assign foo = empty %}{{ foo | concat: bar | join: '--' }}")
+    assert_equal('A--B', template.render!('bar' => ['A', 'B']))
+    template = Template.parse("{% assign foo = empty %}{% assign tar = foo | concat: bar %}{{ tar | join: '--' }}{{ tar | size }}")
+    assert_equal('A--B2', template.render!('bar' => ['A', 'B']))
   end
```
er1: @ashmaroli would using the concat filter here be enough to check if arrays from the empty literal can be filled?

ashmaroli: Yes, that is sufficient.

ashmaroli:
@er1 IMO, `{% assign x = empty %}` doesn't improve readability. It isn't clear if `x` is an empty string or an empty array. It will take time for existing users to adjust to the fact that `empty` is a flexible entity now.
Would it be possible to introduce `[]` or `empty_array` instead..?

er1:
> Would it be possible to introduce `[]` or `empty_array` instead..?
@ashmaroli I agree that it looks a little unclear at first glance however I don't feel comfortable suggesting new literal names into the language.
Existing code might have `empty_array` set via assigns and introducing a new literal this way will break that code since literals cannot be assigned.
Using `[]` doesn't doesn't feel right unless we flat out introduce array literals into the language as I think it would suggest that something like `[2,3,5]` or `[product1, product2]` could be done and that is a much large discussion to have.

ashmaroli:
> Using `[]` doesn't doesn't feel right unless we flat out introduce array literals into the language

I see. Thanks for the feedback.

devdreamsolution (approved):

er1:
failing tests in df49715 depend on Shopify/liquid-c#121

PR #1091, opened 2019-04-17
Rendering with less garbage

fw42:
This is not ready to be merged at all and is just playing around with an idea that I was curious about and wanted to try out. Would love some feedback.

There's three main ideas in this PR:

* (1) **Render directly to a string buffer**: Write directly to the same string buffer while rendering every node rather than each `render` call allocating it's own strings and then joining them together when rendering a block.
* (2) **Preallocate memory for that buffer**: Make use of Ruby 2.4's [String.new(capacity: ...)](https://bugs.ruby-lang.org/issues/12024) feature to preallocate a string that's bigger than we currently need it to be. Idea being that this will hopefully save a bunch of underlying [realloc(3)](https://linux.die.net/man/3/realloc) system calls because the string doesn't need to be dynamically resized as often.
* (3) **Re-use that buffer across templates**: Re-use the same string buffer that we used when rendering previous templates, i.e. instead of allocating a new buffer for each request, we just truncate the old one to length 0 ([rb_str_clear](https://github.com/ruby/ruby/blob/e1eb54b99d592c460fea4b0f897f5e8e61c96c07/string.c#L5354-L5366)).

The code obviously sucks for many reasons, needs to be cleaned up, is not backwards compatible, etc., so please don't code review individual lines here.

### bundle exec rake memory_profiler:run

<table>
  <tr>
    <td></td>
    <th>master</th>
    <th>(1)</th>
    <th>(1)+(2)+(3)</th>
  </tr>
  <tr>
    <th>Objects allocated during rendering</th>
    <td>10,089</td>
    <td>9,274 (-8%)</td>
    <td>9,274</td>
  </tr>
  <tr>
    <th>Bytes allocated during rendering</th>
    <td>1,154,098</td>
    <td>955,894 (-17%)</td>
    <td>595,822 (-48%)</td>
  </tr>
</table>

I didn't have any luck to get stable numbers with `rake benchmark:run`, so I don't trust that command. It does appear (!) to be a bit faster than before though.

I also tried profiling this (with stackprof) on a real store's theme and it shows that garbage collection time was reduced by about half, however overall rendering time increased for some reason (I imagine some Ruby internal detail that I don't understand currently makes this approach slower for some reason... might also be that the size of the string buffer that I chose wasn't ideal).

Any thoughts on this? Worth pursuing? Did someone try this yet? Any ideas how to profile this? Any ideas why overall rendering performance (cpu time) might be slower?


lib/liquid/tags/capture.rb
```diff
@@ -22,11 +22,11 @@ def initialize(tag_name, markup, options)
...
     end
 
-    def render(context)
-      output = super
+    def render(context, output = '')
+      super
       context.scopes.last[@to] = output
-      context.resource_limits.assign_score += output.length
-      ''.freeze
+      context.resource_limits.assign_score = output.length
```
fw42: Note to self for later: This looks like a logic bug.

dylanahsmith:
> Did someone try this yet?

Sounds a lot like what I tried in https://github.com/Shopify/liquid/pull/319 which you even had reviewed

> Any ideas why overall rendering performance (cpu time) might be slower?

I came to a similar conclusion.  I'm not sure why it would be slower, but I lost interest once I realized that it wasn't really fixing a bottleneck

fw42:
Ha. Of course you already tried this, Dylan 😆Although at the time there was no `String.new(capacity: ...)`, so maybe it's still worth investigating this?

Maybe I made the same mistake as you somewhere (forgot to fix some tag and now the results are skewed). Will investigate.

felix-d:
Mmm I'm not sure either why it's slower. I have no knowledge of liquid but in theory in shouldn't be. In one case you copy bytes to a buffer, in the other you allocate then copy bytes to the buffer. I think this idea is worth pursuing, maybe with an exponentially growing buffer or something instead of a fixed length provided by the client.

fw42:
I added another commit that made the performance difference negligible on the production theme I was testing with. It's actually a bit faster now with significantly fewer memory allocations.

Will post more numbers once I figure out how to properly profile this.

ashmaroli:
> once I figure out how to properly profile this.

To profile speed improvements, you can use the `hold!` attribute via the `benchmark-ips` gem. The steps can be roughly outlined as follows. (I have a BASH script to aid me locally..)

1. Set up Benchmark script
```ruby
require 'benchmark/ips'
# other requires........

# [any helper code..]

Benchmark.ips do |x|
  x.report("Proposed") { run_proposed }
  x.report("Master") { run_master }
  x.hold! '.somename'
  x.compare!
end
```
2. Run benchmark script once (ensure you're currently in the feature branch).
3. Checkout the `master` branch locally.
4. Run the benchmark script once again to complete it.

fw42:
Thanks, but my problem isn't so much how to benchmark two versions of the code against each other, but how to get the benchmark to be stable in the first place on a single version... Running `rake benchmark:run` in the Liquid repo multiple times (even on the same revision of the code) gives me different results every time (different enough to not trust the numbers). Might just be that my laptop is too "noisy" (too much other stuff going on) or something. Will take another look tomorrow.

evulse:
I know ruby is a completely different beast, but if it helps to provide any insight. Applying Byte buffer pools in Go completely halved render time for my Go version of Liquid. It was probably one of the most dramatic increase in performance from a single change.

The Go library if it is of any help. https://github.com/valyala/bytebufferpool - built by the fasthttp team.

fw42:
Thanks Mike, I appreciate the input and will check it out!

fw42:
I implemented @pushrax's suggested API which should now guarantee backwards compatibility. I think this should be ready for a first round of review now. For the time being, I removed (2) and (3) (see PR description) and this PR at the moment only implements (1) (which was the vast majority of the changes anyway).

I'm not sure yet if (2) and (3) should be Liquid's responsibility or the caller's responsibility. Will think about that a bit more (and as Mike mentioned, we might need some kind of buffer pool first anyway).

Can I get some eyes on the changes here so far?

lib/liquid/profiler/hooks.rb
```diff
@@ -19,5 +19,14 @@ def render_with_profiling(context)
...
     alias_method :render, :render_with_profiling
+
+    def render_to_output_buffer_with_profiling(context, output)
+      Profiler.profile_children(context.evaluate(@template_name_expr).to_s) do
+        render_to_output_buffer_without_profiling(context, output)
+      end
+    end
+
+    alias_method :render_to_output_buffer_without_profiling, :render_to_output_buffer
+    alias_method :render_to_output_buffer, :render_to_output_buffer_with_profiling
```
pushrax: Do we still need `render_with_profiling`, given `Include#render` is now not valid?

lib/liquid/tag.rb
```diff
@@ -33,7 +33,15 @@ def name
     end
 
     def render(_context)
-      ''.freeze
```
pushrax: Unintentionally unfrozen?

lib/liquid/tags/ifchanged.rb
```diff
@@ -1,16 +1,16 @@
 module Liquid
   class Ifchanged < Block
-    def render(context)
+    def render_to_output_buffer(context, output)
       context.stack do
-        output = super
+        block_output = super(context, '')
```
pushrax: It isn't documented that `render_to_output_buffer` must return the buffer. It seems unnecessary to require that, unless we wanted a way to replace the buffer completely.

lib/liquid/variable.rb
```diff
@@ -85,12 +85,25 @@ def render(context)
...
+      obj = render(context)
+
+      if obj.is_a?(Array)
+        output << obj.join
+      elsif obj.nil?
+      elsif !obj.is_a?(String)
+        output << obj.to_s
+      else
+        output << obj
+      end
```
pushrax:
```rb
      elsif !obj.is_a?(String)
        output << obj.to_s
      else
        output << obj
      end
```

could be

```rb
      else
        output << obj.to_s
      end
```

Were you wanting to avoid duplication if we get a string subclass?
https://github.com/ruby/ruby/blob/cf930985da1ae15b66577ec8286e54ad04ccda78/string.c#L5765-L5772

lib/liquid/block_body.rb
```diff
@@ -67,19 +67,23 @@ def blank?
...
         case node
         when String
-          check_resources(context, node)
           output << node
         when Variable
-          render_node_to_output(node, output, context)
+          render_node(context, output, node)
         when Block
-          render_node_to_output(node, output, context, node.blank?)
+          render_node(context, node.blank? ? '' : output, node)
```
pushrax: I've always found blank handling a bit confusing, what do you think about making the block responsible for not writing anything if it's blank, instead of writing to a temporary buffer that we'll just throw away? Can ignore for this PR in any case.

lib/liquid/block_body.rb
```diff
@@ -88,34 +92,30 @@ def render(context)
...
-      output.join
+      output
     end
 
     private
 
-    def render_node_to_output(node, output, context, skip_output = false)
-      node_output = node.render(context)
-      node_output = node_output.is_a?(Array) ? node_output.join : node_output.to_s
-      check_resources(context, node_output)
```
pushrax: No longer necessary to check resource usage in the `skip_output` case because it'll be checked by the BlockBody of the node being rendered?

ashmaroli:
@fw42 Given that Liquid is deprecating `Tag#render` in favor of `#render_to_output_buffer`, how best can the Jekyll docs be updated to inform novice Rubyists on how to pass a "buffer" argument to the new method..?

fw42:
@ashmaroli: For now, this PR shouldn't break any custom tags that Jekyll might have (I added a backwards compatible `render` method). They wont get the speed benefits of this PR though unless they do convert their tag implementations.

lib/liquid/block_body.rb
fw42: Yea agree, have it on my TODO list to look into that as a follow-up because I saw this `''` allocation come up in the memory profile quite a lot. Will ignore for this PR.

fw42:
Addressed the comments

fw42:
@dylanahsmith, do you have any opinions?

lib/liquid/block_body.rb
pushrax: The resource usage isn’t checked here in the “child node is empty” case now, an empty string is passed as the output buffer and thrown away without checking its size. But this seems fine since the “empty” node will run BlockBody logic itself and increment the render score. My question was aimed at confirming this theory.

ashmaroli:
> They wont get the speed benefits of this PR though unless they do convert their tag implementations.

@fw42 My question was regarding exactly this..
How best should a custom tag be converted to get the speed benefits? That is not clear from the diff here..
- Is the `buffer` just a regular unfrozen string? (**`output = ''`**)
- Is the `buffer` a Ruby 2.4 string with prealloc capacity? (**`output = String.new('', capacity: 2048)`**)

evulse:
If you go down the buffer pool route which I highly suggest. I'd take the same approach fasthttp has and buffer length becomes self aware and optimises over time specific to the application. So the prealloc capacity becomes a dynamic value based on the average size used by the application. This way you're not over or under allocating and finding that perfect spot to avoid memory allocations during execution and in turn speeding liquid up as much as possible without needing to deploy a new update each time you decide to tune your buffer pool size up and down.

evulse:
The one comment I have about this would be why make output an additional parameter, in hewly I just put the buffer inside of context as that way its available to anything with context access.

fw42:
@ashmaroli: I believe the main benefit from this PR comes from using any kind of buffer at all, so while `String.new(capacity: something)` will probably save you some `realloc()` system calls under the hood, just using `''` will already give you a big reduction in memory allocation.

If your application doesn't use any custom tags then this should all just work automatically (but if you can actually try this out and let me know, that would be appreciated). If you do use custom tags, then you have to change your `render` methods (returns the rendered output) to a `render_to_output_buffer` methods (writes the rendered output into the given buffer).

So something like
```ruby
class MyCustomTag < Tag
  def render(context)
    "foobar"
  end
end
```
needs to be converted to something like
```ruby
class MyCustomTag < Tag
  def render_to_output_buffer(context, buffer)
    buffer << "foobar"
  end
end
```

fw42:
@evulse: Thanks for your input. Using some kind of `context.buffer` instead of passing it around sounds like it might simplify the code. I might experiment with that today to see how it looks. @pushrax, any opinions on that?

As far the string buffer pool goes: My benchmarks show a good advantage without it already, but I will definitely experiment with that in a follow-up PR. Although I still feel like that should probably be the application's responsibility, not the library's. What do others think?

Not sure how easy that is to implement in Ruby though since, as far as I can tell, Ruby doesn't really give you a lot of control over expanding/shrinking a String's allocated memory (you can only specify the initial capacity, and then expanding happens automatically by a factor of 2, and shrinking doesn't seem possible at all - but please correct me if I'm wrong on that; Might be worth experimenting with though or adding something custom via a C extension). Might also be worth experimenting with @dylanahsmith's [ruby-string-buffer](https://github.com/dylanahsmith/ruby-string-buffer), which looks similar.

evulse:
Go doesn't shrink as well, how that library works is it discards any buffers that become bigger than the average and any newly created buffers will be created at the average size.  It looks like ruby has a rewind command which is kind of similar to reset on a golang byte buffer. So when you put a string buffer back in to the pool you just rewind it first, then the next writer will grab it and start writing over top of the existing usage that already has capacity allocated. Obviously if not done right you output the entire old string so its something that needs to be tested well to make sure your not leaking existing data. 

I'd usually think no its not part of the package but liquid is a template language and output is its primary function so optimising that I think is within scope. 

I agree you're getting a good performance impact just from using buffers, but pools are where the real improvements happen and its why fasthttp in golang is the fastest http server in benchmarks by far. Fasthttp highlight it pretty well under their performance section -  "sync.Pool is your best friend.". Once you can stop garbage collecting your memory usage and reuse it instead you really hit some significant performance gains.

evulse:
With that said I do admit my implementation even put the byte buffer pool outside of the language implementation. It just took a writer interface. So maybe you're right that by just allowing a buffer is as much as the language needs to know.

lib/liquid/tags/capture.rb
tobi: yea that doesn't look right

dylanahsmith:
> as far as I can tell, Ruby doesn't really give you a lot of control over expanding/shrinking a String's allocated memory (you can only specify the initial capacity, and then expanding happens automatically by a factor of 2, and shrinking doesn't seem possible at all - but please correct me if I'm wrong on that; Might be worth experimenting with though or adding something custom via a C extension). Might also be worth experimenting with @dylanahsmith's ruby-string-buffer, which looks similar.

The only think that library provides that needs the C API right now is the current string capacity, so it won't be very useful at the moment.  If we want to optimize anything using the ruby C API, then we should just do it in liquid-c.  However, feel free to use the library to experiment with the C string APIs.

If you want to expand a string capacity in the ruby C API, then you can use `rb_str_modify_expand` for that.

If we want to shrink the string capacity when returning it to a buffer pool (i.e. you don't care about the contents of the string), then you could use `rb_str_resize(str, 0)` to free the string buffer (making it an embedded string) and then use `rb_str_modify_expand` to set the capacity again.

I'm not sure if there is a more direct way to shrink the string capacity if you care about the string contents, but I don't think we need that in this case.

> do you have any opinions?

It seems like the combination of these ideas should move us in the right direction, especially when we are able to re-use string buffers across renders that typically don't need to be resized.  However, we should make sure we actually improve the CPU benchmark to validate the approach and to catch mistaken assumptions we have made.

> (3) Re-use that buffer across templates: Re-use the same string buffer that we used when rendering previous templates, i.e. instead of allocating a new buffer for each request, we just truncate the old one to length 0 (rb_str_clear).

This is a great example of a mistaken assumption.  String#clear (i.e. `rb_str_clear`) actually frees the string buffer (just not the string object).  This also means the string wouldn't be expanded to the expected capacity needed when we re-use it, negating the benefit of idea (2).

> (1) Render directly to a string buffer: Write directly to the same string buffer while rendering every node rather than each render call allocating it's own strings and then joining them together when rendering a block.

Note that we aren't necessary always writing directly to the string buffer.  If we need to do something like `output << value.to_s` where `value` isn't already a string, then we are already creating an intermediate string.  Ruby's API is biased towards this pattern, so it can be hard to actually avoid intermediate strings like this without dropping to C.  Given that we have these intermediate strings, arrays actually have an advantage of reducing how much memory typically need to be re-sized when it needs to be expanded, because it is just expanding a pointer array rather than a large string array.  I think this is why this idea on its own isn't necessarily better.

evulse:
> Note that we aren't necessary always writing directly to the string buffer. If we need to do something like output << value.to_s where value isn't already a string, then we are already creating an intermediate string. Ruby's API is biased towards this pattern, so it can be hard to actually avoid intermediate strings like this without dropping to C. Given that we have these intermediate strings, arrays actually have an advantage of reducing how much memory typically need to be re-sized when it needs to be expanded, because it is just expanding a pointer array rather than a large string array. I think this is why this idea on its own isn't necessarily better.

I toyed with this exact problem and its a case of theoretically an array makes sense however practically how liquid templates are used makes that null and void. The reasoning is that well over 90% of what runs through Liquid templates especially in Shopify's case is actually the HTML and not even the dynamic content your injecting. So simply,  by optimising solely for the raw HTML you actually gain the largest benefit out of all possible optimisations as its the largest part of the system. If you can get the HTML part of Liquid to render without allocations you're in an amazing position.

Once you have solved that largest segment of HTML tags the second largest optimisation is the {{ variable }}  output aka dynamic variable resolution (map access). This can also be optimised pretty far as Maps are usually hash maps under the covers. So the fact that you know 95% of the string variables up front means you can precompute the hashes which saves most of the compute as the hashing algorithm is what takes up all the time when accessing a hash map.

I get the fun job of not being at all tied to Liquid commercially so I get to point out the obvious and make stupid comments. I'm sure there is a block button somewhere on a repo haha

dylanahsmith:
raw HTML only allocates strings as part of parsing, which is mostly just needed due to BlockBody#render not being implemented in C.  Appending one of those string to an array doesn't require any allocation, it just appends to an array buffer instead of a string buffer as this PR proposes.

An alternative to this PR would be to append to a single flat array instead of a string.  That array could also be pre-allocated and could be re-used across template renders.  This might provide better performance than using a pre-allocated string if that string needs to be reallocated.  However, I think that would just make it harder to optimize things with liquid-c in the future, since we would need to allocate ruby objects to store them in the array rather than writing directly to the output string.

> Once you have solved that largest segment of HTML tags the second largest optimisation is the {{ variable }} output aka dynamic variable resolution (map access). This can also be optimised pretty far as Maps are usually hash maps under the covers.

If the variable name is hard-coded, then there is no reason why we need a hash map to store those variables.  We could store those variables in an array, assign an offset for those variables at parse time then just use an array lookup at render time.

evulse:
> raw HTML only allocates strings as part of parsing, which is mostly just needed due to BlockBody#render not being implemented in C. Appending one of those string to an array doesn't require any allocation, it just appends to an array buffer instead of a string buffer as this PR proposes.

100% yes it can be optimised to an incredible level, but realistically it's not. I agree with the array approach still somewhat as I think if done right as your sort of talking about you can actually keep the entire array with an exact allocation due to the fact you know which blocks output, you can even prefill most of the array with html and not even need to empty it as it won't change, the only unknown are for loops as you don't know the length until render time, however at a top level you can make these for loops render down to one element or even have an array of arrays to deal with it

> If the variable name is hard-coded, then there is no reason why we need a hash map to store those variables. We could store those variables in an array, assign an offset for those variables at parse time then just use an array lookup at render time.

I agree with this completely, for the top level resolution this is 100% the ultimate solution. Liquid is unmatched if you applied all these optimisations. This is why I got excited when I started working on it. Apart from filters which can at best be cached almost everything can be optimised to an absolute extreme. 

People go on about React etc being the future of the web, however it suffers one major flaw and that is that for every blocking logic decision such as resolving an if statement it needs an external call which is at best 35ms latency. You apply a for loop with an if statement and your performance goes out the window. However, If you do all of what has been talked about in here to Liquid your unmatched as you've removed the latency issue all together with server side rendering.

fw42:
If there are no objections then I'm going to merge this later today.

fw42:
We've been running this branch in production for a few weeks now. Looks all good so far.

evulse:
🎉 nice work

ashmaroli:
@fw42 The following are the effects of profiling Jekyll's docs site with `liquid-4.0.3` and `liquid:master@248c54a3864e`:

### Profiler Summary

```diff
- liquid-4.0.3      https://travis-ci.org/jekyll/jekyll/jobs/560387810
+ liquid:master     https://travis-ci.org/jekyll/jekyll/jobs/561100244

- Total allocated: 481.89 MB (4552897 objects)
+ Total allocated: 470.96 MB (4423363 objects)
- Total retained:  18.54 MB (96259 objects)
+ Total retained:  20.03 MB (96049 objects)
```

Looks awesome to me! :tada:

fw42:
Thanks for the profile. Any difference in runtime? Or just memory usage?

ashmaroli:
At a glance, there seems to be gains in overall build times.
However since the repo doesn't have a reliable gauge to measure runtime gains currently, can't say for certain. Either ways, all tests pass so this is a :+1:  move.

Thanks.

PR #1092, opened 2019-04-17
rake memory_profile:run

fw42:
Extracted from https://github.com/Shopify/liquid/pull/1091.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,10 +5,13 @@ end
 
 gemspec
 
-gem 'stackprof', platforms: :mri
-
 group :benchmark, :test do
   gem 'benchmark-ips'
+  gem 'memory_profiler'
+
+  install_if -> { RUBY_PLATFORM !~ /mingw|mswin/ } do
+    gem 'stackprof'
+  end
 end
 
 group :test do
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -5,10 +5,10 @@ end
 
 gemspec
 
-gem 'stackprof', platforms: :mri
-
 group :benchmark, :test do
   gem 'benchmark-ips'
+  gem 'memory_profiler'
+  gem 'stackprof', platforms: :mri
```
ashmaroli: Since you're changing this line, may I suggest using `:install_if`..?
https://bundler.io/v1.10/man/gemfile.5.html#INSTALL_IF-install_if-

**For example,**
```suggestion
install_if -> { RUBY_PLATFORM !~ /mingw|mswin/ } do
  gem 'stackprof'
end
```

ashmaroli:
You *may be able to* DRY this out by `yield`-ing to a helper method:
```ruby
def profile(phase)
  report = MemoryProfiler.report { yield }
  puts
  puts "#{phase}:"
  puts
  print_report(report)
end

profiler = ThemeRunner.new
profile("Parsing")   { profiler.compile }
profile("Rendering") { profiler.render }
```

fw42:
Thanks for the feedback. I updated the PR.

PR #1096, opened 2019-04-23
Run Travis on supported versions of Ruby

davidcornu:
Updates the build matrix to the currently supported versions of Ruby.

https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,11 @@
 language: ruby
 
 rvm:
-  - 2.1
-  - 2.2
-  - 2.3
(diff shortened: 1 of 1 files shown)


davidcornu:
> Hm, should we do this coordinated with a major version bump?

Would make sense. I can let this one sit around until we're ready.

shopmike:
These changes are implemented in https://github.com/Shopify/liquid/pull/1131

PR #1097, opened 2019-04-24
Don't attempt to install stackprof gem on JRuby

ashmaroli:
To allow Travis to proceed testing with `jruby-head`


fw42:
Thank you!

PR #1098, opened 2019-04-24
Rendering with less garbage (2)

fw42:
Same as https://github.com/Shopify/liquid/pull/1091 but with the output buffer being stored on the Liquid context instead of passing it around as a parameter.

@pushrax, which one do you think we should go with? I don't have a strong preference.


pushrax:
I think I prefer the parameter approach. Less method calls, clearer API from the signature. Almost every implementation of `render_to_output_buffer` needs to use the output buffer.

PR #1099, opened 2019-04-29
Use a private constant to stash token-types

ashmaroli:
To avoid allocating a new `[:string, :number]` each time the conditional branch is executed

### DIFF
--- a/lib/liquid/parser.rb
+++ b/lib/liquid/parser.rb
@@ -44,14 +44,14 @@ module Liquid
       tok[0] == type
     end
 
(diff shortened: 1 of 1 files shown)


ashmaroli:
> not sure about the naming of the constant.

I'm open to suggestions. What does `[:string, :number]` represent..?

ashmaroli:
So what is the consensus on this..? Should I commit to naming the constant `SINGLE_TOKEN_EXPRESSION_TYPES` ..?

fw42:
`SINGLE_TOKEN_EXPRESSION_TYPES` works for me

Issue #1087, opened 2019-04-05
abs and round filters


pmyszkier:
I have for example -17.00 price (minus seventeen and zero zero, it is the difference between current variant and default/first variant prices).
I'd like to get 17.00 
When I apply:
{{ -17.00 | abs |round:2 }}
{{ -17.00 | round:2 |abs }}
I'm getting 17.0 not 17.00
How can I get 17.00 ?

PR #1104, opened 2019-05-16
Add round_to_s filter with tests

bimalbhagrath:
Fixes https://github.com/Shopify/liquid/issues/1087

Added `round_to_s` filter to round numbers to strings that allow additional trailing zeros.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -391,6 +391,11 @@ module Liquid
       raise Liquid::FloatDomainError, e.message
     end
 
+    def round_to_s(input, n = 0)
+      result = Utils.to_number(input)
+      sprintf("%.#{n}f", result)
+    end
+
     def ceil(input)
       Utils.to_number(input).ceil.to_i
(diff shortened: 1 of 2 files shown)


fw42:
Hmm this feels very specific to me. Is this really something that many people need?

davidcornu:
> Added `round_to_s` filter to round numbers to strings that allow additional trailing zeros.

🤔 seen as there's already a `| round` filter and formatting with a certain number of decimal places is likely to be an issue for a number of operations involving floats, would we want a number formatting filter instead?

```liquid
{{ 1.2345 | round | num_format: decimal_spaces: 2 }}
```

unixcharles:
I will close for now. I think for this specific use case, the application should provide the filter that are specific to their domain.

For example, Shopify solve this with a [money filter](https://help.shopify.com/en/themes/liquid/filters/money-filters).

If someone want to discuss generic number formatting @davidcornu mentioned, we could open another issue but I'm just not clear on what would be a non business domain specific use case.

PR #1114, opened 2019-07-21
Remove no-op conditional branch

ashmaroli:


fw42:
Hm no super strong opinion but I'm not sure I find this more readable than the previous version (and I don't think it makes any semantic difference, performance or otherwise).

ashmaroli:
Yes, there is no difference whatsoever except for a slight step away from being explicit.
Semantic change would be if I had opted to reduce methods calls:
```ruby
if obj.is_a?(Array)
  output << obj.join
elsif obj             # filter out `nil` and `false`
  output << obj.to_s
end
```

PR #1115, opened 2019-07-22
Reduce string allocations from truncate filters

ashmaroli:
Because `String#+` always generates a new String.
Moreover,  `String#[](range)` and `Array#join`  further generates a new String every time as well.


ashmaroli:
cla shopify com_individual-cla

PR #1117, opened 2019-07-22
Reduce allocations while registering Liquid tags

ashmaroli:
Reduce String allocations by *freezing* `"::"` and Array allocations by using `Array#reject!`

### DIFF
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -50,7 +50,9 @@ module Liquid
       private
 
       def lookup_class(name)
-        name.split("::").reject(&:empty?).reduce(Object) { |scope, const| scope.const_get(const) }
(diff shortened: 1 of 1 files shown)


fw42:
Is this a very hot codepath? If so, we could combine the reject and the reduce into a single loop maybe. Dunno if that's worth it.

ashmaroli:
Its not a hot codepath. Liquid's Tags are registered just once (at the time the file is `require`-d.)
Reducing the iterations here won't have a noticeable effect.

I submitted this PR to reduce memory usage because.. why not? :stuck_out_tongue:

ashmaroli:
On further introspection, why can't we just do **`Object.const_get(name)`** directly?
That'll avoid allocating the array, its contents and the `"::"` string

PR #1120, opened 2019-07-28
Use collections for the sections of the site

ZakCodes:
Instead of adding a `type` to all the pages and filtering the pages by `type`, we can have multiple [collections] for basics, tags and filters.  

Long story short, collections are declared in the `config.yml` and need to be placed in a folder in the root called `_collection` where `collection` is the name of the collection. They can then be accessed using the variable `site.collections.collection` or `site.collections[collection]`. Also, the name of the collection of the current page can be accessed using `page.collection`.

This change doesn't add any functionality, however when I started using [Jekyll] and [Liquid], I looked at the way [the Liquid website] was made to understand how to do certain things. Later on, I found out about collection and changed the way I did it in my repo, so I thought I would change it here so others doing the same thing as me can discover them.

[collections]: https://jekyllrb.com/docs/collections/
[Jekyll]: https://jekyllrb.com/
[Liquid]: https://shopify.github.io/liquid/


ZakCodes:
I signed the CLA, but I don't know how to rerun the check, I tried closing and reopening the merge request, but that didn't seem to work. Is there a way for someone to rerun it, or do I have to open a new merge request?  
I'll reopen a new one tomorrow or the day after if this hasn't been solved.

Issue #1002, opened 2018-04-03
Strange handling of FALSE with include


AlphaHasi:
Inside a template 
```
{% assign var1 = false %}
{% unless var1 %}
  {% assign var1 = 'give var1 a value' %}
{% endunless %}
{% unless var2 %}
  {% assign var2 = 'give var2 a value' %}
{% endunless %}
{{ var1 }} {{ var2 }}
```
The output is `give var1 a value give var2 a value`.

If there is a parent with
```
{% assign var1 = false %}
{% include 'second_template' %}
```
and then inside the second template
```
{% unless var1 %}
  {% assign var1 = 'give var1 a value' %}
{% endunless %}
{% unless var2 %}
  {% assign var2 = 'give var2 a value' %}
{% endunless %}
{{ var1 }} {{ var2 }}
```
The output is also `give var1 a value give var2 a value`.

But if you include like this
```
{% include 'second_template', var1: false %}
```
and then inside the second template again
```
{% unless var1 %}
  {% assign var1 = 'give var1 a value' %}
{% endunless %}
{% unless var2 %}
  {% assign var2 = 'give var2 a value' %}
{% endunless %}
{{ var1 }} {{ var2 }}
```
The output is also `false give var2 a value`.

The odd thing is, that unless is still triggered, but false is not overwritten and just used as plain output. If this is meant as feature it should be like that for all variables, but for example an empty string can get overwritten if a test with if/unless fails.

dylanahsmith:
It looks like what is happening is that the assign (or capture) tags assign to the bottom of liquid's scope stack.  This means those variables are global and not scoped depending on what block tag they appear within.  The include tag attributes are assigned to a new scope that is pushed on the stack, similar to scopes created by the block tags: case, cycle, for, if, unless, ifchanged, table_row.  So those scope variable always get precedence over the variables set with `assign` or `capture`.

I agree that the behaviour is confusing, although changing the behaviour seems likely to break existing templates given that it looks like it has always worked this way.

To avoid this problem, simply avoid re-using the same variable name for include tag attributes as assign/capture variables.

AlphaHasi:
Thanks! 

Yes, I understand the problem, that you can't change a system that is in use so widely. I thought it is a bug, but so it seems to be like that on purpose.

[...]

Issue #911, opened 2017-07-19
Cannot re-assign variable inside snippet


tophattom:
In template:
```liquid
{% include 'my-snippet', test_variable: 'foo' %}
```

my-snippet.liquid
```liquid
{%- assign test_variable = 'bar' -%}
{{ test_variable }}
```

Results in `foo` instead of `bar`.

However, everything works as expected when including the snippet like this:
```liquid
{%- assign test_variable = 'foo' -%}
{% include 'my-snippet' %}
```

danshep:
The assign tag writes to the _last_ scope in the context. Everything else in liquid writes (and reads) from the scopes from first through last.

`
    def render(context)
      val = @from.render(context)
      context.scopes.last[@to] = val
      context.resource_limits.assign_score += assign_score_of(val)
      ''.freeze
    end
`

It feels like a bug, but it's been the behaviour of liquid since forever.

Issue #821, opened 2016-11-10
Possible bug with assign precedence in included partials


jgwmaxwell:
Given a template that includes another template like so:
```
{% include 'item' %}
```

If you then assign a value to the "magic" variable of the template name, a reasonable expectation here would be for the new value to replace the other value.
```
{% assign item = 'foo' %}
{{ item }}
```

However, the assign tag in its render method does:

```ruby
    def render(context)
      val = @from.render(context)
      context.scopes.last[@to] = val
      context.resource_limits.assign_score += assign_score_of(val)
      ''.freeze
    end
```

And the `context.scopes.last` means that it has less variable lookup precedence than the template variable and thus will always, in the case of this example be `nil`. 

It seems to me that it should allocate to `context.scopes.first` as the current active scope? Is this a bug or is there some other logic going on here?

Cheers!

dakotalightning:
👍

PR #1122, opened 2019-07-31
Add new `{% render %}` tag

samdoiron:
Add a new `{% render %}` tag

Example:

```
# the_count.liquid
{{ number }}! Ah ah ah.

# my_template.liquid
{% for number in (1..3) %}
  {% render "the_count", number: number %}
{% endfor %}

Output:
1! Ah ah ah.
2! Ah ah ah.
3! Ah ah ah.
```

The `render` tag is a more strict version of the `include` tag. It is
designed to isolate itself from the parent rendering context both by
creating a new scope (which does not inherit the parent scope) and by
only inheriting "static" registers.

Static registers are those that do not hold mutable state which could
affect rendering. This again helps `render`ed templates remain entirely
separate from their calling context.

Unlike `include`, `render` does not permit specifying the target
template using a variable, only a string literal. For example, this
means that `{% render my_dynamic_template %}` is invalid syntax. This
will make it possible to statically analyze the dependencies between
templates without making Turing angry.

Note that the `environment` of a rendered template is inherited, unlike
the scope, because this environment is immutable from within the
template.

An alternate syntax, which mimics the `{% include ... for %}` tag is
currently in design discussion.


lib/liquid/context.rb
```diff
@@ -12,17 +12,23 @@ module Liquid
...
-    attr_reader :scopes, :errors, :registers, :environments, :resource_limits
+    attr_reader :scopes, :errors, :registers, :environments, :resource_limits, :static_registers
     attr_accessor :exception_renderer, :template_name, :partial, :global_filter, :strict_variables, :strict_filters
 
-    def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_errors = false, resource_limits = nil)
+    def self.build(environments: {}, outer_scope: {}, registers: {}, rethrow_errors: false, resource_limits: nil, static_registers: {})
+      new(environments, outer_scope, registers, rethrow_errors, resource_limits, static_registers)
+    end
+
+    def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_errors = false, resource_limits = nil, static_registers = {})
```
samdoiron: Bit of a difficulty here. Lints aren't allowing this because it has too many arguments. To be fair, it does, but I intentionally avoided splitting up this class to much to keep down on allocations.

Thoughts?

dylanahsmith:
It always seemed to me that snippets were being used for similar reasons as function calls and sharing variables across these "function calls" makes the code hard to understand.  This seems like it essentially avoids sharing those variables across functions, which seems like a step in the right direction!

Should we make this more like a function call by also having a way to declare parameters at the top of these snippets?  That way it will be clearer how to use a snippet by just looking at the top of its file, would make it easy to statically detect invalid use of the snippet and help catch conflicts for themes being worked on by multiple authors.

Thibaut:
Some context about why we're introducing this for those outside Shopify: over the years we've found many problems and limitations with `include`:
-  `include` introduces dependencies between files, which makes solving certain problems such as automated upgrades hard or impossible, and is the cause of much complexity in our themes;
- dynamic includes (`{% include variable %}`) prevent certain performance optimizations;
- there are number of annoying issues with `include` (https://github.com/Shopify/liquid/issues/911 https://github.com/Shopify/liquid/issues/821 https://github.com/Shopify/liquid/issues/1002 https://github.com/Shopify/liquid/issues/293) which we're unable to fix due to our promise for backward compatibility.

As part of the new online store design experience that was [announced](https://www.shopify.ca/partners/blog/shopify-unite-announcements-2019) at Unite, we're introducing three new type of section files in our themes, where we're planning to remove support for `include`. We intend to deprecate `include` in favor of `render` going forward.

lib/liquid/tags/render.rb
```diff
@@ -0,0 +1,79 @@
...
+      @attributes = {}
+      markup.scan(TagAttributes) do |key, value|
+        @attributes[key] = Expression.parse(value)
+      end
+    end
+
+    def parse(_tokens); end
+
+    def render_to_output_buffer(context, output)
+      template_name = context.evaluate(@template_name_expr)
```
Thibaut: Don't we want to disallow that? The template name should always be static.

lib/liquid/tags/render.rb
```diff
@@ -0,0 +1,79 @@
...
+        parse_context.partial = false
+      end
+      cached_partials[template_name] = partial
+      context.registers[:cached_partials] = cached_partials
+      partial
+    end
+
+    def read_template_from_file_system(context)
+      file_system = context.registers[:file_system] || Liquid::Template.file_system
+      file_system.read_template_file(context.evaluate(@template_name_expr))
```
Thibaut: `context.evaluate(@template_name_expr)` is also done in `render_to_output_buffer`. Can we de-dup?

lib/liquid/tags/render.rb
samdoiron: Dynamic template names are prevented at parser time. This simply unescapes the string (eg `"\"foo\""` into `"foo"`).

Thibaut: Ah I see. There may be more performant way to extract the name than calling `context.evaluate` then.

lib/liquid/context.rb
```diff
@@ -89,7 +95,7 @@ def invoke(method, *args)
     # Push new local scope on the stack. use <tt>Context#stack</tt> instead
     def push(new_scope = {})
       @scopes.unshift(new_scope)
-      raise StackLevelError, "Nesting too deep".freeze if @scopes.length > Block::MAX_DEPTH
+      check_overflow
```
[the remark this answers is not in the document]
samdoiron: I've tried to keep the changes separated by-commit for review, but it may be a good idea to merge the scope changes independently :+1:.

shopmike:
This is such a great move. Can i also add instead of just

shopmike:
This is such a great step in the right direction. Can I just add one comment about this

> dynamic includes ({% include variable %}) prevent certain performance optimizations;

I wrote multiple iterations of Carted and one of the versions I made removed dynamic includes and it could render 1 million times in 0.6 secs vs 6 secs for the next best version I had. 

The point I wanted to add was to also gain the optimisation I also had to be able to detect and prevent uploading of templates that caused recursive includes. Think something like include 'product' > include 'quickview' > include 'product' as until you resolve the logic you can't tell if it will just infinitely loop and as such can't work out reliably what is going to happen.

shopmike:
Reading through this again this is a little different to my implementation as you are just limiting scope altogether. The approach I took was that on save the includes were resolved and pretty much made part of the template as if they were always there. Then for any update to an include it would trigger a rebuild of the scripts the include it. For me this still allows users to mess around with the global scope but also gives the optimiser complete visibility of the final program so that you can do type inference, resolve variables etc during a compile stage.

samdoiron:
@shopmike This idea is definitely on my radar! We have a similar feature in our theme toolkit (Slate) which is either in the works or out now that bakes `include`s into the template directly. The ability to do this in all cases is exactly what we're hoping to make possible by disallowing dynamic `render`s.

samdoiron:
Also: the recursive includes point is an interesting one. I believe this was one of the limitations (that we checked) for the Slate implementation.

Since we already limit the degree of recursion in includes (which unfortunately also makes Liquid turing-incomplete 😢) we could in theory include recursive templates until the nesting limit, but this would likely result in huge output files.

In theory we could see similar low render overheads through careful use of fragment caching, which `render` will allow us. There are, unfortunately, legitimate uses for recursive templates (eg. comment threads).

shopmike:
If you limit the scope I guess the recursion doesn't matter any more, the only reason I needed to resolve was to find places the scope was modified. 

I also agree with what @dylanahsmith said about these being used like functions, maybe its worth considering making this more function like, I feel like developers especially people just starting out would then understand why this is different and how it behaves.

With that in mind I'd likely consider a nice output method as right now, building a store if I wanted to bring scope out of this thing my mind instantly makes me consider using the below to make a sort of function return

{% capture 'variable' %}{% render 'something` %}{% endcapture %}

samdoiron:
The plan at this point would be to soft-deprecate `include` and recommend `render` for all future use cases.

The idea of making `render` more powerful / function-like has definitely been considered (with me as an advocate). Eventually we decided to remain as consistent with `include` as possible to ease the upgrade path. That said, adding function-like behaviours in the future is always a possibility (ideally in a way that extends this approach).

lib/liquid/context.rb
```diff
@@ -126,6 +132,24 @@ def stack(new_scope = nil)
       @this_stack_used = old_stack_used
     end
 
+    # Creates a new context inheriting resource limits, filters, environment etc.,
+    # but with an isolated scope.
+    def new_isolated_subcontext
```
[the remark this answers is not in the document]
pushrax: `fork` to me implies that all the scopes are copied, whereas this method produces no scopes.

lib/liquid/tags/render.rb
pushrax: The unquoting is done by `Expression.parse`, which will return a String instance. `context.evaluate` is a no-op here. I'd probably keep it anyway though, in case this tag were extended. Its total cost is a `respond_to?`.

I'd probably add a comment explaining that the template name parameter is enforced to be a string due to the tag syntax. It also wasn't clear to me at first.

lib/liquid/tags/render.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      output
+    end
+
+    private
+
+    alias parse_context options
+    private :parse_context
+
+    def load_cached_partial(template_name, context)
```
pushrax: Can we abstract this into something to share with Liquid::Include?

test/integration/tags/render_tag_test.rb
```diff
@@ -0,0 +1,153 @@
...
+    Liquid::Template.file_system = StubFileSystem.new('snippet' => '{{ price }}')
+    assert_template_result '123', '{% render "snippet", price: 123 %}'
+  end
+
+  def test_render_accepts_multiple_named_arguments
+    Liquid::Template.file_system = StubFileSystem.new('snippet' => '{{ one }} {{ two }}')
+    assert_template_result '1 2', '{% render "snippet", one: 1, two: 2 %}'
+  end
+
+  def test_render_does_inherit_parent_scope_variables
```
pushrax:
```suggestion
  def test_render_does_not_inherit_parent_scope_variables
```

pushrax:
> Should we make this more like a function call by also having a way to declare parameters at the top of these snippets?

Definitely seems like it'd improve maintainability of library code. It's currently very easy to forget to pass or typo a parameter, and the output might be visually indistinguishable even if tested manually.

If we plan to have this, it would be good to coordinate it with the release of the render tag. Something like:

```
{% arguments foo, bar, baz = 3, a:, b:, c: 42 %}
```

or some subset/renaming of that would make sense to me (as a Ruby dev). The tag could be optional, to allow migration from `{% include %}` with minimal work.

An alternative is to create function declaration/call tags that can be used independently of snippets, though giving more ways to do the same thing adds confusion.

I don't think this needs to block `{% render %}` from being merged, but we should decide before any major announcement.

> `{% capture 'string_upper' %}{% render 'to_upper' input: string %}{% endcapture %}`

This could be optimized to be equivalent to a "call a function and assign to a variable" tag, so I think the main question here is what kind of code we want to encourage, and which new mental models we want to require people learn.

lib/liquid/context.rb
pushrax: 👍 let's just silence this one.

shopmike:
So just some thinking on this, what is the motivation for isolating the scope completely? a simple approach to this could possibly just be changing the following line in assign and capture

`context.scopes.last[@to] = val`

to a new variable on the context called `rootScope`

`context.rootScope[@to] = val`

and then all the render tag needs to do is extend assign and in addition change 

`context.roootScope = ||the current assign scope||`

then change it back when it's done. So instead of isolating the scope is limiting mutations to only happen up to the scope of the current block.

What had me thinking about this is that anyone porting `assign` tags across all make the assumption that objects like `product` or `collection` are already in scope ready to use.

samdoiron:
The product / collection case will still work with this method since those are part of the `environment`, which is inherited, and not the calling scope. I suppose this does mean that we will have to consider large context for caching purposes, but hopefully that would lead to more cache hits than just the entire variable set.

So if I understand, your idea would be to prevent writes to the old scope, but allow reads? My worry there is that is still a bit of a foot gun, since you could completely ignore parameters in that case.

pushrax:
> So instead of isolating the scope it is just limiting mutations to only happen up to the scope of the current block.

Creating a strict definition of the data available to the snippet is one of the main motivations I think, to allow for improved caching and concurrent rendering without needing to statically analyze the template (which might use dynamic property access and defeat basic static analysis).

samdoiron:
Good catch @Thibaut on the `increment` etc. tests. For some reason, `increment` stores its value in the _environment_, rather than the current scope.

This almost feels like a bug to me, since I can't think of any reason why you'd want this. However, changing this would be a breaking change to Liquid. Not removing it, however, would mean that it isn't safe to simply inherit the environment for the `render` tag.

Specifically, changing this behaviour would change the output of templates like this:
```liquid
{% assign my_counter = 1 %}
{% increment my_counter %}
{{ my_counter }}
```

In the current version, this will output `01`. If we instead store increments in the scope, it would output `12`.

samdoiron:
As I see it, we have a few options to address this.

1. Do not inherit the environment in `render`ed snippets. This has wins for caching, but would mean you would need to pass the those parameters down explicitly rather than relying on them being ambient. For example, you would have to pass `product: product` on a product page in Shopify. In other words: this would make it more difficult to upgrade to `render`, since you are less likely to be able to find / replace it with `include`.

2. Create a special case for `increment` / `decrement`, and clear those between `render`s but keep the rest of the environment intact. This is a band-aid, though, and would not handle cases where others have defined their own tags which modify the environment.

3. Keep this behaviour, and lose the isolation that `render` gives us from the environment.

Thibaut:
I think we have no choice but to do (2), which is what we do to isolate the scope of [sections](https://help.shopify.com/en/themes/development/sections) in Shopify. The way `increment` works is unfortunate but we can't break backward compatibility and the drawbacks of (1) and (3) are too significant.

pushrax:
Agreed on `I think we have no choice but to do (2)`. To implement that, perhaps we can make a new abstraction, of un-assignable scoped values?

samdoiron:
Similar to `static_registers` then? That might be the answer -- ideally we also wouldn't break liquid gem compatibility, which that would do.

My current thoughts are around recording which variable are used in counters (in a separate register) and then cleansing them before rendering.

shopmike:
I'm going to go against the grain here and say #1. Its cleaner and will open way more options with caching etc as everything is passed in through the same method.

Big reason I say this is that it makes more sense to either have all scope or no scope from actually using the implementation. When I talk about `product` being in scope its not actually the environment version that your usually working with its more likely the case of 

```
{% for product in collection.products %}
  ....
  {% render 'collection-product' %}
  ....
{% endfor %}
```

Where in the above example if you only pass across environment you end up needing to think about which context you're using a render as it will change. I think for this a common example is just the logic you have to show product pricing. You just kind of always assume that a `product` variable is available.

**template/product.liquid** ( Which is fine as product is environment )
```
<div class="price">{% render "product-price" %}</div>
```

However if I want to reuse in the **collection-product.liquid** example above I need to use 
```
<div class="price">{% render "product-price" product: product %}</div>
```

So for me I'd end up defaulting to always defining the product anyway as you would get sick of copy and pasting parts of templates and depending where you do so having to add the `product: product` part anyway

shopmike:
Just another common use case that this will create some weird behaviour on is quick view on collection pages. So you would end up with

**template/collection.liquid** 
```
{% for product in collection.products %}
  ....
  {% render 'collection-product' %}
  ....
{% endfor %}
```

Then inside that you would have
**snippets/collection-product.liquid** 
```
<div class="modal">
   {% render 'product' product:product %}
</div>
```

Which is fine as your defining `product` however every usage in the product template can either assume it knows `product` or not so you would have a mixture of

Without defined
`{% render 'product-price' %}`

With defined
`{% render 'product-upsells' product:product %}`

However the ones that didn't define product will not work as product is no longer in scope as it's just making the assumption you're on the product page but in this case you're not, you're using the product template on a collection page.

shopmike:
Also I wanted to ask whats going to stop someone from just using a `render` tag to include a snippet that is just full of `include` anyway.

Thibaut:
> Also I wanted to ask whats going to stop someone from just using a `render` tag to include a snippet that is just full of `include` anyway.

Good catch. We should prevent that, same as we block `include` in the new section files.

> I'm going to go against the grain here and say 1 or obviously 3.

I think (3) is a non-starter because it removes the main reason behind wanting to add `render`, which in itself comes at the cost of fragmenting the platform / making the Liquid API more complex (which we're very wary of doing since we never want to break backward compatibility, but in this case I think the benefits outweigh the cost). Isolating the scope will lead to cleaner theme code, unlock performance improvements (e.g., parallel rendering and possibly better caching), and is also the basis of an upcoming platform feature where scope isolation is a requirement (app snippets).

(1) will make adoption significantly harder, which Tobi has argued against. If we were starting Liquid from scratch that might be my preferred option but we unfortunately can't ignore the reality of what we have now. It also doesn't play well with the platform feature I mentioned (app sections) because it would require the `{% render %}` code to be modified whenever the snippet wants to access a new global variable, while we want apps to be able to iterate on their Liquid code without modifying theme code.

samdoiron:
> If we were starting Liquid from scratch that might be my preferred option but we unfortunately can't ignore the reality of what we have now. 

:+1: on this. Given the realities of adoption, I'm leaning towards (2) now as well. `increment`/`decrement` are hacks, and almost certainly shouldn't have been implemented how they have been. That said, given they exist, I think painting around them with a special case might be the pragmatic thing.

I mentioned this before in passing, but IMO not only can we not put them in the scope, we probably shouldn't put them in a new specialized container. These filters are part of the "standard library" of liquid, and moving them around would maintain template API compatibility but break our Ruby API. My plan should we go forward with (2) is so

1. Separately record which variables in the topmost environment are `increment` related
2. Clear those variables before creating the `render` subscope.

The only potential issue there would be if other Liquid APIs, following `increment`/`decrement`'s example, have chosen to use `environment` for similar purposes. A solution to that could be to continue the `static_registers` trend this PR introduced and add a similar `static_environment` that would be passed to `render`, while all of `environment` would not.

lib/liquid/partial_cache.rb
```diff
@@ -0,0 +1,41 @@
+module Liquid
+  class PartialCache
+    def self.for(*args)
```
samdoiron: Might seem useless, but I think this helps imply that caches are per-context. Eg. `PartialCache.for(context...)`

samdoiron:
I've implemented option (2) for now, though we could still change this pending discussion.

lib/liquid/partial_cache.rb
pushrax: Rather than a class, what do you think about a singleton `#load`, like

```
def self.load(template_name, context:, parse_context:)
```

Saves a few allocs, and I don't really see much value in the class instance here.

lib/liquid/context.rb
```diff
@@ -203,8 +229,39 @@ def lookup_and_evaluate(obj, key, raise_on_not_found: true)
...
+    # variables in the current scope.
+    #
+    # Since environments are shared between isoated scopes, this would lead
+    # to a state leak and prevent us from caching template fragments like those
+    # used by `{% render %}`. We still want to inherit the rest of the environment,
+    # so we track which values come from counters and reset those when creating
+    # a subscope.
+    def environments_with_reset_counters
+      cleansed_counter_environment = environments.first.clone
+      registers.fetch(:counter_names) { Set.new }.each do |counter_name|
```
pushrax: Is this equivalent to `registers[:counter_names]&.each do |counter_name|`?

lib/liquid/context.rb
```diff
@@ -203,8 +229,39 @@ def lookup_and_evaluate(obj, key, raise_on_not_found: true)
...
+
     private
 
+    attr_reader :base_scope_depth
+
+    # As a legacy quirk, counters (the values modified by the `increment` and
+    # `decrement` tags) are stored in the highest environment, not as
+    # variables in the current scope.
+    #
+    # Since environments are shared between isoated scopes, this would lead
```
pushrax: isolated

shopmike:
Yeah I can't argue with that logic, when you lay out the full roadmap then #2 is the clear winner, especially in terms of apps being able to have generic code that can't be broken by others or impact existing code. So ultimately these are isolated components that have access to the root context which is generally everything you need and much like React you have the ability to pass down props if you really need context from a parent. 

I think I read these were planning to go in the snippets folder but I'm assuming because you need to limit them from using `include` and `section` that these will in fact be in their own?

Thibaut:
> So ultimately these are isolated components that have access to the root context which is generally everything you need and much like React you have the ability to pass down props if you really need context from a parent.

Exactly! :)

> I think I read these were planning to go in the snippets folder but I'm assuming because you need to limit them from using `include` and `section` that these will in fact be in their own?

Yes we're planning to reuse `snippets/` to make adoption easier (no need to duplicate files) and to avoid adding yet-another-directory to themes. The restriction on using `{% include %}` and `{% section %}` should be enforced at render-time. If we're rendering a `{% render %}`, using those tags should return an error.

lib/liquid/context.rb
samdoiron: Yep! It would be. Not sure that lonely operator is supported in all our Ruby versions, though.

lib/liquid/context.rb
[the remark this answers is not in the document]
Thibaut: Ah good point. So maybe just `build_subcontext`.

samdoiron:
> TODO: block {% include %} when rendering {% render %}
Was this something we wanted to do at this level? I imagined this would be a Shopify constraint.

samdoiron:
Also: rewrote the `increment` / `decrement` handling as per @pushrax's suggestion. We now have separate `environments` and `static_environments` fields on `Context`, the later frozen. Only `static_environments` is inherited by isolated subscopes.

Thibaut:
> Was this something we wanted to do at this level? I imagined this would be a Shopify constraint.

We'll need to block `{% section %}` in Shopify but IMHO it would be good to block `{% include %}` in Liquid, as `{% render %}` is intended to be a replacement/alternative for it.

shopmike:
I think because it's impossible to ever deprecate `include` from Liquid, however it's possible for Shopify to stop supporting it, that this should be done within Shopify as it's more a choice on implementation that a limitation of what liquid can do.

Thibaut:
:+1: happy to do it in Shopify then

samdoiron:
To be clear, this would be blocking it within the context of `render`, not entirely. Still :+1: either way, though.

shopmike:
Yeah I did get that, it was more the fact it's still highly useful to be able to have includes inside renders. For something like Jekyll that isn't performance based as its statically built i see this as you still have a way to include snippets that do change context inside new isolated contexts provided by render.

shopmike:
I think you'll also find include will also make some sort of comeback 😊. I truely think the render method is what apps should be doing for the majority. However there is a whole breed of apps that don't fit into this model. Check out apps like Locksmith or any of the apps that do Price Breaks etc. They are built on the fact that scope is allowed outside.

lib/liquid/context.rb
```diff
@@ -203,8 +221,32 @@ def lookup_and_evaluate(obj, key, raise_on_not_found: true)
...
     end
 
+    protected
+
+    attr_writer :base_scope_depth, :warnings, :errors
+
     private
 
+    def find_environment(environments, key, raise_on_not_found:)
+      environments.each do |e|
```
Thibaut: Nit: `environments.each do |environment|`

lib/liquid/context.rb
```diff
@@ -168,17 +195,8 @@ def find_variable(key, raise_on_not_found: true)
...
-          variable = lookup_and_evaluate(e, key, raise_on_not_found: raise_on_not_found)
-          # When lookup returned a value OR there is no value but the lookup also did not raise
-          # then it is the value we are looking for.
-          if !variable.nil? || @strict_variables && raise_on_not_found
-            scope = e
-            break
-          end
-        end
-      end
+      scope ||= find_environment(@environments, key, raise_on_not_found: raise_on_not_found)
```
Thibaut: It seems we're doing the `lookup_and_evaluate` in `find_environment` but then dismissing the value, and doing it again on L202 below. Could we optimize this?

lib/liquid/partial_cache.rb
```diff
@@ -0,0 +1,18 @@
...
+  class PartialCache
+    def self.load(template_name, context:, parse_context:)
+      cached_partials = (context.registers[:cached_partials] ||= {})
+      cached = cached_partials[template_name]
+      return cached if cached
+
+      file_system = (context.registers[:file_system] ||= Liquid::Template.file_system)
+      source = file_system.read_template_file(template_name)
+      parse_context.partial = true
+      Liquid::Template.parse(source, parse_context).tap do |partial|
```
Thibaut: Nit: since this is a hot code path I would avoid the use of `tap`

lib/liquid/tags/render.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      @template_name_expr = Expression.parse(template_name)
+
+      @attributes = {}
+      markup.scan(TagAttributes) do |key, value|
+        @attributes[key] = Expression.parse(value)
+      end
+    end
+
+    def parse(_tokens); end
```
Thibaut: This is the same implementation as the parent class so I think we can :fire:

lib/liquid/partial_cache.rb
samdoiron: This actually ended up even cleaner since assignment returns the value anyway 🙌

lib/liquid/context.rb
```diff
@@ -161,27 +190,25 @@ def evaluate(object)
 
     # Fetches an object starting at the local scope and then moving up the hierachy
     def find_variable(key, raise_on_not_found: true)
+      variable = UNDEFINED
```
samdoiron: Big rewrite of `find_variable` here. The original was... very weird. After verifying that neither `environment` nor `scope` had the variable, it then still looked for them in both -- which would always return `nil`. Now we just fall back to `nil` explicitly.

That is, assuming there's not some behaviour I've missed here. Would appreciate extra :eyes: on this change.

shopmike: I feel like this can be simplified by pulling out these 3 `variable == UNDEFINED` and instead being inside try_find_variable and just return if a match is found or nil if it gets to the end.

samdoiron: Problem is that `nil` is a valid value for a defined value :/. That's why I've pulled out `UNDEFINED` separately.

Previously we relied on the `scope` found being `nil` or not to differentiate between a found `nil` and a transient `nil`, but this refactoring doesn't have that.

shopmike: so what I kinda of mean is
```
if index
....
else 
   variable = try_find_variable(raise_on_not_found: raise_on_not_found)
endif
```

```
def try_find_variable( key, raise_on_not_found:) 
   [@environments, @static_environments].each do |environments|
      ...
   end
   nil
end

Comment (samdoiron) [reply lib/liquid/context.rb]: Ah ok, so checking them each there. And since we return `nil` as a default it doesn't matter if we can tell whether it was found or not 👍. I'll try that out.

Comment (shopmike) [reply lib/liquid/context.rb]: this would avoid needing to use `UNDEFINED` at all

Comment (samdoiron) [reply lib/liquid/context.rb]: That does cost us an allocation technically, though (for the `[@environments, ...]`. Not sure if that's worth worrying about, but this is basically the hottest path in Liquid.

Comment (shopmike) [reply lib/liquid/context.rb]: if you don't want the allocation I guess you could duplicate the code one after another even though its not pretty

Comment (shopmike) [reply lib/liquid/context.rb]: Or would ruby support moving this up to the class level

@all_environments = [@environments, @static_environments]

Comment (samdoiron) [reply lib/liquid/context.rb]: :+1: That would also work

**Comment by samdoiron** [comment]
`find_variable` refactored again 👍

File: lib/liquid/context.rb [hunk 16]
```diff
@@ -35,7 +44,9 @@ def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_erro
       @interrupts = []
       @filters = []
       @global_filter = nil
+
```
Thibaut: 🔥

lib/liquid/context.rb
```diff
@@ -164,25 +193,13 @@ def find_variable(key, raise_on_not_found: true)
...
-          if !variable.nil? || @strict_variables && raise_on_not_found
-            scope = e
-            break
-          end
-        end
+      variable = if index
+        lookup_and_evaluate(@scopes[index], key, raise_on_not_found: raise_on_not_found)
+      else
+        try_find_in_environments(key, raise_on_not_found: raise_on_not_found)
       end
```
Thibaut: Much better!

lib/liquid/context.rb
```diff
@@ -203,8 +220,32 @@ def lookup_and_evaluate(obj, key, raise_on_not_found: true)
...
 
+    protected
+
+    attr_writer :base_scope_depth, :warnings, :errors
+
     private
 
+    attr_reader :base_scope_depth
+
+    def try_find_in_environments(key, raise_on_not_found:)
```
Thibaut: Nit: `try_find_variable_in_environments`

lib/liquid/context.rb
```diff
@@ -164,25 +193,13 @@ def find_variable(key, raise_on_not_found: true)
...
-            break
-          end
-        end
+      variable = if index
+        lookup_and_evaluate(@scopes[index], key, raise_on_not_found: raise_on_not_found)
+      else
+        try_find_in_environments(key, raise_on_not_found: raise_on_not_found)
       end
 
-      scope ||= @environments.last || @scopes.last
```
Thibaut: I'm struggling to understand what this line (which is not in the new implementation) was doing, since at that point we already searched in both `@environments` and `@scopes` and so I'd expect that whatever scope this assigns, the variable is not in there.

samdoiron: Likewise. It seems like it should always return `nil`.

My best guess is that this is copy / pasted from variable _setting_ code, where we would need to know the scope in order to set the value in the right place.

samdoiron:
@pushrax @dylanahsmith Ready for :eyes:

lib/liquid/context.rb
```diff
@@ -126,6 +135,24 @@ def stack(new_scope = nil)
...
+      check_overflow
+
+      Context.build(
+        resource_limits: resource_limits,
+        static_environments: static_environments,
+        static_registers: static_registers
+      ).tap do |subcontext|
+        subcontext.base_scope_depth = base_scope_depth + 1
+        subcontext.exception_renderer = exception_renderer
+        subcontext.add_filters(@filters)
```
shopmike: This is only a suggestion and very weakly held. Performance wise does it help to make this `subcontext.filters = @filters`. Just thinking of the case when this is inside `for` blocks, any gain when building context will help the overall speed.

lib/liquid/tags/render.rb
```diff
@@ -0,0 +1,54 @@
...
+      template_name = context.evaluate(@template_name_expr)
+      raise ArgumentError.new(options[:locale].t("errors.argument.include")) unless template_name
+
+      partial = PartialCache.load(
+        template_name,
+        context: context,
+        parse_context: parse_context
+      )
+
+      inner_context = context.new_isolated_subcontext
```
shopmike: @fw42 you asked what were some of the performance gains I did for carted. This was a big part I didn't mention either. So carted parsed templates on upload because it kept templates in memory. So a big thing I would do in this instance is shift everything out of render that allocates up in to the parse method that way in render everything was available to use. You could maybe consider this for frequently used templates in Shopify.

shopmike: Actually this isn't a good example as its specific to this request. In this case this is where a pool would come in. I'd have a pool of empty subcontexts already created. Then all this would do is grab one of the available and if there were none it would create one. Then at the end of this block there would be a method to reset the subcontext back to its empty state for reuse by the next request.

lib/liquid/context.rb
samdoiron: :+1: That would be slightly faster. Will update.

test/unit/context_unit_test.rb
```diff
@@ -468,11 +468,79 @@ def test_apply_global_filter
...
 
+  def test_static_environments_are_read_with_lower_priority_than_environments
+    context = Context.build(
+      static_environments: { 'shadowed' => 'static', 'unshadowed' => 'static' },
+      environments: { 'shadowed' => 'dynamic' }
+    )
+
+    assert_equal 'dynamic', context['shadowed']
+    assert_equal 'static', context['unshadowed']
+  end
```
pushrax: I'm not sure why the Liquid-C tests didn't fail here. Liquid-C implements Context#find_variable, and doesn't handle static_environments. This is critical to look into before shipping this.

shopmike: This isn't failing as its looking at your https://github.com/Shopify/liquid-c/pull/43 which I think was forked before the new `Context#find_variable` changes.

shopmike: This is the reason I do like the idea of https://github.com/Shopify/liquid/issues/1134

asbjornu:
Can you please create a new release that includes this sweet new `render` tag?

PR #1124, opened 2019-08-08
New Liquid format method

shopmike:
# Liquid Format

`Liquid::Template.parse(template).format`

This is a work in progress but I wanted to keep visibility and start discussions around it. I'm going to spend some of my spare time on this as I believe its a strong way to move Liquid forward. There is still a fair bit to do but shouldn't take more than a couple of weeks given that I'm not spending much time on this.

## Why is this needed

1. **Removal of the :lax and :warn parsers** -  I would like to think a v5 or v6 of Liquid will deprecate both these parsing modes. The main idea behind the format command is it provides a way to parse with the :lax or :warn parsers and then output as :strict compatible syntax. So a possible way to move from v4 to v5 is simply running the format command across previous templates. An online conversion tool could also be provided for end users by liquid platforms so they convert templates outside of the system.

2. **Provides an upgrade path** - There have been requests over time for specific features that make sense in Liquid however they are impossible to implement without impacting previous templates. A simple example of this is brackets within an if statement, people currently have these in their template but they currently don't do anything, introduction of this will likely change the rendering output and the `format` tool can be a way to provide the upgrade tools to make breaking changes.

3. **Provides systems with consistent formatting** - One of the best features of Go is the inbuilt `go fmt` command. For me this shows a great example that by having an in-built format command systems can then leverage it to provide a better experience. For example a theme editor can run the format command on save and make the whitespace around tags and filters uniform so that future developers have a consistent experience.

4. **Provides a way to debug** - By running code through the format command, you actually get a chance to see how Liquid is processing your template. Its rather common to realise that for some reason Liquid is actually just ignoring most of your code.

## What's left to do

**White Space**
- [ ] Retain whitespace in an IgnoredWhitespace token so format can recover.
- [ ] Determine if a tag has white space controls

**Expressions**
- [ ] Detect when a string should be a single quote or double quote
- [ ] Handle integers
- [ ] Handle ranges 
- [ ] Handle floats
- [ ] Handle literals

**Tags**
- [ ] Assign
- [ ] Break
- [ ] Capture
- [ ] Case
- [ ] Comment
- [ ] Continue
- [ ] Cycle
- [ ] Decrement
- [ ] For
- [ ] If
- [ ] Ifchanged
- [ ] Include
- [ ] Increment
- [ ] Raw
- [ ] Table_row
- [ ] Unless

## Discussion topics

1. I decided to put these in the core of the system, I feel like this method can be forgotten if it was just a side project and its compatibility was lost. Also by putting the methods directly on tags it makes it easy to add this support to existing tags. However I would like to know if anyone thinks there is a better approach

2. The whitespace control I'm going to modify to output a tag that has an empty render method like comments but stores what that whitespace originally looked like. Any feedback or comments on this will be helpful?

@pushrax @ashmaroli @fw42 @samdoiron @dylanahsmith @Thibaut @tobi

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -70,6 +70,21 @@ module Liquid
       render_to_output_buffer(context, '')
     end
 
+    def format(output)
+      idx = 0
+      while node = @nodelist[idx]
+        output << case node
+        when String
+          node
+        else
+          node.format
+        end
+        idx += 1
+      end
+
+      output
+    end
+
     def render_to_output_buffer(context, output)
       context.resource_limits.render_score += @nodelist.length
 
--- a/lib/liquid/expression.rb
+++ b/lib/liquid/expression.rb
@@ -45,5 +45,16 @@ module Liquid
         end
       end
     end
+
+    def self.format(expression)
+      case expression
+      when String
+        "'#{expression}'"
+      when VariableLookup
+        expression.format
+      else
+        expression.to_s
+      end
+    end
   end
 end
--- a/lib/liquid/variable.rb
+++ b/lib/liquid/variable.rb
@@ -35,6 +35,44 @@ module Liquid
       @markup
     end
 
+    def format
+      output = Expression.format(name)
+
+      filters.each_index do |i|
+        first_filter = true
+        output << " | #{filters[i][0]}"
+        first_sub_filter = true
+        filters[i][1].each do |j|
+          if first_filter
+            output << ':'
+            first_filter = false
+          end
+          unless first_sub_filter
+            output << ','
+          end
+          output << " "
+          output << Expression.format(j)
+          first_sub_filter = false
+        end
+        filters[i][2].each do |j, k|
+          if first_filter
+            output << ':'
+            first_filter = false
+          end
+          unless first_sub_filter
+            output << ','
+          end
+          output << " "
+          output << j
+          output << ": "
+          output << Expression.format(k)
+          first_sub_filter = false
+        end
+
+      end
+      "{{ #{output} }}"
+    end
+
     def markup_context(markup)
       "in \"{{#{markup}}}\""
     end
--- a/lib/liquid/variable_lookup.rb
+++ b/lib/liquid/variable_lookup.rb
@@ -31,6 +31,21 @@ module Liquid
       end
     end
 
+    def format
+      output = name.to_s
+      @lookups.each_index do |i|
+        output << case lookups[i]
+                 when String
+                   ".#{lookups[i]}"
+                 when VariableLookup
+                   "[#{lookups[i].format}]"
+                 else
+                   ".#{lookups[i]}"
+                 end
+      end
+      output
+    end
+
     def evaluate(context)
       name = context.evaluate(@name)
       object = context.find_variable(name)
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.swp
 pkg
 *.rbc
+.idea
 .rvmrc
 .ruby-version
 Gemfile.lock
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -134,6 +134,11 @@ module Liquid
       self
     end
 
+    def format
+      return ''.freeze if @root.nil?
+      @root.format('')
+    end
+
     def registers
       @registers ||= {}
     end
--- a/test/integration/format_test.rb
+++ b/test/integration/format_test.rb
@@ -1,0 +1,23 @@
+require 'test_helper'
+
+class FormatTest < Minitest::Test
+  include Liquid
+
+  def test_string_only
+    assert_template_format('foo', 'foo')
+    assert_template_format('foo bar', 'foo bar')
+  end
+
+  def test_variable_only
+    assert_template_format('{{ foo }}', '{{foo}}')
+    assert_template_format('{{ foo.bar }}', '{{foo.bar}}')
+    assert_template_format('{{ foo.hello }}', '{{ foo["hello"] }}')
+    assert_template_format('{{ foo[hello] }}', '{{ foo[hello] }}')
+    assert_template_format('{{ foo.bar.world }}', '{{ foo["bar"].world }}')
+    assert_template_format('{{ foo }}{{ bar }}', '{{foo}}{{bar}}')
+    assert_template_format(%({{ input | substitute: first_name: surname, last_name: 'doe' }}), %({{ input | substitute: first_name: surname, last_name: 'doe' }}))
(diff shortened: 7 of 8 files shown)


tobi:
Very cool.

fw42:
The general idea of rewriting bad templates into good templates makes a lot of sense to me. I remember there might have been a prototype (hackdays?) for this already a few years ago (@pushrax?) but it never got shipped.

Do we care about the scenario of someone having an old template stored offline somewhere (maybe a theme they bought and downloaded to their own computer) and in five years from now they come back and expect that theme to still work?

tobi:
Yea offline backwards compatibility has to be ensured. But at the very very
least we can integrate this as a format button into the online editor and
the CLI tools. We should probably format the html around the liquid as well
in those cases though.

- tobi
CEO Shopify

shopmike:
This is where I think an online tool would be beneficial that goes along with this, just thinking a really basic heroku type setup that just runs ruby and you can upload multiple liquid files and get an output. I think if you upgrade a system, the error message could possibly point to this location suggesting you may be using an outdated syntax.

tobi:
That would be too byzantine for merchants who are often non-technical and
treat these theme zip files as found artifacts of unknown power and
functionality.

- tobi
CEO Shopify

evulse:
Actual zip files of themes I think is an easier one as if it fails to load you can pretty much guarantee it was valid code before, so you can just host a service internally that ships the zip off when it fails and converts it and then loads it in. For me working with bigger merchants the guys that have to do to the most work are the ones that run it through version control. They will need to upload all their files to a service, pull them down add them back in and commit.

lib/liquid/expression.rb
```diff
@@ -45,5 +45,16 @@ def self.parse(markup)
         end
       end
     end
+
+    def self.format(expression)
+      case expression
+      when String
+        "'#{expression}'"
```
ashmaroli: This could lead to broken renders if the `expression` itself contains single quotes. For example:
```
{{input | substitute: '<subtext>', "two's company"}}
```
would get *formatted* into:
```
{{ input | substitute: '<subtext>', 'two's company' }}
```

lib/liquid/variable.rb
```diff
@@ -35,6 +35,52 @@ def raw
       @markup
     end
 
+    def format
+      output = Expression.format(name)
+
+      filters.each_index do |i|
+        output << " | "
```
ashmaroli: The substrings *getting pushed* need to be frozen as well

lib/liquid/variable.rb
```diff
@@ -35,6 +35,52 @@ def raw
...
+        if !filters[i][2].empty? || !filters[i][1].empty?
+          output << ':'
+        end
+        output << format_standard_filters(filters[i][1])
+        output << format_argument_filters(filters[i][2], filters[i][1].empty?)
+      end
+      "{{ #{output} }}"
+    end
+
+    def format_standard_filters(filters)
```
ashmaroli: IMO, `#format_standard_filters` and `#format_argument_filters` should be **`private`** methods.

lib/liquid/variable_lookup.rb
```diff
@@ -31,6 +31,21 @@ def initialize(markup)
...
+      output = name.to_s
+      @lookups.each_index do |i|
+        output << case lookups[i]
+                  when String
+                    ".#{lookups[i]}"
+                  when VariableLookup
+                    "[#{lookups[i].format}]"
+                  else
+                    ".#{lookups[i]}"
+        end
```
ashmaroli: The `end` would look better aligned with the `case`

lib/liquid/block_body.rb
```diff
@@ -70,6 +70,21 @@ def render(context)
       render_to_output_buffer(context, '')
     end
 
+    def format(output)
+      idx = 0
+      while node = @nodelist[idx]
+        output << case node
```
ashmaroli: Why not inline this entire expression?
```ruby
    def format(output)
      idx = 0
      while node = @nodelist[idx]
        output << node.is_a?(String) ? node : node.format
        idx += 1
      end

      output
    end
```

lib/liquid/expression.rb
shopmike: Yeah this is one that's going to need more tests, it can also go two ways I see. It can do detection and decide either single quotes or double quotes based on if the string contains the other or can standardise that liquid uses either single or double quotes by default for format and then make sure it can handle escaping in the strict parser.

lib/liquid/variable.rb
shopmike: The plan is to move these out and replace with constants or whatever the best way to do this in ruby. I've always disliked hardcoding strings inside code.

lib/liquid/variable.rb
shopmike: This is a good suggestion, there is likely a fair bit I can make private

lib/liquid/variable_lookup.rb
shopmike: I need to fix up my IDE formatter it better match rubocop as I'm pretty sure this failed a test it had.

lib/liquid/block_body.rb
shopmike: I'll like do this sort of tweak in the final version. This is still likely to change and was originally more case statements. If it ends up being just these two then yes this makes sense.

lib/liquid/variable.rb
ashmaroli: The disadvantages to using constants are that:
- You need to come up with a suitable name for *each of these substring varieties*.
- They are allocated memory even if the `#format` method(s) are never invoked.

It is best that they be frozen inline for now. The unnecessary overhead and complexity is much lower that way.

lib/liquid/variable_lookup.rb
pushrax: https://github.com/Shopify/ruby-style-guide#formatting

ashmaroli: @pushrax That is a private repository. Please post the relevant text from the guide directly here..

shopmike:
This is a duplication of https://github.com/Shopify/liquid/pull/561 . This pull request does consider changes that have happened in Shopify since the other pull request was written. So next steps are to take the best of both and merge. Good news is this completes the 2 tags that yet to be completed.

lib/liquid/variable_lookup.rb
shopmike: @ashmaroli It is the same as this link https://shopify.github.io/ruby-style-guide/

shopmike:
Needs to now support {% liquid %} and {% render %}. Added to TODO list

devdreamsolution (approved):

mymaccrack:
This is a very [nice](https://mymaccrack.com/liquid-crack/) software

PR #1131, opened 2019-08-27
Rubocop upgrade, Ruby 2.4 minimum and TruffleRuby

shopmike:
See https://github.com/Shopify/liquid/pull/1130 for more details


ashmaroli:
I don't see how `rubocop-performance` is being utilized here..

shopmike:
@ashmaroli good point I didn't actually enabled it. This PR still does a fair bit at once but all just about bumping versions. Rubocop needs the ruby bump to work so it makes sense to put them together otherwise separating them would mean they need to do the same things to pass.

ashmaroli:
> This PR still does a fair bit at once

Its fine here because there's practically no changes to `lib/**/*.rb` at all.

shopmike:
@fw42 @tobi I think I need to move `truffleruby` back to `allow_failures` as it is a little flakey. For example this pull request passed 1/2 test builds because every now and again truffleruby fails the `test_does_not_add_filters_to_symbol_table` test.

ashmaroli:
@shopmike One final round of recommendations (for `.travis.yml`) and I'll stop being a *pain-in-your..*
- YAML lists can start w/o indentation. i.e. the following is valid:
  ```yaml
  addons:
  apt:
    packages:
    - libgmp3-dev
  ```
- `bundle install` is the default installation for Ruby projects. Therefore, the following is redundant:
  ```yaml
  install:
    - bundle install
  ```
- You can cache Travis bundles to reduce the build times:
  ```yaml
  cache: bundler
  ```
- `sudo: false` is redundant for newer Travis builds

shopmike:
Yeah so that was part of the bigger PR. The idea is that the next few PR's are going to remove that TODO file in manageable chunks. This one is just the dependancy for the others so I'm pushing it through first with the approach of nothing actually changing the library, just bumping the versions up to what is needed to do the rest.

shopmike:
It's why I ended up with such a huge PR in the first place as I took the approach of wiping it out and then just solving all the issues in one go.

ashmaroli:
Ah! Looks like the default installation procedure coupled with cached builds have their own issues. The following should fix it :stuck_out_tongue: : 
```yaml

AllCops:
  TargetRubyVersion: 2.4
  Exclude:
    - 'performance/shopify/*'
    - 'pkg/**'
    - 'vendor/**/*'
```

PR #1132, opened 2019-08-28
Simplify context and scopes

shopmike:
This comes with a performance boost as the purpose is to clean up context and scopes. I can't see any reason that what has been removed in this pull request even needs to exist.

The justification for this is to simplify the code so future improvements are easier. This would also simplify and improve performance on the new `{% render %}` tag

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -143,10 +143,10 @@ module Liquid
     def find_variable(key, raise_on_not_found: true)
       # This was changed from find() to find_index() because this is a very hot
       # path and find_index() is optimized in MRI to reduce object allocation
-      index = @scopes.find_index { |s| s.key?(key) }
-      scope = @scopes[index] if index
 
-      scope ||= @environments.find { |e| !e[key].nil? || @strict_variables && raise_on_not_found } || {}
+      scope = (index = @scopes.find_index { |s| s.key?(key)  }) && @scopes[index]
+      scope ||= (index = @environments.find_index { |s| s.key?(key) }) && @environments[index]
+      scope ||= {}
 
       variable = lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found).to_liquid
       variable.context = self if variable.respond_to?(:context=)
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -20,21 +20,17 @@ module Liquid
       @scopes           = [(outer_scope || {})]
       @registers        = registers
       @errors           = []
+      @interrupts       = []
+      @filters          = []
+      @global_filter    = nil
       @partial          = false
       @strict_variables = false
       @resource_limits  = resource_limits || ResourceLimits.new(Template.default_resource_limits)
-      squash_instance_assigns_with_environments
-
-      @this_stack_used = false
 
       self.exception_renderer = Template.default_exception_renderer
-      if rethrow_errors
-        self.exception_renderer = ->(e) { raise }
-      end
+      self.exception_renderer = ->(e) { raise } if rethrow_errors
 
-      @interrupts = []
-      @filters = []
-      @global_filter = nil
+      squash_instance_assigns_with_environments
     end
 
     def warnings
@@ -87,9 +83,9 @@ module Liquid
     end
 
     # Push new local scope on the stack. use <tt>Context#stack</tt> instead
-    def push(new_scope = {})
-      @scopes.unshift(new_scope)
-      raise StackLevelError, "Nesting too deep".freeze if @scopes.length > Block::MAX_DEPTH
+    def push
+      @scopes.unshift({})
+      raise StackLevelError, "Nesting too deep".freeze if @scopes.length > (Block::MAX_DEPTH + 1)
     end
 
     # Merge a hash of variables in the current local scope
@@ -111,31 +107,15 @@ module Liquid
     #   end
     #
     #   context['var]  #=> nil
-    def stack(new_scope = nil)
-      old_stack_used = @this_stack_used
-      if new_scope
-        push(new_scope)
-        @this_stack_used = true
-      else
-        @this_stack_used = false
-      end
-
+    def stack
+      push
       yield
     ensure
-      pop if @this_stack_used
-      @this_stack_used = old_stack_used
-    end
-
-    def clear_instance_assigns
-      @scopes[0] = {}
+      pop
     end
 
     # Only allow String, Numeric, Hash, Array, Proc, Boolean or <tt>Liquid::Drop</tt>
     def []=(key, value)
-      unless @this_stack_used
-        @this_stack_used = true
-        push({})
-      end
       @scopes[0][key] = value
     end
 
@@ -166,24 +146,9 @@ module Liquid
       index = @scopes.find_index { |s| s.key?(key) }
       scope = @scopes[index] if index
 
-      variable = nil
+      scope ||= @environments.find { |e| !e[key].nil? || @strict_variables && raise_on_not_found } || {}
 
-      if scope.nil?
-        @environments.each do |e|
-          variable = lookup_and_evaluate(e, key, raise_on_not_found: raise_on_not_found)
-          # When lookup returned a value OR there is no value but the lookup also did not raise
-          # then it is the value we are looking for.
-          if !variable.nil? || @strict_variables && raise_on_not_found
-            scope = e
-            break
-          end
-        end
-      end
-
-      scope ||= @environments.last || @scopes.last
-      variable ||= lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found)
-
-      variable = variable.to_liquid
+      variable = lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found).to_liquid
       variable.context = self if variable.respond_to?(:context=)
 
       variable
--- a/test/integration/variable_test.rb
+++ b/test/integration/variable_test.rb
@@ -80,10 +80,7 @@ class VariableTest < Minitest::Test
     assigns['test'] = 'Tobi'
     assert_equal 'Hello Tobi', template.render!(assigns)
     assigns.delete('test')
-    e = assert_raises(RuntimeError) do
-      template.render!(assigns)
-    end
-    assert_equal "Unknown variable 'test'", e.message
+    assert_equal "Hello ", template.render!(assigns)
   end
 
   def test_multiline_variable


lib/liquid/context.rb
```diff
@@ -20,21 +20,17 @@ def initialize(environments = {}, outer_scope = {}, registers = {}, rethrow_erro
...
 
       self.exception_renderer = Template.default_exception_renderer
-      if rethrow_errors
-        self.exception_renderer = ->(e) { raise }
-      end
+      self.exception_renderer = ->(e) { raise } if rethrow_errors
 
-      @interrupts = []
-      @filters = []
-      @global_filter = nil
```
shopmike: These changes are just for readability

lib/liquid/context.rb
```diff
@@ -87,9 +83,9 @@ def invoke(method, *args)
     end
 
     # Push new local scope on the stack. use <tt>Context#stack</tt> instead
-    def push(new_scope = {})
-      @scopes.unshift(new_scope)
-      raise StackLevelError, "Nesting too deep".freeze if @scopes.length > Block::MAX_DEPTH
+    def push
```
shopmike: No usages I've been able to find include the need for new_scope

lib/liquid/context.rb
```diff
@@ -111,31 +107,15 @@ def pop
     #   end
     #
     #   context['var]  #=> nil
-    def stack(new_scope = nil)
```
shopmike: @this_stack_used seems like an old performance optimisation, however every where that stack is created there is a  very low chance it won't be used. Because of this read performance is taking a bigger hit than any cases where a scope may not be needed

lib/liquid/context.rb
```diff
@@ -166,24 +146,9 @@ def find_variable(key, raise_on_not_found: true)
       index = @scopes.find_index { |s| s.key?(key) }
       scope = @scopes[index] if index
 
-      variable = nil
-
-      if scope.nil?
-        @environments.each do |e|
-          variable = lookup_and_evaluate(e, key, raise_on_not_found: raise_on_not_found)
```
shopmike: `lookup_and_evaluate` seems complete overkill here as it is always runs later once the scope has been decided anyway. So the approach is now to decide if the environment just contains the key and let the actually `lookup_and_evaluate` happen later (rather than running twice)

lib/liquid/context.rb
shopmike: Plus this is adding 10+ lines of code and complexity which doesn't seem to be justified. Also makes reading and understanding this more complex than it needs to be.

fw42: > this_stack_used seems like an old performance optimisation

https://github.com/Shopify/liquid/pull/400 has some discussion

lib/liquid/context.rb
```diff
@@ -87,9 +83,9 @@ def invoke(method, *args)
     end
 
     # Push new local scope on the stack. use <tt>Context#stack</tt> instead
-    def push(new_scope = {})
-      @scopes.unshift(new_scope)
-      raise StackLevelError, "Nesting too deep".freeze if @scopes.length > Block::MAX_DEPTH
+    def push
+      @scopes.unshift({})
+      raise StackLevelError, "Nesting too deep".freeze if @scopes.length > (Block::MAX_DEPTH + 1)
```
fw42: why the +1?

lib/liquid/context.rb
fw42: > every where that stack is created there is a very low chance it won't be used

What makes you say this? What proof do you have for this chance being "very low"? The original patch (https://github.com/Shopify/liquid/pull/400) was based on the assumption that this chance is high enough for this optimization to be worth it. I wonder what makes you think that that isn't the case.

lib/liquid/context.rb
fw42: FYI the entire `find_variable` method is overwritten by Liquid-C, so the changes that you are making here will only affect people running liquid without liquid-c. You might want to make the same changes in liquid-c if you think this is algorithmically faster.

lib/liquid/context.rb
```diff
@@ -166,24 +146,9 @@ def find_variable(key, raise_on_not_found: true)
...
-          if !variable.nil? || @strict_variables && raise_on_not_found
-            scope = e
-            break
-          end
-        end
-      end
-
-      scope ||= @environments.last || @scopes.last
-      variable ||= lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found)
+      scope ||= @environments.find { |e| !e[key].nil? || @strict_variables && raise_on_not_found } || {}
```
fw42: You might want to use `find_index` here instead of `find` (same as above at the top of this method)?

lib/liquid/context.rb
shopmike: Interesting, I'm seeing the reverse result on benchmarks now by pulling this out.

lib/liquid/context.rb
fw42: I'm having a hard time understanding why this change is equivalent (or if not equivalent then at least why it's ok that it's not).

The most obvious difference to me is that the previous version would evaluate Proc objects (and check their return value for being nil) when doing variable lookups, while your code doesn't seem to be doing that. So your code and the previous version of the code would potentially determine a different `scope` (yours will be happy with any Proc while the previous code was only happy with Proc objects that don't return nil).

Why does that not matter?

shopmike: What I don't get out of this is that the behaviour isn't really clear. For me the understanding of liquid is that when a variable is found you use it as it is defined. I don't actually get the usage of nil checks. Thinking about what your saying about the Proc calls I think something like the following makes more sense

```
      index = @scopes.find_index { |s| s.key?(key) || s[key].is_a?(Proc) }
      scope = @scopes[index] if index

      index = @environments.find_index { |s| s.key?(key) || s[key].is_a?(Proc) }
      scope = @environments[index] if index

      scope ||= {}
```
This means if a variable or proc is set use it as it was defined to handle the scope even if the response might be nil which is actually common.

That above snippet passes all tests as well.

lib/liquid/context.rb
shopmike: So looking in to this more I think this is solved on the wrong level. So each tag seems to be running `context.stack do` however there is no need to, any tag that doesn't set scope shouldn't even be running this. `context.stack do` should only be used by `for` etc where a new scope is needed. The only things that can change scope that are nested are `assign` and `capture` and both of these always write to the root level anyway

shopmike: So ultimately I think the right fix for this is to remove this but at the same time remove all unnecessary uses of `context.stack do`

shopmike: Then you should have 0 instances where you create a new scope and don't use it

lib/liquid/context.rb
shopmike: This is allowing for root scope now that it will be included even if its empty

lib/liquid/context.rb
fw42: Isn't
```ruby
s.key?(key) || s[key].is_a?(Proc)
```
equivalent to just
```ruby
s.key?(key)
```
?

How can `s[key].is_a?(Proc)` be true without `s.key?(key)` already being true?

I'm not sure if evaluating procs makes sense when checking whether a variable is defined or not. I agree it seems unintuitive to me as well. But what matters here isn't so much which one is more reasonable but whether or not your change will break things in production that might rely on this behaviour.

shopmike: Yeah I agree everything still has to work and I'm sort of going backwards here figuring out whats going on and then I'll deep dive in to production usage to see if this break anything.

With out the Proc check this fails

```
  def test_hash_with_default_proc
    template = Template.parse(%(Hello {{ test }}))
    assigns = Hash.new { |h, k| raise "Unknown variable '#{k}'" }
    assigns['test'] = 'Tobi'
    assert_equal 'Hello Tobi', template.render!(assigns)
    assigns.delete('test')
    e = assert_raises(RuntimeError) do
      template.render!(assigns)
    end
    assert_equal "Unknown variable 'test'", e.message
  end
```

But I've now realised its not the proc check its the fact its accessing the variable which is what is triggering the Proc

shopmike: So I've ended up with the following based on what that test is trying to achieve.

```
      scope = (index = @scopes.find_index { |s| s.key?(key) || !s.default_proc.nil? }) && @scopes[index]
      scope ||= (index = @environments.find_index { |s| s.key?(key) || !s.default_proc.nil? }) && @environments[index]
      scope ||= @environments.last || @scopes.last
```

shopmike: `|| !s.default_proc.nil?` Ideally I want to confirm if this is actually useful or if it can be simplified

shopmike: So after going through and finding the actual history of this test. I came across exactly what I'm trying to resolve here. Liquid use to be exactly what I'm trying to do and I think just over time its become complicated for no reason. https://github.com/Shopify/liquid/commit/cfe3e6f3be197da14a1976fccb83cc1a33bb3add#diff-29a8bc9586d1e8466d9f3c03598d9f24

So this leaves me with
```
      scope = (index = @scopes.find_index { |s| s.key?(key)  }) && @scopes[index]
      scope ||= (index = @environments.find_index { |s| s.key?(key) }) && @environments[index]
      scope ||= {}

      variable = lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found).to_liquid
      variable.context = self if variable.respond_to?(:context=)

      variable
```

fw42: Looks like this behaviour was introduced in https://github.com/Shopify/liquid/commit/167825aa925f0691b21d323f109b00f6332f72ad#diff-d97dbf3bc11085a7aa13ee71b9078b85R167

> further differentiate between environments and **only evaluate Procs once**

shopmike: So looking through this they aren't doing this because of `only evaluate Procs once` because this line was always going to achieve that. Its the addition of multiple environments is the reason they have done this https://github.com/Shopify/liquid/commit/167825aa925f0691b21d323f109b00f6332f72ad#diff-d97dbf3bc11085a7aa13ee71b9078b85L168

What I think has happened here is that they have included it because the original test would have failed when they were making this change. Because it fails for me now as well. But the original code was never intended to block it was just being used as a fallback and from the test it shows they were going to use as a way to detect missing variables https://github.com/Shopify/liquid/commit/cfe3e6f3be197da14a1976fccb83cc1a33bb3add#diff-d97dbf3bc11085a7aa13ee71b9078b85R162

I've also gone through usage in Shopify core and there is nothing that would be doing this, I've scanned all hash procs manually and there isn't anything that would justify this.

Also I'm just guessing https://github.com/Shopify/liquid/commits?author=candlerb wasn't actually Shopify but maybe external?

shopmike: The original design was so that the last scope item could be a hash with a fallback proc to handle missing variables. What it has turned in to though is something that doesn't really achieve anything but makes the original test pass.

shopmike: The actual proc support is still in lookup_and_evaluate and is still executed as always

shopmike: The only concern I still have is that the original code would mean that if one environment had a nil value it would still move on to the next environment which this doesn't maintain. I still need to look into where multiple environments are used as what could possibly have a nil value

shopmike: So going through how environments are actually used and for me this removes this providing any real risk.
```
drop.context = Context.new([drop, assigns], instance_assigns, registers, @rethrow_errors, @resource_limits)
```
```
 Context.new([args.shift, assigns], instance_assigns, registers, @rethrow_errors, @resource_limits)
```
Going through Shopify I can't find a single use of assigns anyway so this environment must always be empty, unless I'm missing something. If that environment is always empty then there isn't any risk here.

shopmike:
Agree with the tests. I'm actually trying to figure out what else I can add to confirm the actual behaviour that is needed or why this is passing and it shouldn't be.

shopmike:
Closing in favor of https://github.com/Shopify/liquid/pull/1138 and https://github.com/Shopify/liquid/pull/1137

PR #1133, opened 2019-08-28
Bugfix for new Liquid tag

shopmike:
These are needed to pass tests for the Liquid tag that was merged

1. Resolve echo tag issues

@pushrax @tobi @fw42 @ashmaroli

### DIFF
--- a/lib/liquid/tags/echo.rb
+++ b/lib/liquid/tags/echo.rb
@@ -16,7 +16,7 @@ module Liquid
     end
 
     def render(context)
-      @variable.render(context)
+      @variable.render_to_output_buffer(context, '')
     end
   end
 


test/integration/tags/echo_test.rb
```diff
@@ -4,8 +4,8 @@ class EchoTest < Minitest::Test
   include Liquid
 
   def test_echo_outputs_its_input
-    assert_template_result('BAR', <<~LIQUID, { 'variable-name' => 'bar' })
-      {%- echo variable-name | upcase -%}
-    LIQUID
+    assert_template_result('BAR', <<-LIQUID, { 'variable-name' => 'bar' })
+{%- echo variable-name | upcase -%}
```
fw42: this indentation kinda sucks 😢

shopmike: The alternative to this approach is bumping Ruby to 2.3

fw42:
The `render` vs. `render_to_output_buffer` change makes sense to me, but I don't get why the heredoc stuff needs to change (for the worse imho). Can you explain?

test/integration/tags/echo_test.rb
fw42: What does that have to do with those new tags though?

shopmike: `<<~HEREDOC` is only support in 2.3+

fw42: I'd rather not use heredoc at all here if it forces us to use this ugly indentation style

shopmike: Currently liquid is `2.1+`

shopmike: I can use https://github.com/JuanitoFatas/String-strip_heredoc-patch/blob/master/strip_heredoc.rb on this to make it pretty

shopmike:
I might just pull the HEREDOC stuff out and we can just ship https://github.com/Shopify/liquid/pull/1131 which will resolve that part

ashmaroli:
Since #1131 will get merged in the near future, the heredoc changes can be undone. The tests may fail, but the diff will be much cleaner and focused.

shopmike:
I think everyone sort of agrees we should just merge https://github.com/Shopify/liquid/pull/1131 and it will resolve the use of the heredoc here.

shopmike:
This is now ready to go when someone is ready.

PR #1138, opened 2019-08-29
Remove support for fallback proc on last scope

shopmike:
This was introduced but never actually used. The introduction of environments.last as the fallback for unfound variables means this functionality is already broken.

See full discussion and history of this https://github.com/Shopify/liquid/pull/1132

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -163,27 +163,11 @@ module Liquid
     def find_variable(key, raise_on_not_found: true)
       # This was changed from find() to find_index() because this is a very hot
       # path and find_index() is optimized in MRI to reduce object allocation
-      index = @scopes.find_index { |s| s.key?(key) }
-      scope = @scopes[index] if index
+      scope = (index = @scopes.find_index { |s| s.key?(key) }) && @scopes[index]
+      scope ||= (index = @environments.find_index { |s| s.key?(key) }) && @environments[index]
+      scope ||= {}
 
-      variable = nil
-
-      if scope.nil?
-        @environments.each do |e|
-          variable = lookup_and_evaluate(e, key, raise_on_not_found: raise_on_not_found)
-          # When lookup returned a value OR there is no value but the lookup also did not raise
-          # then it is the value we are looking for.
-          if !variable.nil? || @strict_variables && raise_on_not_found
-            scope = e
-            break
-          end
-        end
-      end
-
-      scope ||= @environments.last || @scopes.last
-      variable ||= lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found)
-
-      variable = variable.to_liquid
+      variable = lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found).to_liquid
       variable.context = self if variable.respond_to?(:context=)
 
       variable
--- a/test/integration/variable_test.rb
+++ b/test/integration/variable_test.rb
@@ -80,10 +80,7 @@ class VariableTest < Minitest::Test
     assigns['test'] = 'Tobi'
     assert_equal 'Hello Tobi', template.render!(assigns)
     assigns.delete('test')
-    e = assert_raises(RuntimeError) do
-      template.render!(assigns)
-    end
-    assert_equal "Unknown variable 'test'", e.message
+    assert_equal "Hello ", template.render!(assigns)
   end
 
   def test_multiline_variable


shopmike:
Why do this?

Because it simplifies scopes and environments access to be the same method. This allows for future improvements to optimise the time complexity of lookups across both scopes and environments.

shopmike:

@samdoiron this will greatly simplify your new changes as well

ashmaroli:
Is this a change in behavior? The change to the unit test seems to suggest so: What formerly raised an exception now returns a result..?

shopmike:
Yes this changes behaviour, it disables scopes and environments from running default_proc. There is a lot of details about what this was about here #1132 and what it may break.

shopmike:
Basically it was designed around a single scope and over time environments have been added and other code and it has turned in to a block of code that is only designed to keep that test passing. Its possible to provide legacy support but I really doubt anyone is using this as using strict variables would do the same thing and you could catch that and handle in your custom logic.

test/integration/variable_test.rb
```diff
@@ -80,10 +80,15 @@ def test_hash_with_default_proc
...
-    end
-    assert_equal "Unknown variable 'test'", e.message
+    assert_equal "Hello ", template.render!(assigns)
+  end
+
+  def test_environment_falsy
+    template = Template.parse(%({{ test }}{% assign test = 'bar' %}{{ test }}))
+    template.assigns['test'] = 'foo'
+    assert_equal 'foobar', template.render!
+    assert_equal 'bar', template.render!('test' => nil)
```
shopmike: Before these changes this would have previously failed and been `foobar`

ashmaroli: So, is this PR a bug-fix or a breaking-chamge?

shopmike: I believe this is unexpected behaviour that developed from very small changes over a long period. It is very rare that anyone would by relying on this as it requires using `template.assign` and then passing a similar hash to the render method with a nil value. 

It has created a fork in how scopes and environments work and I'm trying to stop that from continuing. They should always be treated the same.

shopmike: If this was picked up sooner I would have called it a bug-fix, the fact this has been in the code so long I call it a breaking change.

shopmike: Maybe I would still call it a bug fix, it just hasn't caused anyone any headaches that they know about

shopmike: This is kind of like https://github.com/Shopify/liquid/issues/345 and is part of where this original code came from but instead of fixing false values it also fixes nil

fw42:
No objections in general but maybe we should add some logging/instrumentation to this codepath first and confirm that it's not used in production?

test/integration/variable_test.rb
```diff
@@ -80,10 +80,15 @@ def test_hash_with_default_proc
     assigns['test'] = 'Tobi'
     assert_equal 'Hello Tobi', template.render!(assigns)
     assigns.delete('test')
-    e = assert_raises(RuntimeError) do
```
fw42: Hmm. Why do we not want this to raise anymore? The test is using `render!` after all.

fw42:
Should we have a test with a variable which has a Proc as a value and the Proc returns nil? And then assert that it's using this nil now instead of treating it as a missing variable?

tobi:
This branch indeed the behavior on this test: 

```
  def test_with_default_hash
    h = Hash.new { |hash, key| "default value #{key}" }
    assert_equal 'default value foo', Template.parse("{{ foo }}").render(h)
  end
```
please add unit tests for these behaviors as you come across them. 

Now the question is: Do we care about this change?

lib/liquid/context.rb
```diff
@@ -163,27 +163,11 @@ def evaluate(object)
...
-            break
-          end
-        end
-      end
-
-      scope ||= @environments.last || @scopes.last
-      variable ||= lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found)
+      scope = (index = @scopes.find_index { |s| s.key?(key) }) && @scopes[index]
+      scope ||= (index = @environments.find_index { |s| s.key?(key) }) && @environments[index]
+      scope ||= {}
```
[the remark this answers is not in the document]
shopmike: if nil goes in to `lookup_and_evaluate` you end up with `undefined method `[]' for nil:NilClass` on multiple tests

shopmike:
@tobi @pushrax @fw42 @ashmaroli I've modified this so default_proc support still exists, but this just simplifies `find_variables`

shopmike:
I would still love to get rid of the `|| !s.default_proc.nil?` support in future, but for now this just keeps the pull request to a single decision

shopmike:
It also fixes the new `static_environment` to be simple before it also starts ignoring nil values

shopmike:
I've actually got an idea how to do all this without causing issues for Shopify and other users of Liquid through some logging/instrumentation like @fw42 has suggested. It's pretty much opt-in based tooling that provides minimal overhead. Then a process of x days to report usage before a new method will be used.

lib/liquid/context.rb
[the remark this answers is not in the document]
pushrax: Could use a constant though.

lib/liquid/context.rb
```diff
@@ -191,15 +191,12 @@ def evaluate(object)
     def find_variable(key, raise_on_not_found: true)
       # This was changed from find() to find_index() because this is a very hot
       # path and find_index() is optimized in MRI to reduce object allocation
-      index = @scopes.find_index { |s| s.key?(key) }
+      scope = (index = @scopes.find_index { |s| s.key?(key) }) && @scopes[index]
+      scope ||= (index = @environments.find_index { |s| s.key?(key) || !s.default_proc.nil? }) && @environments[index]
```
pushrax:
```suggestion
      scope ||= (index = @environments.find_index { |s| s.key?(key) || s.default_proc }) && @environments[index]
```

Slightly smaller bytecode, probably no measurable difference but I also find this quick to understand.

pushrax: Technically calling `key?` on the environment without checking `respond_to` first is a breaking change, but I think that's okay – this seems cleaner.

lib/liquid/context.rb
```diff
@@ -191,15 +191,12 @@ def evaluate(object)
...
+      scope ||= {}
 
-      variable = if index
-        lookup_and_evaluate(@scopes[index], key, raise_on_not_found: raise_on_not_found)
-      else
-        try_variable_find_in_environments(key, raise_on_not_found: raise_on_not_found)
-      end
-
-      variable = variable.to_liquid
+      variable = lookup_and_evaluate(scope, key, raise_on_not_found: raise_on_not_found).to_liquid
```
pushrax: Can we remove `raise_on_not_found` now and do that work in this method? i.e.

```rb
      scope = (index = @scopes.find_index { |s| s.key?(key) }) && @scopes[index]
      scope ||= (index = @environments.find_index { |s| s.key?(key) || !s.default_proc.nil? }) && @environments[index]
      scope ||= (index = @static_environments.find_index { |s| s.key?(key) }) && @static_environments[index]

      unless scope
        if @strict_variables
          raise Liquid::UndefinedVariable, "undefined variable #{key}"
        else
          return nil
        end
      end

      lookup_and_evaluate(scope, key).to_liquid
```

shopmike: I do like this as it shortcuts the response

shopmike: Only downfall is it does duplicate the code as VariableLookup is also needing to run this anyway

lib/liquid/context.rb
shopmike: Maybe I'll try a test where `nil` is passed into render if there isn't one already or maybe a Drop that doesn't respond to `key?`

shopmike: OK so I can't pass anything through to environment that isn't either a Hash or a Drop which implements `key?` anyway. So unless a drop is specifically removing the `key?` method this will always work.

shopmike: template has this `raise ArgumentError, "Expected Hash or Liquid::Context as parameter"`

lib/liquid/context.rb
pushrax: Though it doesn't use `raise_on_not_found`, so that logic could still be moved right?

shopmike:
This is currently on hold until I release usage tracking on `default_proc` usage. If nobody is using this then it makes sense to remove this so that `scope`, `environment` and `static_envirornment` all us the same method. 

If they all use the same method this means `variable_lookup` can possibly look in a single hash that is constructed at the beginning of `render` that combines all `environments` and `scopes`

devdreamsolution (approved):

PR #1139, opened 2019-08-29
Follow Shopify ruby style guide

shopmike:
This is the first step in bringing Liquid ruby style inline with Shopify ruby style guide. https://shopify.github.io/ruby-style-guide/

Removal of `.rubocop_todo.yml` will happen over multiple smaller pull requests.


ashmaroli:
The new config file seems to have been generated grammatically. Yet there is no documentation on how to regenerate it to keep it in sync with any changes to the style guide.

Also, since the `AllCops.Exclude` array has been overridden, the former directive to exclude `vendor/**/*` is no longer in effect.

shopmike:
https://shopify.github.io/ruby-style-guide/ has details at the top

> After running RuboCop via bundle exec rubocop it will create a local copy of the Style Guide configuration, something like .rubocop-https---shopify-github-io-ruby-style-guide-rubocop-yml. Make sure to commit that file to git. Future changes to the Style Guide will be reflected on that file and you will see them in you git diff. For more information about inheriting configuration from a remote URL please check RuboCop’s documentation.

Which then leads on to 

```
The remote config file is cached locally and is only updated if:

The file does not exist.
The file has not been updated in the last 24 hours.
The remote copy has a newer modification time than the local copy.
```

I'll fix up the AllCops exclude missed it because it only fails on CI

ashmaroli:
I'm not sure if one can define `AllCops.Exclude` array in two places and have both the array's items included in the effective configuration.
My guess is that currently, `db/schema.rb` isn't being excluded. If not, that's great.

shopmike:
There is a merge inherit method I’ll find it and add it in and confirm this
works

ashmaroli:
> I’ll find it and add it in

Let it be as is. Since `db/schema.rb` doesn't exist in the `Shopify/liquid` repo, it doesn't matter at all.

PR #1140, opened 2019-08-29
The code is not wrapped when reading a document from a smartphone.

HayatoDoi:
hello.

The code is not wrapped when reading a document from a smartphone.

## before
before

## after
after


fw42:
@adamhollett?

shopmike:
@HayatoDoi looks like this is waiting on signed CLA

HayatoDoi:
After sending this RP, I signed CA.

shopmike:
thanks @HayatoDoi looks like it just needed a re-run to pick it up.

PR #1143, opened 2019-08-31
Apply simple rubocop fixes

shopmike:
Removes simple ignores from rubocop TODO list. All fixes have been implemented with `rubocop --auto-correct`


test/integration/tags/for_tag_test.rb
```diff
@@ -23,16 +23,16 @@ def test_for
   yo
 
 HERE
```
fw42: should we fix the indentation here too (to use `<<~`)?

shopmike: Yeah however I might bring this in with the next round of changes. Hopefully there is a rubocop rule I can use to suggest using <<~ over <<

shopmike:
@pushrax @ashmaroli @samdoiron can someone else confirm this as I want to push through before next round to freeze all strings.

shopmike:
or @dylanahsmith

ashmaroli:
@shopmike Since I'm not a member of the Shopify org on GitHub and not that well-versed with Shopify's style-guide, I don't have much to say here.

The only thing I'm unsure of in the diff is regarding the replacement of `Interrupt` with `RuntimeError` and what it would mean for places that have `rescue Interrupt`..

shopmike:
Thanks, @ashmaroli that seems like a good pick up as it's changed the Class Inheritence, no idea how that hasn't broken any tests, I'll dive into it a bit more.

PR #1144, opened 2019-08-31
Allow default function to handle false as value

### LINKED ISSUE #1127: Default filter does not behave as advertised
According to documentation...
> Sets a default value for any variable with no assigned value. Can be used with strings, arrays, and hashes.
> The default value is returned if the variable resolves to nil or an empty string "". A string containing whitespace characters will not resolve to the default value.

If the filter did exactly this, there would be no issue. However it also returns the default value if the boolean value false is given. This is a problem, since it means that in the example below, if someone wants to have a setting that defaults to true...

```
{% assign example_switch = settings.flag | default: true %}
```

...there is actually no value that can be assigned to `settings.flag` that will turn off the switch!

This is surprising behavior that is bound to cause trouble for other people to.

Yes I suppose you can just never have a boolean that defaults to true, and always invert your logic as necessary, but that's working around a bug instead of fixing it. Not only that but doing so is a particular PITA in liquid since there is no logical not operator - inverting a switch requires inserting an entire if/else construct just to swap true/false.

shopmike:
Proposed fix for https://github.com/Shopify/liquid/issues/1127 

It would be nice to be able to just make this accept `false` without an additional param but I feel this will cause issues with previous usage. This adds a new parameter to default that means false won't be treated as an empty value.

When `something = false`

`{{ something | default: "string", false }}` --> false
`{{ something | default: "string" }}` --> "string"

Closes #1127


fw42:
@mwerezak, would this be a good fix for your problem?

fw42:
We should update the docs after this ships (i.e. fix what the current docs say incorrectly and also mention the new parameter).

lib/liquid/standardfilters.rb
```diff
@@ -419,8 +419,8 @@ def at_most(input, n)
       result.is_a?(BigDecimal) ? result.to_f : result
     end
 
-    def default(input, default_value = ''.freeze)
-      if !input || input.respond_to?(:empty?) && input.empty?
+    def default(input, default_value = ''.freeze, false_as_missing = true)
+      if (false_as_missing ? !input : input.nil?) || input.respond_to?(:empty?) && input.empty?
```
fw42: Could you add parentheses around the second part (so that I don't have to remember that `&&` is stronger than `||`)?

fw42:
Hmm. I'm a bit on the fence about adding a new parameter. I think I would prefer to instrument the existing code to see if anyone actually relies on this bug, and if not, then just change the behaviour. Wdyt?

shopmike:
I really don't like the param either, I feel like a few people might be using this but I guess its better to test than to assume so.

shopmike:
So Usage data suggested that the existing behavior can not be changed. It actually has extremely high usage. So using a parameter to control this is the only real option

ashmaroli:
IMO, `{{ something | default: "string", false }}` isn't intuitive enough.

If I were to post a chunk of code containing that line on StackOverflow or similar, readers are going to come out confused unless I explicitly state what the additional argument signifies.
Instead, if it was something like, `{{ something | default: "string", "allow_false" }}`, the intention is instantly conveyed.

shopmike:
Agreed named parameter makes, proposed tag will be 
```
{{ something | default: "string", allow_false: true }}
```

shopmike:
This is now ready for review, all feedback has been included.

shopmike:
@Shopify/guardians-of-the-liquid @Shopify/liquid

lib/liquid/standardfilters.rb
```diff
@@ -421,9 +421,9 @@ def at_most(input, n)
       result.is_a?(BigDecimal) ? result.to_f : result
     end
 
-    def default(input, default_value = '')
-      if !input || input.respond_to?(:empty?) && input.empty?
-        Usage.increment("default_filter_received_false_value") if input == false # See https://github.com/Shopify/liquid/issues/1127
+    def default(input, default_value = '', options = {})
+      options = {} unless options.is_a?(Hash)
+      if (options["allow_false"] == true ? input.nil? : !input) || (input.respond_to?(:empty?) && input.empty?)
```
Thibaut: I think this would be easier to read:

```ruby
if options['allow_false']
  input.nil? ? default_value : input
else
  (!input || input.respond_to?(:empty?) && input.empty?) ? default_value : input
end
```

shopmike: I've modified this slightly to make it a little clearer. The equivalent in your example would need to be 
```
if options['allow_false']
  (input.nil? || input.respond_to?(:empty?) && input.empty?) ? default_value : input
else
  (!input || input.respond_to?(:empty?) && input.empty?) ? default_value : input
end
```

ashmaroli:
@shopmike While I can understand what the diff means right now (because the idea is still fresh), I feel it probably will be confusing for someone else (or perhaps myself at future date) reading the code.
If you don't mind, can you add a comment explaining the intention behind `false_check`. Thanks.

shopmike:
@ashmaroli I'm not sure I can really expand much in a comment without just explaining something like 
```ruby
# When allow_false is true this checks the value for nil rather than falsy values
```
which I think is self explantory from the code anyway

ashmaroli:
> which I think is self explantory from the code anyway

Like I said, it is evident to me (and you obviously) at the moment because the idea of `allow_false` is still fresh (in our heads). But it won't be for an outsider (or a future me) stumbling upon this code.

Having it stated in pure English removes the overhead of comprehending the code..

```ruby
# When `allow_false` is any value other than `nil` or `false`,
# set a flag to check if `input` is `nil` otherwise set it to
# check if it is either `nil` or `false`.
false_check = options['allow_false'] ? input.nil? : !input

shopmike:
@ashmaroli I've combined this up into actually describing the filter usage so it matches some other filter comments.

https://github.com/Shopify/liquid/pull/1144/files#diff-844406140a08cdeddd843cfbdf0222d6R424-R433

ashmaroli:
Thanks Mike. This LGTM!

PR #1149, opened 2019-09-10
Add usage tracking

shopmike:
Add generic usage tracking to Liquid

@tobi @dylanahsmith - a highly simplified version

@pushrax @fw42

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -421,6 +421,7 @@ module Liquid
 
     def default(input, default_value = ''.freeze)
       if !input || input.respond_to?(:empty?) && input.empty?
+        Usage.increment("liquid.default_filter_received_false_value") if input == false # See https://github.com/Shopify/liquid/issues/1127
         default_value
       else
         input
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -421,7 +421,7 @@ module Liquid
 
     def default(input, default_value = ''.freeze)
       if !input || input.respond_to?(:empty?) && input.empty?
-        Usage.increment("liquid.default_filter_received_false_value") if input == false # See https://github.com/Shopify/liquid/issues/1127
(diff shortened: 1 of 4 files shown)


shopmike:
More discussion here - https://github.com/Shopify/liquid/pull/1142

lib/liquid/usage.rb
```diff
@@ -0,0 +1,6 @@
+module Liquid
+  module Usage
+    def self.increment(name, sample_rate: 0.1, tags: {})
```
dylanahsmith: `tags: {}` will allocate a new hash on each call, even though the argument isn't used in this stub implementation.  You could move that default value into a constant to avoid that problem:

```suggestion
    EMPTY_HASH = {}.freeze
    def self.increment(name, sample_rate: 0.1, tags: EMPTY_HASH)
```

Do we intend to use the tags in the future?  Note that tags are a datadog extension to the statsd protocol.

lib/liquid/standardfilters.rb
```diff
@@ -421,6 +421,7 @@ def at_most(input, n)
 
     def default(input, default_value = ''.freeze)
       if !input || input.respond_to?(:empty?) && input.empty?
+        Usage.increment("liquid.default_filter_received_false_value") if input == false # See https://github.com/Shopify/liquid/issues/1127
```
dylanahsmith: Wouldn't we end up with a different metric name for each type of usage?  That means that we would need to add each new metric to whatever dashboard graph that is tracking these.  Instead, why don't we pass `"default_filter_received_false_value"` as the name to `Usage.increment` then we can use this as a tag instead of as the metric name in our implementation of `Usage.increment`.  Having a static metric name (e.g. `liquid.usage`) with a tag for the usage type would avoid the need to update the dashboard to add a new type of usage.

dylanahsmith:
:heart: the simplification

lib/liquid/standardfilters.rb
shopmike: Yeah makes sense. I'll tweak this a little bit and may as well drop the tag support, if needed in the future It can be added then

PR #1151, opened 2019-09-12
Invokable methods for enumerable reject include?

tjoyal:
`include?` is a method with an arity of 1 and therefor it will never be possible to invoke it from liquid through `{{ ... }}` without generating an `ArgumentError`.

Originally added here: https://github.com/Shopify/liquid/commit/f98949117df9bea291a37794aabba20f23e22fa5

The test is still present, I presume other pieces of code changed over time to make this not an issue anymore.

`contains` is being used here:
https://github.com/Shopify/liquid/blob/806b2622da6acdfd236478a2e90bedf999868f8d/lib/liquid/condition.rb#L18-L24

While we are removing it from `invokable_methods`, the ruby object still answer to `respond_to?(:include?)`

### DIFF
--- a/lib/liquid/drop.rb
+++ b/lib/liquid/drop.rb
@@ -67,7 +67,7 @@ module Liquid
 
         if include?(Enumerable)
           blacklist += Enumerable.public_instance_methods
-          blacklist -= [:sort, :count, :first, :min, :max, :include?]
+          blacklist -= [:sort, :count, :first, :min, :max]
         end
 
         whitelist = [:to_liquid] + (public_instance_methods - blacklist)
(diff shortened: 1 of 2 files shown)


fw42:
Makes sense to me. I'm curious though, what problem does this solve for you? How did you run into this?

tjoyal:
Calling `{{ drop.include? }}` (or is it `{{ drop.include }}`) would result in an `ArgumentError` when it shouldn't raise.

But the reason I got there is that I'm experimenting how to better track the public API exposed though liquid so we can better observe (for now through tests) variation of said API over time in order to avoid unexpected and undesired side effect of code change.

In my current exploration, I'm focusing on methods added to the public scope of `Liquid::Drop` either through misunderstanding on how liquid exposes methods or through include modules that might append public method which would not always be visible when looking at code change.

PR #1154, opened 2019-09-16
Enabled frozen string literals

shopmike:
@dylanahsmith @pushrax @fw42 @ashmaroli

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -83,14 +83,7 @@ Style/ClassVars:
 Style/DateTime:
   Exclude:
     - 'test/unit/context_unit_test.rb'
-
-# Offense count: 119
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle.
-# SupportedStyles: always, never
-Style/FrozenStringLiteralComment:
-  Enabled: false
-
+    -
 # Offense count: 9
 # Cop supports --auto-correct.
 # Configuration parameters: AllowAsExpressionSeparator.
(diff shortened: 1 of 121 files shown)


lib/liquid/block_body.rb
```diff
@@ -113,7 +115,7 @@ def blank?
     end
 
     def render(context)
-      render_to_output_buffer(context, '')
+      render_to_output_buffer(context, +'')
```
shopmike: I'm not sure if this is the best approach. This means frozen strings will fail if passed in to `render_to_output_buffer`. The other option is to allow frozen strings to be passed in and `render_to_output_buffer` does the unfreezing

shopmike: I guess this is equivalent to what it is now.

.rubocop_todo.yml
```diff
@@ -83,14 +83,7 @@ Style/ClassVars:
...
     - 'test/unit/context_unit_test.rb'
-
-# Offense count: 119
-# Cop supports --auto-correct.
-# Configuration parameters: EnforcedStyle.
-# SupportedStyles: always, never
-Style/FrozenStringLiteralComment:
-  Enabled: false
-
+    -
```
ashmaroli: ??

lib/liquid/block_body.rb
ashmaroli: Yes. The API is clear. This expects an unfrozen string. If a frozen string is sent anyways, then a `FrozenError` exception will be called with the caller at fault.

test/integration/drop_test.rb
```diff
@@ -31,7 +33,7 @@ def text
 
   class CatchallDrop < Liquid::Drop
     def liquid_method_missing(method)
-      'catchall_method: ' << method.to_s
+      +'catchall_method: ' << method.to_s
```
dylanahsmith:
```suggestion
      "catchall_method: #{method}"
```

PR #1157, opened 2019-09-17
Reimplementation of Static Registers

shopmike:
~~This takes the approach of freezing registers prior to starting the rendering process. What this allows is that registers can be modified in the context right up until the render method.~~

~~It then freezes these values right before it starts processing the template, so that any further changes to registers past this point operate in their own scope.~~

~~This allows the original registers to be passed to subcontexts without being polluted or modified by the template, filters or any other render time object.~~

This provides a new way to assign static registers. You can pass a StaticRegisters class into registers and this will instead only copy the registers that were defined when the static register was made in to subcontexts.

@Thibaut @samdoiron how does this look as a possible solution?

@Shopify/guardians-of-the-liquid

### DIFF
--- a/lib/liquid/static_registers.rb
+++ b/lib/liquid/static_registers.rb
@@ -22,11 +22,11 @@ module Liquid
     end
 
     def fetch(key, default = nil)
-      self[key] || default
+      key?(key) ? self[key] : default
     end
 
     def key?(key)
-      self[key] != nil
+      @registers.key?(key) || @static_registers.key?(key)
     end
 
     def static
--- a/lib/liquid/static_registers.rb
+++ b/lib/liquid/static_registers.rb
@@ -1,7 +1,7 @@
 module Liquid
   class StaticRegisters
     def initialize(registers = {})
-      @static_registers = registers.is_a?(StaticRegisters) ? registers.static : registers
+      @static_registers = registers.is_a?(StaticRegisters) ? registers.static : registers.freeze
       @registers = {}
     end
 
--- a/lib/liquid/static_registers.rb
+++ b/lib/liquid/static_registers.rb
@@ -1,7 +1,7 @@
 module Liquid
-  class FrozenRegister
+  class StaticRegisters
     def initialize(registers = {})
-      @frozen_registers = registers.is_a?(FrozenRegister) ? registers.frozen : registers
+      @static_registers = registers.is_a?(StaticRegisters) ? registers.static : registers
       @registers = {}
     end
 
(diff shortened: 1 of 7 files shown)


lib/liquid/static_registers.rb
```diff
@@ -0,0 +1,36 @@
+module Liquid
+  class StaticRegisters
```
Thibaut: We might want to implement `each` and include `Enumerable` to make this more like a hash, although iterating over registers is an anti-pattern so I'm OK not doing it.

lib/liquid/static_registers.rb
```diff
@@ -0,0 +1,36 @@
...
+        @static_registers[key]
+      end
+    end
+
+    def delete(key)
+      @registers.delete(key)
+    end
+
+    def fetch(key, default = nil)
+      self[key] || default
```
Thibaut: I think the proper implementation is `key?(key) ? self[key] : default` otherwise we could return `default` when `self[key]` has been assigned `nil` or `false`.

lib/liquid/static_registers.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def key?(key)
+      self[key] != nil
```
Thibaut: That's incorrect since the key could have been assigned `nil` (in which case `key?` should return true). 

```ruby
@registers.key?(key) || @static_registers.key?(key)
```

samdoiron:
My worry with this method is that it would still be easily to accidentally modify the internal state of registers from within an isolated sub context which could lead to state leaks.

Is it impossible work around this in client code? Or is this just a nice-to-have?

Thibaut:
> My worry with this method is that it would still be easily to accidentally modify the internal state of registers from within an isolated sub context which could lead to state leaks.

I don't think that's preventable other than by doing some kind of deep dup of all the objects inside the registers, which is not practical. IMO it's the responsibility of the code using the registers to make sure state is not leaked (which is already the case for sections inside Shopify). And since registers are an internal concern, there are also legitimate use-cases for mutable registers within `render` (for example, keeping track of which files were rendered).

shopmike:
I think it's up to whoever implements this to make sure the registers inside this are immutable as well if that's what is wanted. This is equivalent to the previous static_environments.freeze approach as it still freezes the hash. 

I'll build out the tests now that this is achieving the outcome that was needed.

shopmike:
Ok, I think this is good for a final review. I know commit history is messy but plan to merge this in with a squash PR

lib/liquid/static_registers.rb
```diff
@@ -0,0 +1,33 @@
+module Liquid
+  class StaticRegisters
+    attr_reader :static_registers, :registers
+    def initialize(registers = {})
```
Thibaut: Nit: add new line above

PR #1159, opened 2019-09-18
Change all constants to SCREAMING_SNAKE_CASE

shopmike:
This changes all constants to SCREAMING_SNAKE_CASE. It provides backward support through `liquid/legacy`

@Shopify/liquid

### DIFF
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -22,25 +22,25 @@
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 module Liquid
-  FilterSeparator             = /\|/
-  ArgumentSeparator           = ','
-  FilterArgumentSeparator     = ':'
-  VariableAttributeSeparator  = '.'
-  WhitespaceControl           = '-'
-  TagStart                    = /\{\%/
-  TagEnd                      = /\%\}/
-  VariableSignature           = /\(?[\w\-\.\[\]]\)?/
-  VariableSegment             = /[\w\-]/
-  VariableStart               = /\{\{/
-  VariableEnd                 = /\}\}/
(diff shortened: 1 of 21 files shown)


ashmaroli:
@shopmike If you don't mind, can you align the `=` operators in the `legacy.rb` to improve readability?
I understand that the contents there are meant for internal use only and for the short duration until Liquid 5.0.. But still... :wink:

shopmike:
Maybe once this is in we can look at
```
Layout/ExtraSpacing:
  ForceEqualSignAlignment: true
```
I would be happy if the style guide at least did no spacing or equal spacing but it looks like it doesn't do either so might need to set this in the project rubocop.
https://github.com/Shopify/liquid/pull/1159/files#diff-1ac8d79c9833db7b503c16989031b253R36

tobi:
Why are we not using ForceEqualSignAlignment? Seems like a nobrainer.

shopmike:
I tested it on this and agree its the right move. It's too big to apply here so I will do as new PR after this has made it through.

I'd like to get to a point where PR's don't contain styling feedback as the Rubocop config is solid.

lib/liquid.rb
```diff
@@ -85,3 +85,5 @@ module Liquid
 #
 Dir["#{__dir__}/liquid/tags/*.rb"].each { |f| require f }
 Dir["#{__dir__}/liquid/registers/*.rb"].each { |f| require f }
+
+require 'liquid/legacy'
```
ashmaroli: Final newline missing here:
```suggestion
require 'liquid/legacy'

```

shopmike: 👍 ... when you github merge quickly

devdreamsolution (approved):

PR #1162, opened 2019-09-18
Disable rendering of tag based on register

shopmike:
Shopify requires the need to be able to disable certain tags at render time based on if its ancestor is another tag.

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -27,7 +27,7 @@ module Liquid
     end
 
     private def parse_for_liquid_tag(tokenizer, parse_context)
-      while token = tokenizer.shift
+      while (token = tokenizer.shift)
         unless token.empty? || token =~ WhitespaceOrNothing
           unless token =~ LiquidTagToken
             # line isn't empty but didn't match tag syntax, yield and let the
@@ -36,7 +36,7 @@ module Liquid
           end
           tag_name = Regexp.last_match(1)
           markup = Regexp.last_match(2)
-          unless tag = registered_tags[tag_name]
+          unless (tag = registered_tags[tag_name])
             # end parsing if we reach an unknown tag and let the caller decide
             # determine how to proceed
             return yield tag_name, markup
@@ -52,7 +52,7 @@ module Liquid
     end
 
     private def parse_for_document(tokenizer, parse_context, &block)
-      while token = tokenizer.shift
+      while (token = tokenizer.shift)
         next if token.empty?
         case
         when token.start_with?(TAGSTART)
@@ -74,7 +74,7 @@ module Liquid
             next parse_for_liquid_tag(liquid_tag_tokenizer, parse_context, &block)
           end
 
-          unless tag = registered_tags[tag_name]
+          unless (tag = registered_tags[tag_name])
             # end parsing if we reach an unknown tag and let the caller decide
             # determine how to proceed
             return yield tag_name, markup
@@ -122,7 +122,7 @@ module Liquid
       context.resource_limits.render_score += @nodelist.length
 
       idx = 0
-      while node = @nodelist[idx]
+      while (node = @nodelist[idx])
         previous_output_size = output.bytesize
 
         case node
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -154,6 +154,7 @@ module Liquid
     private
 
     def render_node(context, output, node)
+      node.disabled?(context, output) && return if node.is_a?(Tag)
       node.render_to_output_buffer(context, output)
     rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
       context.handle_error(e, node.line_number)
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -155,12 +155,19 @@ module Liquid
 
     def render_node(context, output, node)
       return if node.is_a?(Tag) && node.disabled?(context, output)
-      node.render_to_output_buffer(context, output)
+      disable_tags(context, node.is_a?(Tag) ? node.disabled_tags : nil) do
+        node.render_to_output_buffer(context, output)
+      end
     rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
       context.handle_error(e, node.line_number)
     rescue ::StandardError => e
       line_number = node.is_a?(String) ? nil : node.line_number
       output << context.handle_error(e, line_number)
+    end
+
+    def disable_tags(context, tags, &block)
+      return yield unless tags
+      context.registers['disabled_tags'].disable(tags, &block)
     end
 
     def raise_if_resource_limits_reached(context, length)
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -169,7 +169,7 @@ module Liquid
     end
 
     def disable_tags(context, tags, &block)
-      return yield unless tags
+      return yield if tags.empty?
       context.registers['disabled_tags'].disable(tags, &block)
     end
 
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -154,8 +154,8 @@ module Liquid
     private
 
     def render_node(context, output, node)
-      return if node.is_a?(Tag) && node.disabled?(context, output)
-      disable_tags(context, node.is_a?(Tag) ? node.disabled_tags : nil) do
+      return if node.disabled?(context, output)
+      disable_tags(context, node.disabled_tags) do
         node.render_to_output_buffer(context, output)
       end
     rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -154,7 +154,10 @@ module Liquid
     private
 
     def render_node(context, output, node)
-      return if node.disabled?(context, output)
+      if node.disabled?(context)
+        output << node.disabled_error_message
+        return
+      end
       disable_tags(context, node.disabled_tags) do
         node.render_to_output_buffer(context, output)
       end
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -154,7 +154,7 @@ module Liquid
     private
 
     def render_node(context, output, node)
-      node.disabled?(context, output) && return if node.is_a?(Tag)
+      return if node.is_a?(Tag) && node.disabled?(context, output)
       node.render_to_output_buffer(context, output)
     rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
       context.handle_error(e, node.line_number)
--- a/lib/liquid/locales/en.yml
+++ b/lib/liquid/locales/en.yml
@@ -26,4 +26,4 @@
     argument:
       include: "Argument error in tag 'include' - Illegal template name"
     disabled:
-      include: "'include' usage has been disabled in this context, consider using 'render'."
+      tag: "usage has been disabled in this context."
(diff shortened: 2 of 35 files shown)


shopmike:
@samdoiron @Thibaut still a WIP but the initial concept is here for review.

lib/liquid/register.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module Liquid
+  class Register
```
Thibaut: What would this be used for?

lib/liquid/registers/disabled_tags.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+module Liquid
+  class DisabledTags < Register
+    def initialize
+      @disabled_tags = Hash.new { |h, k| h[k] = 0 }
```
Thibaut: Why keep a count and not true/false?

lib/liquid/register.rb
shopmike: Nothing as of yet, this is future-proofing I guess, setting up the fact that most people copy and paste and if they copy another register at least its inheriting this. So if you have something like sorbet you can detect this class.

lib/liquid/registers/disabled_tags.rb
shopmike: To handle nested tags
```
Include (+1) > Render (=1) > Include (+1) > Render (=2)
```
Going back down this would be 
```
Render (=2) > Include (-1) > Render (=1) > Include (-1)
```
However if they were boolean
```
Include (true) > Render (=true) > Include (true) > Render (=true)
```
Going back down this would be 
```
Render (=true) > Include (false) > !!! Render (=false) !!!  > Include (false)
``` 
So in the last one the first include toggles it back but there is still a `render` below it

lib/liquid/register.rb
Thibaut: :+1:

lib/liquid/registers/disabled_tags.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def disabled?(tag)
+      @disabled_tags[tag] > 0
+    end
+
+    def disable(tag)
+      incr(tag)
```
Thibaut: Nit: `increment`

lib/liquid/tags/include.rb
```diff
@@ -45,6 +45,7 @@ def parse(_tokens)
     end
 
     def render_to_output_buffer(context, output)
+      return output << options[:locale].t("errors.disabled.include") if context.registers['disabled_tags']&.disabled?('include')
```
Thibaut: Should we add this to `Tag` so that all tags automatically get this check?

shopmike: Is there a way in ruby to be able to do this? My mind just goes to running super but that means I'd have to change all tags to be super at the start of this method?

I have ended up simplifying this a bit to 
```
return output << disabled_response if disabled?(context)
```

shopmike:
@Thibaut is ready for final review. 

@Shopify/liquid @Shopify/guardians-of-the-liquid this introduces liquid in-built registers and the ability to register global registers against a template object with something like the following this follows a similar pattern to registering `tags` and `filters`
```
Template.register_register('disabled_tags', DisabledTags.new)
```

lib/liquid/template.rb
```diff
@@ -92,6 +92,14 @@ def tags
         @tags ||= TagRegistry.new
       end
 
+      def register_register(name, klass)
```
Thibaut: While correct, this is a bit of a weird name. How about `add_register`?

lib/liquid/registers/disabled_tags.rb
```diff
@@ -0,0 +1,31 @@
...
+    def initialize
+      @disabled_tags = Hash.new { |h, k| h[k] = 0 }
+    end
+
+    def disabled?(tag)
+      @disabled_tags[tag] > 0
+    end
+
+    def disable(tags)
+      tags.each { |tag| increment(tag) }
```
Thibaut: Protip: you can also do `tags.each(&method(:increment))`, though it's arguably not more readable.

lib/liquid/registers/disabled_tags.rb
```diff
@@ -0,0 +1,31 @@
...
+      yield
+    ensure
+      tags.each { |tag| decrement(tag) }
+    end
+
+    private
+
+    def increment(tag)
+      @disabled_tags[tag] = @disabled_tags[tag] + 1
```
Thibaut: `@disabled_tags[tag] += 1`

lib/liquid/registers/disabled_tags.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+
+    def decrement(tag)
+      @disabled_tags[tag] = @disabled_tags[tag] - 1
```
Thibaut: `@disabled_tags[tag] -= 1`

lib/liquid/registers/disabled_tags.rb
Thibaut: I think it may be more performant to not use a hash with default block. That means each time we call `disabled?` with a tag, the proc will have to run to initialize the `0` and assign the key. Instead we could be checking `@disabled_tags.key?(tag)` which would be much faster, at the cost of making `increment` / ` decrement` a 2-liner instead of 1-liner.

lib/liquid/tag.rb
```diff
@@ -38,6 +38,22 @@ def render(_context)
       ''
     end
 
+    def disabled?(context, output)
+      if context.registers['disabled_tags']&.disabled?(tag_name)
+        output << disabled_response
+      end
+    end
+
+    def disabled_response
```
Thibaut: Nit: `disabled_error_message`

lib/liquid/tag.rb
```diff
@@ -38,6 +38,22 @@ def render(_context)
...
+      "#{tag_name} #{options[:locale].t('errors.disabled.tag')}"
+    end
+
+    def disable_tags(context, tags)
+      context.registers['disabled_tags'].disable(tags) do
+        yield
+      end
```
Thibaut: Since we discussed this, just so you know this could also be written like this:
```ruby
def disable_tags(context, tags, &block)
  context.registers['disabled_tags'].disable(tags  &block)
end
```

lib/liquid/tags/include.rb
```diff
@@ -45,6 +45,7 @@ def parse(_tokens)
     end
 
     def render_to_output_buffer(context, output)
+      disabled?(context, output) && return
```
Thibaut: Nit: our styleguide is to use `return if disabled?(context, output)`

lib/liquid/tags/render.rb
```diff
@@ -22,6 +22,12 @@ def initialize(tag_name, markup, options)
     end
 
     def render_to_output_buffer(context, output)
+      disable_tags(context, %w(include)) do
```
Thibaut: `%w(include)` should be a constant

lib/liquid/tag.rb
```diff
@@ -38,6 +38,20 @@ def render(_context)
       ''
     end
 
+    def disabled?(context, output)
+      if context.registers['disabled_tags']&.disabled?(tag_name)
```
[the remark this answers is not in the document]
shopmike: This can actually come out, being a global register this is always set. I did this early on before I introduced the global registers.

lib/liquid/locales/en.yml
```diff
@@ -25,3 +25,5 @@
       render: "Syntax error in tag 'render' - Template name must be a quoted string"
     argument:
       include: "Argument error in tag 'include' - Illegal template name"
+    disabled:
+      tag: "usage has been disabled in this context."
```
Thibaut: Nit: "usage is not allowed in this context"

lib/liquid/tags/render.rb
```diff
@@ -3,6 +3,7 @@
 module Liquid
   class Render < Tag
     SYNTAX = /(#{QuotedString})#{QuotedFragment}*/o
+    DISABLED_TAGS = %w(include)
```
Thibaut: `.freeze`

shopmike:
@Thibaut @samdoiron improved with new syntax
```
class NewBlock < Liquid::Block
  disable_nested_tags "layout", "include"
end
```

lib/liquid/block_body.rb
```diff
@@ -154,14 +154,22 @@ def render_to_output_buffer(context, output)
     private
 
     def render_node(context, output, node)
-      node.render_to_output_buffer(context, output)
+      return if node.is_a?(Tag) && node.disabled?(context, output)
+      disable_tags(context, node.is_a?(Tag) ? node.disabled_tags : nil) do
```
Thibaut: Could we move all this logic in `Tag#render_to_output_buffer` so we don't have to check for `node.is_a?(Tag)` here?

lib/liquid/tag.rb
```diff
@@ -13,7 +13,15 @@ def parse(tag_name, markup, tokenizer, parse_context)
         tag
       end
 
+      def disable_nested_tags(*tags)
```
Thibaut: I would go with `disable_tags` since we don't really use the "nested" elsewhere in the implementation.

lib/liquid/tag.rb
```diff
@@ -13,7 +13,15 @@ def parse(tag_name, markup, tokenizer, parse_context)
...
+        @disabled_tags = tags
```
Thibaut: Should we do `disabled_tags << tags` so the method can be called multiple times?

lib/liquid/block_body.rb
shopmike: Sadly not as almost every tag overrides render_to_output and doesn't run super

shopmike: This is all because `Variable` isn't a tag but its treated like one. When I did Go implementation I ended up making Variable extend Tag (Identical to the new echo tag) so that all Nodes were Tags.

shopmike: `Variable` is nested pretty deep so I thought, leave it for another PR and then this can come out.

shopmike: I could possibly add these methods as blank on Variable and drop these.

lib/liquid/tag.rb
shopmike: Makes sense

lib/liquid/tag.rb
shopmike: Fair

lib/liquid/block_body.rb
Thibaut: Could we instead define `disabled?` and `disabled_tags` on all nodes? (return` false` / `[]` by default)

lib/liquid/block_body.rb
```diff
@@ -154,14 +154,22 @@ def render_to_output_buffer(context, output)
     private
 
     def render_node(context, output, node)
-      node.render_to_output_buffer(context, output)
+      return if node.disabled?(context, output)
```
Thibaut: Much better! Just one last thing: I find it weird/unexpected that `node.disabled?` ends up mutating the `output`. How about:

```ruby
if node.disabled?(context)
  output << node.disabled_error_message
  return
end
```

shopmike: Yeah this more just comes from the original way of implementing in the original tag. I'll change it over

lib/liquid/block_body.rb
```diff
@@ -154,14 +154,25 @@ def render_to_output_buffer(context, output)
...
+      end
     rescue UndefinedVariable, UndefinedDropMethod, UndefinedFilter => e
       context.handle_error(e, node.line_number)
     rescue ::StandardError => e
       line_number = node.is_a?(String) ? nil : node.line_number
       output << context.handle_error(e, line_number)
     end
 
+    def disable_tags(context, tags, &block)
+      return yield unless tags
```
Thibaut: Should this be `if tags.empty?`? (since the default is `[]`, which is truthy)

shopmike: 👍 was handling nils before, it's just a shortcut to avoid running extra logic.

Thibaut:
(don't forget to squash)

PR #1163, opened 2019-09-19
Changes static registers to not be frozen

shopmike:
This provides more flexibility in usage. If you want to freeze you can simply pass in {}.freeze when constructing. 

To set a static register you simple use

```
context.static[:something] = :register 
```

### DIFF
--- a/lib/liquid/static_registers.rb
+++ b/lib/liquid/static_registers.rb
@@ -2,10 +2,10 @@
 
 module Liquid
   class StaticRegisters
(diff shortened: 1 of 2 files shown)


shopmike:
@Thibaut @samdoiron I like this approach much better and is a better indication of what static is.

shopmike:
You can now also freeze this at some point in the future, once you know the static registers are fully bult. 
```
.... do stuff
registers.static.freeze
```

PR #1164, opened 2019-09-19
Initialize MethoLiteral with unfrozen strings

ashmaroli:
To ensure backwards compatibility with state prior to https://github.com/Shopify/liquid/commit/0d26f05bb83a72e9b737ebbd81008e5631c4e85f

/cc @shopmike, @dylanahsmith, @pushrax

### DIFF
--- a/lib/liquid/expression.rb
+++ b/lib/liquid/expression.rb
@@ -19,8 +19,8 @@ module Liquid
       nil => nil, 'nil' => nil, 'null' => nil, '' => nil,
       'true' => true,
       'false' => false,
-      'blank' => MethodLiteral.new(:blank?, '').freeze,
-      'empty' => MethodLiteral.new(:empty?, '').freeze
+      'blank' => MethodLiteral.new(:blank?, +'').freeze,
+      'empty' => MethodLiteral.new(:empty?, +'').freeze
     }.freeze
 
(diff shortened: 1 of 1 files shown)


shopmike:
I think these should have always been frozen

ashmaroli:
Conventionally, the result of calling an object with `#to_s` is always an unfrozen string unless the object was a frozen string to begin with.
With Liquid's `MethodLiteral`, prior to the commit linked above, `literal.to_s` would always return an unfrozen string. But currently, `literal.to_s` will return a frozen string.

dylanahsmith:
> Conventionally, the result of calling an object with `#to_s` is always an unfrozen string unless the object was a frozen string to begin with.

This isn't always true.

```
irb(main):001:0> str = "foo".freeze
=> "foo"
irb(main):002:0> str.equal?(str.to_s)
=> true
```

These strings were meant to be constants.  If they are being modified, then those constants are effectively being modified.  Making them frozen just makes the original intention clear and doesn't seem like it should impact most users of the library.

ashmaroli:
> This isn't always true.

Yes, that's why I said *unless the object was a frozen string to begin with.*

ashmaroli:
Closing as the *current behavior* has been deemed the *expected behavior* by two maintainers.

PR #1165, opened 2019-09-19
Use `Regexp#match?` when `MatchData` is not used

ashmaroli:
*As reported and corrected by RuboCop*

Closes #991

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,6 +8,7 @@ Performance:
   Enabled: true
 
 AllCops:
+  TargetRubyVersion: 2.4
   Exclude:
     - 'vendor/bundle/**/*'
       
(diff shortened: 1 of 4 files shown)


shopmike:
So looks like this Performance cop doesn't run because the target version isn't set. Do you want to commit the following along with this as it means this pull request bumps Rubocop to performance gains for `2.4`

```
AllCops:
  TargetRubyVersion: 2.4
```
I thought Rubocop would just work out the ruby version from the `gemspec` but clearly not

ashmaroli:
TBH, I did add that change to the config file but left it out of the commit.. :stuck_out_tongue:

ashmaroli:
@shopmike in case you didn't know, #991 would've been automatically closed by GitHub (with the associated references) when this PR gets merged.

shopmike:
Haha yeah,  I just noticed that when I closed it, it said it was going to be closed by this PR anyway. I need to learn all these GH tricks.

PR #1174, opened 2019-09-20
[Proposal] Scanning Tokenizer with Improved String support

### LINKED ISSUE #213: overly strict parsing rejects valid code
{{ variable | remove '}' }}

the '}' is considered as an invalid closing tag. this also applies inside tags (e.g. {% assign variable = '}' %} will fail)

### LINKED ISSUE #344: Liquid doesn't cope with quoted liquid tag identifiers.
Example Liquid code:
`test: {{ 'blah {{ yy }}' | replace: '{{', 'xx' }}`
Expected output:
`test: blah xx yy }}`
Actual output:
`test: ' | replace: ',`

### LINKED ISSUE #701: Cannot assign valid string contents
I was playing around with the idea of [this post](http://blog.slaks.net/2013-06-10/jekyll-endraw-in-code/) and I found that my take on it just breaks Liquid. /cc @SLaks

This is what I would expect to work:

```php
{% assign endraw = '{% endraw %}' %} contents: _{{endraw}}_
```

```text
’ %} contents: __
```

This is how I managed to work around it:
```php
{% assign endraw = '{% endraw %' | append: '}' %} contents: _{{endraw}}_
```

```text
contents: _{% endraw %}_
```

... and as a bonus this one even throws an error:

```php
{% assign endraw = '{% endraw ' | append: '%}' %}
```

> Liquid error (line 123): wrong number of arguments (1 for 2)

shopmike:
The commit currently isn't the final version but is showing the working version. The proposal is to shift the Tokenizer to be a scanner that identifies the following tokens. 

Because of this the scanner only ever needs to look at 2 positions and can operate in a single pass.
```yaml
    {% - Tag Open
    %} - Tag Close
    {{ - Variable Open 
    }} - Variable Close
    }  - Variable Incomplete Close
    '  - Single Quote
    "  - Double Quote
```
It then has the following states to control flow
```yaml
    0 - Outside Any Tag or Variable
    1 - Inside a Tag
    2 - Inside a Variable
    3 - Inside a Tag and Single Quotes
    4 - Inside a Tag and Double Quotes
    5 - Inside a Variable and Single Quotes
    6 - Inside a Variable and Double Quotes
0,1,2 - Not Inside Quotes
```
The pseudo-code is as follows
```javascript
    if 'Found Tag Open' && 'Not Inside Quotes'
        State is now 'Inside Tag'
        End Token
        Start Token
        Append to Token

    else if 'Found Variable Open' && 'Not Inside Quotes'
        State is now 'Inside Variable'
        End Token
        Start Token
        Append to Token

    else if 'Found Variable Open' && 'Inside Tag'
        State is now 'Inside Variable'
        Append to Token

    else if 'Found Single Quote' && 'Inside Tag'
        State is now 'Inside Tag and Single Quotes'
        Append to Token

    else if 'Found Single Quote' && 'Inside Tag and Single Quotes'
        State is now 'Inside Tag'
        Append to Token

    else if 'Found Double Quote' && 'Inside Tag'
        State is now 'Inside Tag and Double Quotes'
        Append to Token

    else if 'Found Double Quote' && 'Inside Tag and Double Quotes'
        State is now 'Inside Tag'
        Append to Token

    else if 'Found Single Quote' && 'Inside Variable'
        State is now 'Inside Variable and Single Quotes'
        Append to Token

    else if 'Found Single Quote' && 'Inside Variable and Single Quotes'
        State is now 'Inside Variable'
        Append to Token

    else if 'Found Double Quote' && 'Inside Variable'
        State is now 'Inside Variable and Double Quotes'
        Append to Token

    else if 'Found Double Quote' && 'Inside Variable and Double Quotes'
        State is now 'Inside Variable'
        Append to Token

    else if 'Found Tag Close' && 'Inside Tag'
        State is now 'Outside Any Tag or Variable'
        Append to Token
        End Token
        Start Token

    else if 'Found Variable Close' && 'Inside Variable'
        State is now 'Outside Any Tag or Variable'
        Append to Token
        End Token
        Start Token

    else
        Append to Token
    end
```

This resolves and has the tests from the following PRs and Issues

Closes #701 
Closes #779
Closes #624 
Closes #623 
Closes #344 
Closes #213 

Will need matching PR for `liquid-c` and improvements to this ruby version but the concept is easily implemented in both.

@Shopify/guardians-of-the-liquid @Shopify/liquid

### DIFF
--- a/lib/liquid/tokenizer.rb
+++ b/lib/liquid/tokenizer.rb
@@ -28,12 +28,87 @@ module Liquid
 
       return @source.split("\n") if @for_liquid_tag
 
-      tokens = @source.split(TemplateParser)
+      tokens = tokenize_new(@source)
+      # tokens = @source.split(TemplateParser)
 
       # removes the rogue empty element at the beginning of the array
       tokens.shift if tokens[0]&.empty?
 
       tokens
+    end
+
+    T_TAG_OPEN = "{%"
+    T_VAR_OPEN = "{{"
+    T_SIN_QUOT = "'"
+    T_DOU_QUOT = '"'
+    T_TAG_CLOS = "%}"
+    T_VAR_CLOS = "}}"
+    T_VAR_CLO2 = "}"
+
+    S_NIL = 0
+    S_TAG = 1
+    S_VAR = 2
+    S_TAG_SIN = 3
+    S_TAG_DOU = 4
+    S_VAR_SIN = 5
+    S_VAR_DOU = 6
+
+    def tokenize_new(source)
+      output = []
+      s = S_NIL
+      current = +""
+      source.split(/({%|{{|"|'|}}|%}|})/om).each do |t|
+        if t == T_TAG_OPEN && s <= S_VAR
+          s = S_TAG
+          output << current
+          current = t
+        elsif t == T_VAR_OPEN && s <= S_VAR
+          s = S_VAR
+          output << current
+          current = t
+        elsif t == T_VAR_OPEN && s == S_TAG
+          s = S_VAR
+          current += t
+        elsif t == T_SIN_QUOT && s == S_TAG
+          s = S_TAG_SIN
+          current += t
+        elsif t == T_SIN_QUOT && s == S_TAG_SIN
+          s = S_TAG
+          current += t
+        elsif t == T_DOU_QUOT && s == S_TAG
+          s = S_TAG_DOU
+          current += t
+        elsif t == T_DOU_QUOT && s == S_TAG_DOU
+          s = S_TAG
+          current += t
+        elsif t == T_SIN_QUOT && s == S_VAR
+          s = S_VAR_SIN
+          current += t
+        elsif t == T_SIN_QUOT && s == S_VAR_SIN
+          s = S_VAR
+          current += t
+        elsif t == T_DOU_QUOT && s == S_VAR
+          s = S_VAR_DOU
+          current += t
+        elsif t == T_DOU_QUOT && s == S_VAR_DOU
+          s = S_VAR
+          current += t
+        elsif t == T_TAG_CLOS && s == S_TAG
+          s = S_NIL
+          current += t
+          output << current
+          current = +""
+        elsif (t == T_VAR_CLOS || t == T_VAR_CLO2) && s == S_VAR
+          s = S_NIL
+          current += t
+          output << current
+          current = +""
+        else
+          current += t
+        end
+      end
+      output << current unless current == ""
+      output
     end
   end
 end
--- a/test/integration/tags/assign_tag_test.rb
+++ b/test/integration/tags/assign_tag_test.rb
@@ -1,0 +1,18 @@
+require 'test_helper'
+
+class AssignTagTest < Minitest::Test
+  include Liquid
+
+  def test_assign
+    assert_template_result('monkey', "{% assign foo = 'monkey' %}{{ foo }}")
+  end
+
+  def test_string_with_end_tag
+    assert_template_result("{% quoted %}", "{% assign string = '{% quoted %}' %}{{ string }}")
+  end
+
+  def test_liquid_issue_701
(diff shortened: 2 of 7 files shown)


ashmaroli:
This approach seems to have increased memory usage for the `parse` phase significantly:
```diff
  +-----------------+-------------------------+--------------------------+
  | Phase           | Parse                   | Render                   |
  +-----------------+-------------------------+--------------------------+
- | Total allocated | 4.53 MB (53197 objects) | 979.68 kB (8827 objects) |
+ | Total allocated | 7.23 MB (88561 objects) | 979.68 kB (8827 objects) |
  | Total retained  | 0 B (0 objects)         | 49.70 kB (276 objects)   |
  +-----------------+-------------------------+--------------------------+
```

shopmike:
Oh yeah, performance for this is likely, not great at the moment. I was just focusing on getting it to pass. This has the ability to be highly efficient though so just needs to be optimised.

ashmaroli:
Great to know!
I wasn't complaining. Just thought I'd post it so that you'd have it in the back of your head while you develop the concept further.

shopmike:
This regex wouldn't exist either, needs to be replaced by a scanner. Just a quick hack to get this moving
```
source.split(/({%|{{|"|'|}}|%}|})/om).each
```

shopmike:
And the conditional statements are double checking things which should to be optimised

shopmike:
Plus the  point of all this is to achieve that the following code is valid (basically curly brackets in tags)
```
{{ variable | prepend: '{' | append: '}' }}
```
and
```
{{ 'blah {{ yy }}' | replace: '{{', 'xx' }}
```

shopmike:
I'm also starting to wonder if we should solve the few following issues as these can be fixed with a few more tokens to scan for

### Allow escaping in " so that it is possible to use both a single quote and double quote in a string, which is not currently possible
```
{% assign = "we can handle \"double quotes\" and 'single quotes'" %}
```
> This **does not break backwards support** as double quotes are not currently possible inside double quotes.

### The new liquid tag splits on newlines, this means strings will no longer be able to express new lines in the new liquid tag
```
{% liquid 
    assign = "we can handle\n new lines" 
%}
```

> This **can possibly break templates** that use "\n" in a string today. The liquid.format() command could be a way to migrate templates without issues.

### Carriage returns and tabs as these are also common, not as much with the web but windows and Tab Seperated Values
```
{% assign = "we can handle returns\r and tabs \t" %}
```
> This **can possibly break templates** that use "\r" or "\t" in a string today. The liquid.format() command could be a way to migrate templates without issues.

### Finally because we support the above we have to handle escaping escapes
```
{% assign = "we can handle a real backslash next to the character n \\n we are not on a newline" %}
```
> This **can possibly break templates** that use "\\" before a control character in a string today. The liquid.format() command could be a way to migrate templates without issues.

shopmike:
An alternative proposal for the changes above is to bring in a third quoting method using backticks

### Allow escaping in ` so that it is possible to use both a single quote and double quote, and back ticks in a string, which is not currently possible
```
{% assign = `we can handle "double quotes" and 'single quotes' and \`back ticks\`` %}
```
> This **does not break backwards support** for the tokenizer as double quotes are not currently possible inside double-quotes. However **string detection in other areas will need to be updated**

### The new liquid tag splits on newlines, this means strings will no longer be able to express new lines in the new liquid tag
```
{% liquid 
    assign = `we can handle\n new lines`
%}
```

> This **does not break backwards support** for the tokenizer as backticks are new functionality. However **string detection in other areas will need to be updated**

### Carriage returns and tabs as these are also common, not as much with the web but windows and Tab Separated Values
```
{% assign = `we can handle returns\r and tabs \t` %}
```
> This **does not break backwards support** for the tokenizer as backticks are new functionality. However **string detection in other areas will need to be updated**

### Finally because we support the above we have to handle escaping escapes
```
{% assign = `we can handle a real backslash next to the character n \\n we are not on a newline` %}
```
> This **does not break backwards support** for the tokenizer as backticks are new functionality. However **string detection in other areas will need to be updated**

devdreamsolution (approved):

Darkangel956 (approved):

PR #1180, opened 2019-10-04
Test all filters against random data to detect exceptions

shopmike:
Built this test out of curiosity. The good news is it passes. 💯 

### DIFF
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -780,7 +780,6 @@ class StandardFiltersTest < Minitest::Test
       { 1 => "bar" },
       ["foo", 123, nil, true, false, Drop, ["foo"], { foo: "bar" }],
     ]
-    errors = []
     test_types.each do |first|
       test_types.each do |other|
         (@filters.methods - Object.methods).each do |method|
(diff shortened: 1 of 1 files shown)


test/integration/standard_filter_test.rb
```diff
@@ -754,6 +754,53 @@ def test_where_array_of_only_unindexable_values
...
+            @filters.send(method, *inputs)
+          rescue Liquid::ArgumentError => e
+            errors << "#{method} returned #{e.message}"
+          rescue Liquid::ZeroDivisionError => e
+            errors << "#{method} returned #{e.message}"
+          end
+        end
+      end
+    end
+    errors
```
Thibaut: This doesn't seem to be used for anything.

shopmike: I can pull these out, however, I might just use this to assert that it doesn't contain an incorrect parameter count error in any response.

shopmike: test my test

shopmike: Not needed as realized this is outside of Strainer so the exception is still thrown.

PR #1181, opened 2019-10-07
Add support for `as` in Render and Include tags

shopmike:
This is to support a new syntax that allows `with` and `for` to be aliased to something other than the template name.

## Include now supports `as`

Example
```twig
{% include 'product_alias' with products[0] as product %}
```
Usage inside `product_alias` would be 
```handlebars
{{ product.title }}
```

## Render now supports `with`, `for` and `as`

Render now supports `with` and `for` with a major difference being that `render` generates a new subcontext for each loop in `for`. This means each iteration of `for` cannot impact others, this will allow loops to be `async` in the future for performance improvements.

@Shopify/guardians-of-the-liquid @samdoiron @Thibaut

### DIFF
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -16,18 +16,20 @@ module Liquid
   #   {% include 'product' for products %}
   #
   class Include < Tag
-    Syntax = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?/o
+    SYNTAX = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSignature}+))?/o
+    Syntax = SYNTAX
 
     attr_reader :template_name_expr, :variable_name_expr, :attributes
 
     def initialize(tag_name, markup, options)
       super
 
-      if markup =~ Syntax
+      if markup =~ SYNTAX
 
         template_name = Regexp.last_match(1)
         variable_name = Regexp.last_match(3)
 
+        @alias_name = Regexp.last_match(5) || nil
         @variable_name_expr = variable_name ? Expression.parse(variable_name) : nil
         @template_name_expr = Expression.parse(template_name)
         @attributes = {}
@@ -54,7 +56,7 @@ module Liquid
         parse_context: parse_context
       )
 
-      context_variable_name = template_name.split('/').last
+      context_variable_name = @alias_name ? @alias_name : template_name.split('/').last
 
       variable = if @variable_name_expr
         context.evaluate(@variable_name_expr)
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -29,7 +29,7 @@ module Liquid
         template_name = Regexp.last_match(1)
         variable_name = Regexp.last_match(3)
 
-        @alias_name = Regexp.last_match(5) || nil
+        @alias_name = Regexp.last_match(5)
         @variable_name_expr = variable_name ? Expression.parse(variable_name) : nil
         @template_name_expr = Expression.parse(template_name)
         @attributes = {}
@@ -56,7 +56,7 @@ module Liquid
         parse_context: parse_context
       )
 
-      context_variable_name = @alias_name ? @alias_name : template_name.split('/').last
+      context_variable_name = @alias_name || template_name.split('/').last
 
       variable = if @variable_name_expr
         context.evaluate(@variable_name_expr)
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -16,7 +16,7 @@ module Liquid
   #   {% include 'product' for products %}
   #
   class Include < Tag
-    SYNTAX = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSignature}+))?/o
+    SYNTAX = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSegment}+))?/o
     Syntax = SYNTAX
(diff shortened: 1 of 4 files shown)


lib/liquid/tags/include.rb
```diff
@@ -16,18 +16,20 @@ module Liquid
...
     def initialize(tag_name, markup, options)
       super
 
-      if markup =~ Syntax
+      if markup =~ SYNTAX
 
         template_name = Regexp.last_match(1)
         variable_name = Regexp.last_match(3)
 
+        @alias_name = Regexp.last_match(5) || nil
```
Thibaut: The ` || nil` is unnecessary.

lib/liquid/tags/render.rb
```diff
@@ -38,13 +41,25 @@ def render_tag(context, output)
         parse_context: parse_context
       )
 
-      inner_context = context.new_isolated_subcontext
-      inner_context.template_name = template_name
-      inner_context.partial = true
-      @attributes.each do |key, value|
-        inner_context[key] = context.evaluate(value)
+      context_variable_name = @alias_name ? @alias_name : "this"
```
Thibaut: To ease the adoption of `render` over `include` I think we decided to keep support for the template name fallback.

lib/liquid/tags/include.rb
```diff
@@ -16,18 +16,20 @@ module Liquid
   #   {% include 'product' for products %}
   #
   class Include < Tag
-    Syntax = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?/o
+    SYNTAX = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSignature}+))?/o
```
Thibaut: Why `#{VariableSignature}+` and not just `#{VariableSignature}`? (if we expect only one)

lib/liquid/tags/render.rb
```diff
@@ -38,13 +41,25 @@ def render_tag(context, output)
...
-      inner_context.template_name = template_name
-      inner_context.partial = true
-      @attributes.each do |key, value|
-        inner_context[key] = context.evaluate(value)
+      context_variable_name = @alias_name || template_name.split('/').last
+
+      variable = if @variable_name_expr
+        context.evaluate(@variable_name_expr)
+      else
+        context.find_variable(template_name, raise_on_not_found: false)
```
Thibaut: Does this automatically pass a variable whose name is the same as the snippet name? I wouldn't keep that behavior. If there is no `with` or `for`, we should pass nothing down.

lib/liquid/tags/render.rb
```diff
@@ -38,13 +41,25 @@ def render_tag(context, output)
...
+      end
+
+      variable = [variable] unless variable.is_a?(Array)
```
Thibaut: Protip: `variable = Array(variable)`

Thibaut: That said we could also maybe turn the `.each` into a method, so that we can avoid making a one-item array and do something like instead:

```ruby
if variable_is_a?(Array) then variable.each(&:method_name) else method_name(variable)
```

shopmike: This doesn't seem to be fully equivalent `variable = Array(variable)`. I think its treating `nil` as `[]` however need it to be `[nil]` so it loops.

I was originally going down the method path but there are so many params in this scope that would need to pass in `context`, `template_name`, `output`, `var`, `partial`

lib/liquid/tags/render.rb
shopmike: I've got a change ready to set this to `nil`. It was ignoring this down further by checking if `with` was set

lib/liquid/tags/include.rb
shopmike: `VariableSignature` = `/\(?[\w\-\.\[\]]\)?/` only matches characters so we expect many

shopmike: I'm actually thinking now I should change this to `VariableSegment` = `/[\w\-]/`

lib/liquid/tags/render.rb
Thibaut: Good point. Let's add for that then.

shopmike: Not sure on the Rubyist view on this, but just used a lambda to group the functionality, gets around excessive params and needing to create a fake array.

lib/liquid/tags/include.rb
```diff
@@ -16,18 +16,20 @@ module Liquid
   #   {% include 'product' for products %}
   #
   class Include < Tag
-    Syntax = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?/o
+    SYNTAX = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSegment}+))?/o
```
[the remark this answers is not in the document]
shopmike: I think it should be the way it is, It's not a string it's forcing you to define the variable you want it to be. Matches up with how other tags are defined 
```django
{% render "product" product as product %}
{% for product in products %}
{% assign product = products %}
```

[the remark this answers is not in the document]
Thibaut: Hmm but then I would think it sets a variable in the parent context. What do you not like about: `{% render "product" with product as "product" %}`?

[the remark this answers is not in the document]
shopmike: it's just so out of place, it's the only instance where a variable name would be quoted. For me, anything inside quotes means you can have any character. In this instance, it is introducing a new concept.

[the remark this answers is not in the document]
shopmike: instead of the normal `"(.*)"` you have a similar but different `"([\w-]*)"`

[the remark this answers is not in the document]
Thibaut: Ah I see. Good point.

PR #1184, opened 2019-10-08
Resolve InputIterator dropping context

shopmike:
Fixes a bug where context isn't available in a nested Drop when using filters that leverage InputIterator. Issue is reproducible with

```ruby
  def test_context_drop_array_with_map
    output = Liquid::Template.parse(' {{ contexts | map: "bar" }} ').render!('contexts' => [ContextDrop.new, ContextDrop.new], 'bar' => "carrot")
    assert_equal(' carrotcarrot ', output)
  end
```

This works by using the `@context` available on the Strainer and passing it to InputIterator to then apply to any Drops it encounters.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -432,9 +432,7 @@ module Liquid
 
     private
 
-    def context
-      @context ||= nil
-    end
+    attr_reader :context
 
     def raise_property_error(property)
       raise Liquid::ArgumentError, "cannot select the property '#{property}'"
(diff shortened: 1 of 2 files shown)


shopmike:
@Thibaut @pushrax @carolineschnapp

lib/liquid/standardfilters.rb
```diff
@@ -432,6 +432,10 @@ def default(input, default_value = '')
 
     private
 
+    def context
```
Thibaut: We could simply add `attr_reader :context`.

`||= nil` is unnecessary (the method could simply return `@context`). When the variable is `nil` (or `false`), `||=` will keep re-evaluating/re-assigning the value. To avoid that we'd need to use `defined?`.

shopmike: Thanks.

PR #1185, opened 2019-10-08
Fixes #1156 Liquid::Utils has no tests

### LINKED ISSUE #1156: Liquid::Utils has no tests
There are many filters relying on Liquid::Utils methods such as `to_number`, `to_date` etc however there are no tests covering these methods and their expected behavior.

alebruck:
Fixes #1156

Added tests to Liquid::Utils, tried to understand the context and constraints of each one of the methods based on the use cases.

Didn't add a test for `#slice_collection` as it seems to be a workaround created for `ForTagTest`.

I've also added a `skip` for `test_to_date_with_string` as it seems like minitest is failing to stub the timezone offset, I'm investigating it further on minitest.

### DIFF
--- a/test/unit/utils_unit_test.rb
+++ b/test/unit/utils_unit_test.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class UtilsUnitTest < Minitest::Test
+  include Liquid
+
+  def test_to_number_with_float
+    assert_equal(BigDecimal("0.1"), Utils.to_number(0.1))
+  end
+
+  def test_to_number_with_numeric
+    assert_equal(1, Utils.to_number(1))
+  end
+
+  def test_to_number_with_string
+    assert_equal(123, Utils.to_number("123"))
+  end
+
+  def test_to_number_with_precision_string
+    assert_equal(BigDecimal("0.1"), Utils.to_number("0.1"))
+  end
+end
--- a/test/unit/utils_unit_test.rb
+++ b/test/unit/utils_unit_test.rb
@@ -34,4 +34,29 @@ class UtilsUnitTest < Minitest::Test
       Utils.to_integer("FOOBAR")
     end
   end
+
+  def test_to_date_with_time
+    t = Time.new(2019, 01, 01, 12, 00, 00, "+00:00")
+    assert_equal(t, Utils.to_date(t))
+  end
+
+  def test_to_date_with_now
+    t = Time.new(2019, 01, 01, 12, 00, 00, "+00:00")
+    Time.stub(:now, t) do
+      assert_equal(t, Utils.to_date("now"))
+    end
+  end
+
+  def test_to_date_with_timestamp
+    t = Time.new(2019, 01, 01, 12, 00, 00, "+00:00")
+    assert_equal(t, Utils.to_date(t.to_i))
+  end
+
+  def test_to_date_with_string
+    skip "minitest mock is failing to stub offset"
+    t = Time.new(2019, 01, 01, 12, 00, 00, "-02:00")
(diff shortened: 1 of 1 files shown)


alebruck:
CLA done

test/unit/utils_unit_test.rb
```diff
@@ -0,0 +1,86 @@
...
+    assert_equal(t, Utils.to_date(t.to_i))
+  end
+
+  def test_to_date_with_string
+    skip "minitest mock is failing to stub offset"
+    t = Time.new(2019, 01, 01, 12, 00, 00, "-02:00")
+    Time.stub(:now, t) do
+      assert_equal("2019-01-01 16:00:00 -0200", Utils.to_date("16:00"))
+    end
+  end
```
shopmike: What happens when an invalid string is passed?

test/unit/utils_unit_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  def test_slice_collection_using_each_with_array
+    assert_equal([0, 1, 2, 3], Utils.slice_collection_using_each([0, 1, 2, 3, 4, 5], 0, 4))
+  end
+
+  def test_slice_collection_using_each_with_no_range
+    assert_equal([], Utils.slice_collection_using_each([0, 1, 2, 3, 4, 5], 0, 0))
+  end
+
+  def test_slice_collection_using_each_with_middle_range
+    assert_equal([2], Utils.slice_collection_using_each([0, 1, 2, 3, 4, 5], 2, 3))
```
shopmike: How does this handle negative inputs in the 2nd and 3rd parameters? Or if the second is larger than the third.

test/unit/utils_unit_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def test_to_number_with_string
+    assert_equal(123, Utils.to_number("123"))
+  end
+
+  def test_to_number_with_precision_string
+    assert_equal(BigDecimal("0.1"), Utils.to_number("0.1"))
+  end
+
```
shopmike: Can we try something like a string starting and/or ending in digits but also having letters? Also maybe exponential format to see if it can handle it.

test/unit/utils_unit_test.rb
```diff
@@ -0,0 +1,86 @@
...
+  def test_to_date_with_now
+    t = Time.new(2019, 01, 01, 12, 00, 00, "+00:00")
+    Time.stub(:now, t) do
+      assert_equal(t, Utils.to_date("now"))
+    end
+  end
+
+  def test_to_date_with_timestamp
+    t = Time.new(2019, 01, 01, 12, 00, 00, "+00:00")
+    assert_equal(t, Utils.to_date(t.to_i))
```
shopmike: I think I remember seeing an issue where someone was trying to pass in the year but it was treating it as a timestamp. This would also be a good example

alebruck:
@shopmike What's your opinion on adding https://github.com/travisjeffery/timecop to the test dependencies? I'm struggling to test time methods that depend on current time/zone and `timecop`would make it easier.

shopmike:
Yeah, I think it is worth trying and seeing what it looks like. Just make sure it's only loaded for tests, so it can't impact Liquid users.

test/unit/utils_unit_test.rb
alebruck: This was solved on `test_slice_collection_using_each_with_negative_start` :)

test/unit/utils_unit_test.rb
alebruck: solved on `test_to_date_with_invalid_string`

devdreamsolution (approved):

PR #1186, opened 2019-10-09
Fix flaky tests

uchoudh:
This PR is intended to fix some flaky test behaviour introduced in this [PR](https://github.com/Shopify/liquid/pull/1172)

The purpose of the tests is to make sure that these new added attributes give some time greater than zero and in the event they are removed the tests should be able to catch this. Hence, the time comparisons which can be flaky sometimes are removed for a more simpler check.


shopmike:
Sorry, half wrote a review but didn't submit correctly.

PR #1187, opened 2019-10-10
[DO NOT MERGE] Introduce a liquid template factory + the ability to restrict paths in the include tag

mmorissette:
This is only meant to be an exploration and will not be merged.

1. Create liquid templates for partials using a `TemplateFactory` that can be injected via the liquid context
2. Add the ability for `Liquid::Template` to restrict the liquid context
3. Add the ability to configure `Liquid::Include` to support specific paths

### DIFF
--- a/lib/liquid/partial_cache.rb
+++ b/lib/liquid/partial_cache.rb
@@ -11,7 +11,10 @@ module Liquid
       source = file_system.read_template_file(template_name)
       parse_context.partial = true
 
-      partial = Liquid::Template.parse(source, parse_context)
+      liquid_template_factory = (context.registers[:liquid_template_factory] ||= Liquid::TemplateFactory)
+      liquid_template = liquid_template_factory.for(template_name)
+
+      partial = liquid_template.parse(source, parse_context)
       cached_partials[template_name] = partial
     ensure
       parse_context.partial = false
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -21,6 +21,23 @@ module Liquid
 
     attr_reader :template_name_expr, :variable_name_expr, :attributes
 
+    class << self
+      def allow_path(matcher)
+        allowed_paths.push(matcher)
+      end
+
+      def allowed_paths
+        @allowed_paths ||= []
+      end
+
+      def allowed_path?(path)
+        return true if allowed_paths.empty?
+        allowed_paths.any? do |matcher|
+          matcher.match?(path)
+        end
+      end
+    end
+
     def initialize(tag_name, markup, options)
       super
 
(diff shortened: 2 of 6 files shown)


lib/liquid/partial_cache.rb
```diff
@@ -11,7 +11,10 @@ def self.load(template_name, context:, parse_context:)
       source = file_system.read_template_file(template_name)
       parse_context.partial = true
 
-      partial = Liquid::Template.parse(source, parse_context)
+      liquid_template_factory = (context.registers[:liquid_template_factory] ||= Liquid::TemplateFactory)
```
shopmike: You may need to modify this if this moves forward. `context.registers` are dropped when `context` is a `StaticRegister` and you enter a subcontext eg. `render` tags.

If you want this to flow down through subcontexts you will need to use some form of `context.registers.static[:liquid_template_factory]` however `static` only exists on `StaticRegisters`

lib/liquid/tags/include.rb
```diff
@@ -49,6 +66,7 @@ def parse(_tokens)
     def render_to_output_buffer(context, output)
       template_name = context.evaluate(@template_name_expr)
       raise ArgumentError, options[:locale].t("errors.argument.include") unless template_name
+      raise ArgumentError unless allowed_path?(template_name)
```
shopmike: I don't think this should be the responsibility of the tag, instead, it should be the responsibility of the filesystem to reject access. Another way to accomplish this could be to pass through the caller in this case `include` through to `PartialCache.load` likely though either `context` or `parse_context` and all this logic can live inside the FileSystem and implement its own authentication.

lib/liquid/tags/render.rb
```diff
@@ -44,14 +44,16 @@ def render_tag(context, output)
       context_variable_name = @alias_name || template_name.split('/').last
 
       render_partial_func = ->(var) {
-        inner_context = context.new_isolated_subcontext
-        inner_context.template_name = template_name
-        inner_context.partial = true
-        @attributes.each do |key, value|
-          inner_context[key] = context.evaluate(value)
+        partial.with_context_restrictions(context) do
```
shopmike: I feel like this is the role of `new_isolated_subcontext`. This is just a further restriction and think It should be handled in there.

lib/liquid/template_factory.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Liquid
+  class TemplateFactory
```
shopmike: I think the only issue with this approach is you need to build a whole new `Template` which means needing to register all the basic tags again. Would an approach similar to how context creates a restricted version of itself with `context.new_isolated_subcontext` be used here with a `template.new_isolated_subtemplate`

lib/liquid/tags/include.rb
shopmike: For example, the base user could be  
```
context.registers[`current_user`] = 'system'
```
and then when you enter a scope that you want to restrict
```
context.registers[`current_user`] = 'restricted_user'
```
This is all obviously a very basic example, but then inside the `File System` you can do a check if the current user/callee has access to the files you are trying to access, if not throw an exception up.

lib/liquid/template_factory.rb
shopmike: After seeing more of the plan to implement this, I think this can likely stay similar.

PR #1188, opened 2019-10-11
Shift all Tag parsing into parse method

shopmike:
This is a clean up that accomplishes a few things. It doesn't add any new methods, it just shifts parsing logic inside `initialize` to `parse` which is pre-existing functionality.

## Tag initialization is now lightweight and universal

There are no tags that need an `initialize` method. By removing this it opens up the ability to dynamically create Tags outside of the normal parsing process. There is a current overlap of functionality between some tags and this moves towards the path of possibly allowing `Tag A` to wrap itself in `Tag B`.

**Example:** `render` wrapping itself with `for`

## Separation between Object creation and Parsing

By moving away from the `initialize` method, it will be easier to move tag parsing to `liquid-c`

@pushrax @Thibaut @Shopify/guardians-of-the-liquid @Shopify/liquid

### DIFF
--- a/lib/liquid/block.rb
+++ b/lib/liquid/block.rb
@@ -7,6 +7,7 @@ module Liquid
     def initialize(tag_name, markup, options)
       super
       @blank = true
+      @body = nil
     end
 
     def parse(tokens)
@@ -17,7 +18,7 @@ module Liquid
 
     # For backwards compatibility
     def render(context)
-      @body.render(context)
+      @body&.render(context)
     end
 
     def blank?
--- a/lib/liquid/tag.rb
+++ b/lib/liquid/tag.rb
@@ -17,8 +17,6 @@ module Liquid
         disabled_tags.push(*tags)
       end
 
-      private :new
-
       def disabled_tags
         @disabled_tags ||= []
       end
--- a/lib/liquid/tag.rb
+++ b/lib/liquid/tag.rb
@@ -2,7 +2,7 @@
 
 module Liquid
   class Tag
-    attr_reader :nodelist, :tag_name, :line_number, :parse_context
+    attr_reader :nodelist, :tag_name, :line_number, :parse_context, :markup
     alias_method :options, :parse_context
     include ParserSwitching
 
--- a/lib/liquid/tags/assign.rb
+++ b/lib/liquid/tags/assign.rb
@@ -19,11 +19,11 @@ module Liquid
     attr_reader :to, :from
 
     def parse(_tokens)
-      if @markup =~ Syntax
+      if markup =~ Syntax
         @to = Regexp.last_match(1)
-        @from = Variable.new(Regexp.last_match(2), @parse_context)
+        @from = Variable.new(Regexp.last_match(2), options)
       else
-        raise SyntaxError, @parse_context[:locale].t(self.class.syntax_error_translation_key)
+        raise SyntaxError, options[:locale].t(self.class.syntax_error_translation_key)
       end
     end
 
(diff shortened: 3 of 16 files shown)


lib/liquid/block.rb
```diff
@@ -7,6 +7,7 @@ class Block < Tag
     def initialize(tag_name, markup, options)
       super
       @blank = true
+      @body = nil
```
Thibaut: Why not `BlockBody.new`? Would avoid the `&.` below.

lib/liquid/tag.rb
```diff
@@ -17,8 +17,6 @@ def disable_tags(*tags)
         disabled_tags.push(*tags)
       end
 
-      private :new
```
Thibaut: Do you know what was the intent behind making `new` private?

lib/liquid/tags/assign.rb
```diff
@@ -18,13 +18,12 @@ def self.syntax_error_translation_key
...
     attr_reader :to, :from
 
-    def initialize(tag_name, markup, options)
-      super
-      if markup =~ Syntax
+    def parse(_tokens)
+      if @markup =~ Syntax
         @to = Regexp.last_match(1)
-        @from = Variable.new(Regexp.last_match(2), options)
+        @from = Variable.new(Regexp.last_match(2), @parse_context)
```
Thibaut: Nit: we could use the accessors (`parse_context`)

lib/liquid/block.rb
shopmike: I know it would save the `&.` below but it also initializes the class when it might not be needed. I'm not sure which one is worse.

lib/liquid/tag.rb
shopmike: I can have a look back through, but I'm assuming this is to avoid creating tags outside of Liquid.

lib/liquid/block.rb
Thibaut: Right. I'm OK to keep it as is.

lib/liquid/tags/assign.rb
shopmike: I'm not sure why I have done this as `options` is also available.

shopmike: I think I discovered all these methods are accessors anyway so most don't even need to be modified.

shopmike: OK, I've fixed these all up, there are very minimal changes now.

lib/liquid/tags/case.rb
```diff
@@ -7,18 +7,15 @@ class Case < Block
 
     attr_reader :blocks, :left
 
-    def initialize(tag_name, markup, options)
-      super
+    def parse(tokens)
```
Thibaut: Why not call `super` in here as in `Capture`? (same below)

I'm also not a fan of `super`. It couples base class and subclasses very tightly. We could address this by splitting the method in two (one for the base class, which calls another method implemented by the subclasses that doesn't need to call `super`).

shopmike:
Yeah, it just makes it easier to replace the method. This is just a tiny step to improve things. I've been playing around a bit with replacing `Tag` and `Block` with a new `Node` that has `properties` and `children` and handles all the logic it does now.

shopmike:
Just a comment here to remember, this will likely mess with Tags that extend the core tags.

lib/liquid/tags/case.rb
shopmike: I'm not either, I've been playing a bit on how to clean this up to not need it, really these super calls are just telling it to handle the block content which should just be done anyway.

devdreamsolution (approved):

PR #1189, opened 2019-10-11
[Exploration] Extract partial creation to a method in the render tag

mmorissette:
Extract the creation of a partial in the `Render` tag to its own method so that it can be extended.

### DIFF
--- a/lib/liquid/tags/render.rb
+++ b/lib/liquid/tags/render.rb
@@ -35,15 +35,10 @@ module Liquid
       template_name = context.evaluate(@template_name_expr)
       raise ArgumentError, options[:locale].t("errors.argument.include") unless template_name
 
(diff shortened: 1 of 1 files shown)


lib/liquid/tags/render.rb
```diff
@@ -35,15 +35,10 @@ def render_tag(context, output)
...
-      partial = PartialCache.load(
-        template_name,
-        context: context,
-        parse_context: parse_context
-      )
-
       context_variable_name = @alias_name || template_name.split('/').last
 
       render_partial_func = ->(var) {
+        partial = load_partial(template_name, context)
```
shopmike: I don't think this needs to go inside `render_partial_func`, when using the `for` syntax it will run this for every iteration.

Issue #1169, opened 2019-09-19
Fix equal alignment in ruby styling


shopmike:
Resolves issues in https://github.com/Shopify/liquid/pull/1159

PR #1190, opened 2019-10-12
Add ForceEqualSignAlignment to .rubocop.yml

alebruck:
Fixes https://github.com/Shopify/liquid/issues/1169 by adding cop `ForceEqualSignAlignment` and updating all reported issues.

For now, I had to disable `SpaceAroundOperators` as this cop seems to have an issue that causes an infinite loop(not possible to achieve a green state). I reported it on rubocop https://github.com/rubocop-hq/rubocop/issues/7422 and I can enable it again once this issue is resolved.


shopmike:
@ashmaroli you raised this originally, so will await feedback

.rubocop-https---shopify-github-io-ruby-style-guide-rubocop-yml
```diff
@@ -892,7 +892,7 @@ Lint/FormatParameterMismatch:
   Enabled: true
 
 Lint/HandleExceptions:
-  Enabled: true
+  AllowComments: true
```
ashmaroli: I don't think it's a good idea to edit this file since it is auto-generated using upstream data.
If this change is absolutely necessary for this PR to pass, the better alternative would be to redefine the setting in the local `.rubocop.yml`

lib/liquid/block_body.rb
```diff
@@ -34,15 +34,15 @@ def parse(tokenizer, parse_context, &block)
...
+          tag_name    = Regexp.last_match(1)
+          markup      = Regexp.last_match(2)
           unless (tag = registered_tags[tag_name])
             # end parsing if we reach an unknown tag and let the caller decide
             # determine how to proceed
             return yield tag_name, markup
           end
-          new_tag = tag.parse(tag_name, markup, tokenizer, parse_context)
-          @blank &&= new_tag.blank?
+          new_tag     = tag.parse(tag_name, markup, tokenizer, parse_context)
```
ashmaroli: Does RuboCop force this realignment if you were to revert manually and then autocorrect after?
The reason I ask is the benefits of operator-alignment is lost when there is a block of code separating two realigned lines..

lib/liquid/block_body.rb
```diff
@@ -92,7 +92,7 @@ def parse(tokenizer, parse_context, &block)
           end
           parse_context.trim_whitespace = false
           @nodelist << token
-          @blank &&= !!(token =~ WhitespaceOrNothing)
+          @blank                      &&= !!(token =~ WhitespaceOrNothing)
```
ashmaroli: This is the side-effect of using automated corrections.. :disappointed: :
As a human, you can tell from a distance that this change significantly hampers readability instead of improving it..
(View the entire file for proper judgement).

lib/liquid/context.rb
```diff
@@ -85,9 +85,9 @@ def pop_interrupt
     end
 
     def handle_error(e, line_number = nil)
-      e = internal_error unless e.is_a?(Liquid::Error)
+      e                 = internal_error unless e.is_a?(Liquid::Error)
```
ashmaroli: It would be better if you could replace single letter variables with *something longer yet apt* to mitigate this excess whitespace..

If its not too much trouble, I recommend that you follow this suggestion for all similar scenarios

lib/liquid/block_body.rb
shopmike: I think the only thing that breaks the behaviour is an empty line from when I was playing around with this. In cases where the formatting isn't ideal, you need to space it.

lib/liquid/block_body.rb
shopmike: This is a good way to be able to find where needing to space should be done. Searching for 15+ spaces provides most of the weird auto-corrections in which case I would space

lib/liquid/context.rb
shopmike: I think the only issue with most of these examples is that by making the variable longer, the next line is usually the same variable but accessing long method names or nested methods.

.rubocop-https---shopify-github-io-ruby-style-guide-rubocop-yml
ashmaroli: I see.. Then you (@shopmike) should push this change to `master` directly then. It may not be ideal but a lot better than pulling the change out from multiple PRs..

Shopify could also consider shipping the Official Style Guide as a Ruby gem so that its updates do not explicitly show up in downstream repos.

shopmike: Yeah, the plan is to push this through as its own PR, just hasn't been a priority.

The config rarely changes and when it does I don't mind the fact it forces everyone to follow it as generally it would only be done when enough people have agreed that it applies to all projects.

alebruck:
> Now that I see the extent of the _damage to readability_ this can cause, I'm not so inclined on having this merged in the current state.
> IMHO, the better alternative would be to **not enforcing EqualSignAlignment globally** but rather have the best of both worlds by aligning where it improves readability and forgoing the alignment if it hampers readability..

As now we have a reference on all equal alignments in the code base, should we revert the ForceEqualSignAlignment cop and leave all the changes that make sense(reverting cases that don't bring much benefit)?

lib/liquid/context.rb
alebruck: should we keep this cope and try to address these cases individually by small refactoring to improve readability or should we postpone the enforcement of this cop?

ashmaroli:
> should we revert the ForceEqualSignAlignment cop and leave all the changes that make sense

@alebruck I would prefer that if you don't mind the extra effort. Either ways, I'm thankful that you attempted to resolve this issue.

alebruck:
@ashmaroli Could you please check this one again? No new cops but the alignment improvements required for the cop are already here. I would leave any refactoring related to var name in a different PR as it looks like we could use more refactoring than just renaming.

ashmaroli:
@alebruck I had sent you a PR on your fork: https://github.com/alebruck/liquid/pull/1
Did you not see that?

alebruck:
@ashmaroli oh, somehow I got no notification from that, sorry about that 😅

ashmaroli:
@shopmike My concerns have been addressed.
Requesting a re-review and a *squash-merge*..

test/unit/tokenizer_unit_test.rb
```diff
@@ -34,15 +34,15 @@ def test_calculate_line_numbers_per_token_with_profiling
 
   def tokenize(source)
     tokenizer = Liquid::Tokenizer.new(source)
-    tokens = []
-    while (t = tokenizer.shift)
+    tokens    = []
+    while (t  = tokenizer.shift)
```
shopmike: This is about the only one that doesn't seem to make sense.

ashmaroli: Ah! I missed that! :stuck_out_tongue: 
It should be something like:
```ruby
    tokenizer = Liquid::Tokenizer.new(source)
    tokens    = []

    while (t = tokenizer.shift)
```
I can't push to this branch directly,.. so its your call..

shopmike:
Awesome thanks for resolving this outstanding issue @ashmaroli @alebruck . Also congrats on your first contribution to Liquid @alebruck 🎉

PR #1191, opened 2019-10-14
[New Feature] Add forloop inside render tag when using for syntax

shopmike:
This populates `forloop` inside `render` when using the syntax 
```twig
{% render 'something' with products as product %}
```
@Thibaut @samdoiron @Shopify/guardians-of-the-liquid

### DIFF
--- a/lib/liquid/tags/render.rb
+++ b/lib/liquid/tags/render.rb
@@ -43,19 +43,26 @@ module Liquid
 
       context_variable_name = @alias_name || template_name.split('/').last
 
-      render_partial_func = ->(var) {
+      render_partial_func = ->(var, forloop) {
         inner_context = context.new_isolated_subcontext
         inner_context.template_name = template_name
         inner_context.partial = true
+        inner_context['forloop'] = forloop if forloop
         @attributes.each do |key, value|
           inner_context[key] = context.evaluate(value)
         end
         inner_context[context_variable_name] = var unless var.nil?
         partial.render_to_output_buffer(inner_context, output)
+        forloop&.send(:increment!) if forloop
       }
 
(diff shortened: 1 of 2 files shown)


lib/liquid/tags/render.rb
```diff
@@ -43,19 +43,26 @@ def render_tag(context, output)
...
         inner_context = context.new_isolated_subcontext
         inner_context.template_name = template_name
         inner_context.partial = true
+        inner_context['forloop'] = forloop if forloop
         @attributes.each do |key, value|
           inner_context[key] = context.evaluate(value)
         end
         inner_context[context_variable_name] = var unless var.nil?
         partial.render_to_output_buffer(inner_context, output)
+        forloop&.send(:increment!) if forloop
```
tobi: feels weird to reach into the object and call a private method. Let's change the forloop drop signature instead. 

Also isn't forloop& with a if forloop at the end redundant?

shopmike: I think its reaching in because its a drop and not wanting to expose increment inside Liquid.

Also yes `if forloop` is redundant here, I'll fix it up

lib/liquid/tags/render.rb
```diff
@@ -43,19 +43,26 @@ def render_tag(context, output)
 
       context_variable_name = @alias_name || template_name.split('/').last
 
-      render_partial_func = ->(var) {
+      render_partial_func = ->(var, forloop) {
```
Thibaut: Nit: I think the `forloop` argument is unnecessary. Since it's defined in the outer scope, it'll be available in the proc.

shopmike: Doesn't seem to like it considering it's nested inside the if statement and moving it up would mean putting the same guard around it anyway.

PR #1193, opened 2019-10-22
Allow render to handle with and for correctly

shopmike:
Currently `render` can only handle rendering `Array` and not `Drop` which is the most common object type in Shopify.

The basic solution was to enable `variable.respond_to?(:each)` however, this leads to maps being passed to `with` to also loop.

This PR means `with` will never loop, and `for` will loop when the object implements `each` and `count`

### DIFF
--- a/lib/liquid/tags/render.rb
+++ b/lib/liquid/tags/render.rb
@@ -2,7 +2,7 @@
 
 module Liquid
   class Render < Tag
-    SYNTAX = /(#{QuotedString}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSegment}+))?/o
+    SYNTAX = /(#{QuotedString}+)(\s+(with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSegment}+))?/o
 
     disable_tags "include"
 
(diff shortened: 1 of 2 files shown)


lib/liquid/tags/render.rb
```diff
@@ -58,7 +60,7 @@ def render_tag(context, output)
       }
 
       variable = @variable_name_expr ? context.evaluate(@variable_name_expr) : nil
-      if variable.is_a?(Array)
+      if variable.respond_to?(:each) && variable.respond_to?(:count) && @loop_variable
```
Thibaut: `count` is part of `Enumerable` so Hash-like drops would still render incorrectly.

How about `variable.respond_to?(:each) && !variable.respond_to?(:keys)`?

Also we could move the `@loop_variable` check to the front since it's a fast check.

shopmike: In this implementation, It doesn't mind about `Hash` or `Hash-like Drop` as the idea is that the `for` has been explicitly set. 

`variable.respond_to?(:each) && !variable.respond_to?(:keys)` could be a variation to this approach and means `with` and `for` could stay the same

Thibaut: Ah of course. In that case this is fine then. The developer's intent was to iterate, so even if the argument is `Hash`-like, iterate we should.

lib/liquid/tags/render.rb
```diff
@@ -14,11 +14,13 @@ def initialize(tag_name, markup, options)
...
       template_name = Regexp.last_match(1)
-      variable_name = Regexp.last_match(3)
+      with_or_for = Regexp.last_match(3)
+      variable_name = Regexp.last_match(4)
 
-      @alias_name = Regexp.last_match(5)
+      @alias_name = Regexp.last_match(6)
       @variable_name_expr = variable_name ? Expression.parse(variable_name) : nil
       @template_name_expr = Expression.parse(template_name)
+      @loop_variable = (with_or_for == "for")
```
Thibaut: Nit: `@for`? Or maybe we could have `@for_variable` and `@with_variable`?

lib/liquid/tags/render.rb
shopmike: :+1 on the `if` shortcutting, will change now

PR #1197, opened 2019-11-05
Fix post-trim for the raw tag

XrXr:
The added test did not pass before this commit.

### DIFF
--- a/lib/liquid/tags/raw.rb
+++ b/lib/liquid/tags/raw.rb
@@ -16,7 +16,10 @@ module Liquid
       while (token = tokens.shift)
         if token =~ FullTokenPossiblyInvalid
           @body << Regexp.last_match(1) if Regexp.last_match(1) != ""
-          return if block_delimiter == Regexp.last_match(2)
+          if block_delimiter == Regexp.last_match(2)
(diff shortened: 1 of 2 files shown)


XrXr:
I also have a more general solution that tries to solve this for anything that implements a custom `parse`. I almost have it working, but it didn't seem worth it to pursue more as it introduces a lot of code.

shopmike:
Whitespace control was the first time I had touched the Liquid repo and now in hindsight, I feel like it should have been part of the tokenizer. By changing the tokenizer it wouldn't need to look backwards and forwards as it does right now.

**We currently have this**
```
  TemplateParser              = /(#{PartialTemplateParser}|#{AnyStartingTag})/om
```

By having it also consume whitespace around it would keep all the whitespace attached to the token that controls whether it is displayed or not.

```
  TemplateParser              = /(/s*#{PartialTemplateParser}|#{AnyStartingTag}/s*)/om
```

@pushrax thoughts?

shopmike:
This change would also impact the `format` command which is why I'm suggesting a bigger change. I've pretty much had to store a `left` and `right` of whitespace on tokens anyway for it to be able to rebuild a template from the AST.

XrXr:
Closing as this has rotted and doesn't seem to be going anywhere.

PR #1206, opened 2019-12-12
Improve error message of `MemoryLimit` error.

tgxworld:
The existing error message is too generic since there are three types of limits in place.
It is useful to know which limit was reached to make it easier to debug the error.

### DIFF
--- a/lib/liquid/errors.rb
+++ b/lib/liquid/errors.rb
@@ -23,11 +23,14 @@ module Liquid
 
     def message_prefix
       str = +""
-      str << if is_a?(SyntaxError)
-        "Liquid syntax error"
-      else
-        "Liquid error"
-      end
+      str <<
+        if is_a?(SyntaxError)
+          "Liquid syntax error"
+        elsif is_a?(MemoryError)
+          "Liquid memory limit error"
+        else
+          "Liquid error"
+        end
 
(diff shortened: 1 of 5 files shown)


tobi:
That’s a good change but the error messages need to be user facing. Currently they just reveal an implementation detail. 

What’s an assign score?  I think we should reveal the why behind the error. Render runtime too high, render complexity too high, assign calls exhausted memory. Or errors like this. It would be a bit more actionable.

lib/liquid/errors.rb
```diff
@@ -40,14 +43,19 @@ def message_prefix
     end
   end
 
+  class MemoryError < Error
+    RENDER_LENGTH_ERROR_MESSAGE = 'Render length limits exceeded.'
+    RENDER_SCORE_ERROR_MESSAGE = 'Render score limits exceeded.'
+    ASSIGN_SCORE_ERROR_MESSAGE = 'Assign score limits exceeded.'
```
pushrax: Maybe just wording like:

"too many bytes rendered"
"too many tags rendered"
"too many bytes assigned to variables"

tgxworld:
@tobi @pushrax Thank you for the review. I've updated the PR to include the suggestions.

PR #1207, opened 2019-12-12
Use String literal instead

celsodantas:
# Why?

Simple code clean up.

Just moving a class method to a private constant.

### DIFF
--- a/lib/liquid/tags/assign.rb
+++ b/lib/liquid/tags/assign.rb
@@ -12,10 +12,6 @@ module Liquid
   class Assign < Tag
     Syntax = /(#{VariableSignature}+)\s*=\s*(.*)\s*/om
 
-    def self.syntax_error_translation_key
-      "errors.syntax.assign"
-    end
-
     attr_reader :to, :from
 
     def initialize(tag_name, markup, options)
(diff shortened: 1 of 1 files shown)


celsodantas:
> we might as well just pass the string literal to t at this point.

We could. Some people prefer no string literals but everything in CONSTs for readability. That's what I picked here.

If you feel strong about moving to string literal, I can do that. I'm easy here.

> There must have been some previous reason for this that's not needed now...

Can you elaborate on that? What is no longer needed here?

pushrax:
> Can you elaborate on that? What is no longer needed here?

It turns out that I added this method because I needed it for something else in 8d1cd4145335d9ee0c7cc0b99da0143c8334bba2 but then deleted the other use of `syntax_error_translation_key` before it was merged. The PR was merged before I finished self review..

> Some people prefer no string literals but everything in CONSTs for readability. That's what I picked here.

If the string is something non-descriptive like `"c"` or `"f1d2d2f924e986ac86fdf7b36c94bcdf32beec15"` I agree with this, but IMO for self descriptive string literals it's much easier to understand code that doesn't address them indirectly behind a constant.

ashmaroli:
@celsodantas I think you can amend the commit message as well since now there is no constant involved..

PR #1208, opened 2019-12-13
[Strainer] Separate factory from template

tjoyal:
The original `Strainer` grew to a point where it was really hard for someone unfamiliar (arguably even if you were) with the code to understand what was happening and propose future changes.

I didn't do here all of what I wanted as it was getting out of hand. I limited myself to mostly splitting the class in 2 so we can make future changes to it.

Api wise, `Liquid::Strainer.global_filter` has became `Liquid::StrainerFactory.add_global_filter`. Now I think we should :fire: this and prevent direct access of implementation details. `Liquid::Template.register_filter` is already exposed to do just that.

Notable followup changes I want to do:
- It is possible to corrupt global caches by calling `context.strainer.class.add_filter`. Adding a filter should most likely not be possible post class generation.
- There is a safe navigation on `@context&.strict_filters`. Most likely only for test purposes.
- `Liquid::StrainerFactory.add_global_filter` to not be invokable directly, use `Liquid::Template.register_filter` or some other centralized implementation.
- I'd like to explore as a followup not directly adding modules on top of one another, but rather have filters be classes that are sandboxes from one another with proper api defined in them (a filter shouldn't be allowed to call another filter private method) 

Small steps, heading forward.

Thoughts?


test/unit/strainer_template_unit_test.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  module PublicMethodOverrideFilter
+    def public_filter
+      "public"
+    end
+  end
+
+  def test_add_filter_does_not_raise_when_module_overrides_previously_registered_method
+    strainer = Context.new.strainer
+    with_global_filter do
```
tjoyal: This was a leaky test. `with_global_filter` silenced the problem for now.

lib/liquid/strainer_factory.rb
```diff
@@ -0,0 +1,36 @@
...
+        strainer_class_cache[filters] ||= begin
+          klass = Class.new(StrainerTemplate)
+          global_filters.each { |f| klass.add_filter(f) }
+          filters.each { |f| klass.add_filter(f) }
+          klass
+        end
+      end
+
+      def strainer_class_cache
+        @strainer_class_cache_var ||= {}
```
davidcornu: nit - any reason for the `_var`?

lib/liquid/strainer_template.rb
```diff
@@ -3,54 +3,41 @@
...
+  class StrainerTemplate
     def initialize(context)
       @context = context
     end
 
     class << self
-      attr_reader :filter_methods
-    end
+      def add_filter(filter)
+        raise ArgumentError, "Expected module but got: #{filter.class}" unless filter.is_a?(Module)
```
davidcornu: We should probably add a test for this because

```
irb(main):007:0> Class.new.is_a?(Module)
=> true
```

This should probably be `filter.instance_of?(Module)` because 

```
irb(main):011:0> Class.ancestors
=> [Class, Module, Object, Kernel, BasicObject]
```

test/test_helper.rb
```diff
@@ -55,19 +55,21 @@ def assert_match_syntax_error(match, template, assigns = {})
     end
 
     def with_global_filter(*globals)
-      original_global_strainer = Liquid::Strainer.class_variable_get(:@@global_strainer)
-      Liquid::Strainer.class_variable_set(:@@global_strainer, Class.new(Liquid::Strainer) do
-        @filter_methods = Set.new
-      end)
-      Liquid::Strainer.class_variable_get(:@@strainer_class_cache).clear
+      original_global_strainer = Liquid::StrainerFactory.instance_variable_get(:@global_filters)
```
davidcornu: nit - `original_global_filters`?

test/unit/strainer_factory_unit_test.rb
```diff
@@ -0,0 +1,95 @@
...
+    assert_raises(Liquid::ArgumentError) do
+      strainer.invoke("public_filter", 1)
+    end
+  end
+
+  def test_stainer_argument_error_contains_backtrace
+    strainer = StrainerFactory.create(nil)
+    begin
+      strainer.invoke("public_filter", 1)
+    rescue Liquid::ArgumentError => e
```
davidcornu: This might not be your code but this test will pass if `strainer.invoke("public_filter", 1)` doesn't raise.

Might I suggest

```ruby
exception = assert_raises(Liquid::ArgumentError) do
  strainer.invoke("public_filter", 1)
end

assert_match(
  /\ALiquid error: wrong number of arguments \((1 for 0|given 1, expected 0)\)\z/,
  exception.message
)

# ...
```

davidcornu:
Refactor makes sense and aids readability. 

It's a API breaking change so it'll have to be a major version bump.

tjoyal:
> It's a API breaking change so it'll have to be a major version bump.

Maybe, maybe not.

Sure it was publicly accessible, but I couldn't find any references to `Liquid::Strainer.global_filter(mod)` in our documentation. Everything seemed to have been telling to use `Liquid::Template.register_filter(mod)` which has the same behaviour.

https://github.com/Shopify/liquid/wiki/Liquid-for-Programmers#create-your-own-filters

I could arguably make a stub to make it a patch? 

eg.:
```
module Liquid
  class Strainer
    def self.global_filter(mod)
      Liquid::Template.register_filter(mod)
    end
  end
end 
```

What do we prefer?

davidcornu:
My general assumption is that publicly-accessible constants can be included in downstream code so changing their names/behaviour is a semver breaking change.

lib/liquid/strainer_template.rb
tjoyal: I've simply removed it since we now call `include?` which will raise a `TypeError`.

This dated from before the code had this shape from the initial git import: https://github.com/Shopify/liquid/blame/1d647361e1a39ee09f1bd2d62701925c0128c578/lib/liquid/strainer.rb#L27

PR #1211, opened 2019-12-14
Pass caller as option to file system register

mmorissette:
## Context
With the `{% include %}` tag now [deprecated](https://help.shopify.com/en/themes/liquid/tags/deprecated-tags) in Shopify, we want to restrict certain new platform features to only work with the `{% render %}` tag. To do so, we are passing the calling tag name to the file system and delegating it the responsibility of determining if the partial is supported with the given tag.

## How
* Pass the caller via an `options` hash to a newly added `read_template_file_with_options` method to the file system register. 
* To avoid breaking backwards compatibility, we only invoke `read_template_file_with_options` if the file system supports it. Otherwise fallback to the previous `read_template_file`. 
* Cache results per calling tag in `PartialCache` 

@Thibaut

### DIFF
--- a/lib/liquid/partial_cache.rb
+++ b/lib/liquid/partial_cache.rb
@@ -2,18 +2,22 @@
 
 module Liquid
   class PartialCache
-    def self.load(template_name, context:, parse_context:)
+    def self.load(template_name, context:, parse_context:, caller:)
       cached_partials = (context.registers[:cached_partials] ||= {})
-      cached = cached_partials[template_name]
+      cached = cached_partials["#{caller}:#{template_name}"]
       return cached if cached
 
       file_system = (context.registers[:file_system] ||= Liquid::Template.file_system)
-      source      = file_system.read_template_file(template_name)
-
       parse_context.partial = true
 
+      source = if file_system.respond_to?(:read_template_file_with_options)
+        file_system.read_template_file_with_options(template_name, caller: caller)
+      else
+        file_system.read_template_file(template_name)
+      end
+
       partial = Liquid::Template.parse(source, parse_context)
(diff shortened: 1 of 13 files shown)


mmorissette:
CI failures are unrelated and only occur in Ruby 2.7

lib/liquid/partial_cache.rb
```diff
@@ -2,18 +2,22 @@
...
       return cached if cached
 
       file_system = (context.registers[:file_system] ||= Liquid::Template.file_system)
-      source      = file_system.read_template_file(template_name)
-
       parse_context.partial = true
 
-      partial = Liquid::Template.parse(source, parse_context)
-      cached_partials[template_name] = partial
+      source = if file_system.respond_to?(:read_template_file_with_options)
```
mmorissette: I chose this approach to avoid breaking backwards compatibility. I am open to other ideas but this seemed like the safest at the moment. 

Also, I went with a rather generic name `read_template_file_with_options` to allow to pass in more options in the future if needed. I am open to being more specific if someone feels like this is too early of a generalization.

lib/liquid/partial_cache.rb
```diff
@@ -2,18 +2,22 @@
 
 module Liquid
   class PartialCache
-    def self.load(template_name, context:, parse_context:)
+    def self.load(template_name, context:, parse_context:, caller:)
```
pushrax: This shadows the `caller` built-in; maybe a clearer and more unique name would be `caller_tag_name`.

mmorissette: Thanks for looking into this @pushrax but we decided to take another approach that doesn't require changes here in `liquid`

PR #1212, opened 2019-12-15
Introduce template factory

mmorissette:
## Context
Currently, liquid will always use a `Liquid::Template` when rendering partials. In Shopify, we want the ability to use different templates depending on the type of partial. 

## What changed

* Use a template factory to create templates for partial
* Use template factory provided in context registers when available
* Add the ability for templates to apply context restrictions during render.

This change has no impact on backward compatibility since we have a default template factory included.

### DIFF
--- a/lib/liquid/template_factory.rb
+++ b/lib/liquid/template_factory.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+module Liquid
+  class TemplateFactory
+    def for(_template_name)
+      Liquid::Template.new
+    end
+  end
+end
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -80,6 +80,7 @@ require 'liquid/partial_cache'
 require 'liquid/usage'
 require 'liquid/register'
 require 'liquid/static_registers'
+require 'liquid/template_factory'
 
 # Load all the tags of the standard library
 #
--- a/lib/liquid/partial_cache.rb
+++ b/lib/liquid/partial_cache.rb
@@ -12,7 +12,10 @@ module Liquid
 
       parse_context.partial = true
 
-      partial = Liquid::Template.parse(source, parse_context)
+      template_factory = (context.registers[:template_factory] ||= Liquid::TemplateFactory.new)
+      template = template_factory.for(template_name)
+
+      partial = template.parse(source, parse_context)
       cached_partials[template_name] = partial
     ensure
       parse_context.partial = false
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -136,3 +136,16 @@ class StubFileSystem
     @values.fetch(template_path)
   end
 end
+
+class StubTemplateFactory
+  attr_reader :count
+
+  def initialize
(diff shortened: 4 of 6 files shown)


lib/liquid/tags/render.rb
```diff
@@ -47,17 +47,19 @@ def render_tag(context, output)
...
 
       render_partial_func = ->(var, forloop) {
-        inner_context               = context.new_isolated_subcontext
-        inner_context.template_name = template_name
-        inner_context.partial       = true
-        inner_context['forloop']    = forloop if forloop
-
-        @attributes.each do |key, value|
-          inner_context[key] = context.evaluate(value)
+        partial.with_context_restrictions(context) do
```
mmorissette: This will allow template files to apply certain restrictions to rendering if needed

Thibaut: Could this be done by overriding `render_to_output_buffer` instead?

mmorissette: Overriding `render_to_output_buffer` would not be enough unfortunately because attributes that are passed in to the partial are evaluated here, outside of the partial's rendering (line 56 below). Overriding `render_to_output_buffer` would not restrict what's passed in as argument to render and therefore we would not be able to block restrict cases like below:

```liquid
{% assign t = theme %}
{% render "snippet_name", theme: t %}
```

mmorissette: By doing it at this level we avoid having to deal with restricting what's passed into attributes below at line 56.

mmorissette: Actually, I take it back, overriding `render_to_output_buffer` is a much cleaner approach. I'll revert the introduction of `with_context_restrictions`

lib/liquid/template_factory.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Liquid
+  class TemplateFactory
+    def for(_template_name)
+      Liquid::Template.new
+    end
+  end
```
dylanahsmith: Why bother having a class for this, since it is just a method without state.  Instead, this could just be

```ruby
  module TemplateFactory
    def self.for(_template_name)
      Liquid::Template.new
    end
  end
```

and use `Liquid::TemplateFactory` instead of `Liquid::TemplateFactory.new` as the default when `context.registers[:template_factory]` isn't set.

mmorissette: I had gone with a class level factory method at first but since all other registries in the liquid `context` are objects, I decided to make it an instance method. This way, registers remain object and avoids surprise. I'll leave it as is if that's ok.

dylanahsmith: > This way, registers remain objects

A singleton object is an object.  You also wouldn't need to memoize the default in the registry, since that was just an optimization to avoid object allocation.

mmorissette: You are correct, I should have been more clear. I meant that all entries in the register get initialized `*.new` as opposed to being a class. I wanted to follow this pattern 🤷‍♂

PR #1213, opened 2019-12-16
Make ruby 2.7 optional in CI

mmorissette:
CI currently fails in Ruby 2.7, this is blocking ongoing work. Making 2.7 optional in CI for now until we can dedicate time to fixing the root issues.


tjoyal:
I'd like to add that we should aim to not have a required CI step which is a moving target.

Last master was built on preview 1 which passed. Preview 2 was still passing and now preview 3 is failing.

PR #1215, opened 2019-12-18
Fix liquid tag nested in outer block

### LINKED ISSUE #1214: `{% liquid %}` tag does not work inside any other tags
@pushrax

The new `{% liquid %}` tag does not seem to work properly when it's used inside other tags.

For example, this code works as expected, and outputs `Hello world`.
```
{% liquid
  echo "Hello world"
%}
```

This code doesn't work at all and generates a `Liquid syntax error (line 4): 'if' tag was never closed` error
```
{% if true %}
  {% liquid
    echo "Hello world"
  %}
{% endif %}
```

The same issue occurs inside other tags I tried.

dylanahsmith:
Fixes #1214

## Problem

The liquid tag seems to have intentionally implemented some weird quirks which made it kind of act like a block and kind of not, as demonstrated by the [test_cannot_open_blocks_living_past_a_liquid_tag and test_quirk_can_close_blocks_created_before_a_liquid_tag tests](https://github.com/Shopify/liquid/blob/6c6382ed69e639e238e4f5088c88a9a828190012/test/integration/tags/liquid_tag_test.rb#L84-L99).  To support the quirk of closing outer block tags, the liquid block body yielded to delegate the handling of end tags, making it behave unlike a block tag.  To support the former test, it yielded a `nil` end tag at the end of the liquid tag to indicate that there weren't any further tokens.  However, that end tag was also being delegated to any block tag around the liquid tag, which prevented the outer block tag from being properly terminated outside the liquid tag.

## Solution

To fix #1214, I stopped delegating the `nil` end tag to the outer block tag.

Ideally, we would have the liquid tag act more like a block, in that block tags inside should be closed inside the liquid tag and blocks outside the liquid tag should be closed outside the liquid tag.  Since that could be a breaking change, I've added a usage increment so that we can see if any templates are currently relying on that behaviour.

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -60,6 +60,14 @@ module Liquid
       Usage.increment("liquid_tag_contains_outer_tag") unless $ERROR_INFO.is_a?(SyntaxError)
     end
 
+    private def parse_liquid_tag(markup, parse_context, &block)
+      liquid_tag_tokenizer = Tokenizer.new(markup, line_number: parse_context.line_number, for_liquid_tag: true)
+      parse_for_liquid_tag(liquid_tag_tokenizer, parse_context) do |end_tag_name, end_tag_markup|
+        next unless end_tag_name
+        self.class.unknown_tag_in_liquid_tag(end_tag_name, end_tag_markup, &block)
+      end
+    end
+
     private def parse_for_document(tokenizer, parse_context, &block)
       while (token = tokenizer.shift)
         next if token.empty?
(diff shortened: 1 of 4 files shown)


lib/liquid/block_body.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'English'
```
Thibaut: TIL

Thibaut:
@pushrax I documented it a month ago https://help.shopify.com/en/themes/liquid/tags/theme-tags#liquid https://github.com/Shopify/help/pull/12170

PR #1216, opened 2019-12-19
Remove support for taint_mode on ruby versions that don't support it

dylanahsmith:
## Problem

Ruby is deprecating support for taint checking (https://bugs.ruby-lang.org/issues/16131).  It won't immediately remove the related methods, it will just treat all objects as untainted.  This means our test suite will fail on the next version of ruby.

## Solution

Deprecate the `Template.taint_mode=` method and have it raise if the taint mode isn't supported by the version of ruby being used.  We can remove this feature in the next breaking release of liquid.

I've also wrapped taint mode tests in a conditional to prevent them from running if taint checking isn't supported by the ruby version being used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,17 +5,12 @@ rvm:
   - 2.4
   - 2.5
   - &latest_ruby 2.6
-  - 2.7
-  - ruby-head
 
 matrix:
   include:
     - rvm: *latest_ruby
       script: bundle exec rake memory_profile:run
       name: Profiling Memory Usage
-  allow_failures:
-    - rvm: ruby-head
-    - rvm: 2.7
 
 branches:
   only:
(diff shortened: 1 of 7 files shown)


dylanahsmith:
> I assume from the `taint_supported?` method that objects are tainted by default historically until the new ruby where nothing is ever tainted.

The objects aren't tainted by default, which is why `taint_supported?` calls `taint` on the object before calling `tainted?`.  Yes, the new version of ruby will make it so nothing is ever tainted, where it previously kept track of whether objects were tainted.

.travis.yml
```diff
@@ -5,17 +5,12 @@ rvm:
   - 2.4
   - 2.5
   - &latest_ruby 2.6
-  - 2.7
-  - ruby-head
```
dylanahsmith: Oops.  I meant to leave these two lines in.

PR #1218, opened 2020-01-06
Remove handling of a nil context in the Strainer class

tjoyal:
From #1208

The strainers are only ever created via a single code path that will lead to always have a `@context` existing.

https://github.com/Shopify/liquid/blob/24e81267b96dad4601f342e0c11182e9cfa16689/lib/liquid/context.rb#L55

Removes the safe navigation operator and fix tests associated.

### DIFF
--- a/History.md
+++ b/History.md
@@ -3,6 +3,7 @@
 ### Unreleased
 
 * Split Strainer class as a factory and a template (#1208) [Thierry Joyal]
+* Remove handling of a nil context in the Strainer class (#1218) [Thierry Joyal]
 
 ## 4.0.3 / 2019-03-12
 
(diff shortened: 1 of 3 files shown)


History.md
```diff
@@ -3,6 +3,7 @@
 ### Unreleased
 
 * Split Strainer class as a factory and a template (#1208) [Thierry Joyal]
+* Remove safe navigation (#1218) [Thierry Joyal]
```
dylanahsmith: It won't be clear what this is referring to without actually looking at this pull request.  Perhaps "Remove handling of a nil context in the Strainer class".

tjoyal: Updated code and PR description

Issue #1205, opened 2019-12-11
Execution paths missing @context


tjoyal:
### General

I've been struggling for some time with certain code paths in my application where the `@context` within `Liquid::Drop` is `nil` and finally decided to deep dive.

In liquid, when the original object is retrieved from the vm environment, context is applied properly:

https://github.com/Shopify/liquid/blob/57c9cf64ebc777fe5e92d4408d31a911f087eeb4/lib/liquid/variable_lookup.rb#L67-L68
https://github.com/Shopify/liquid/blob/57c9cf64ebc777fe5e92d4408d31a911f087eeb4/lib/liquid/context.rb#L191-L192

The @context, is in these cases, used as a global variable to alter behaviour based on the current user for which the template is rendered. This is not always great, but it is what it is it purpose.

### Example

```rb
require 'liquid'

class BaseDrop < Liquid::Drop
  def initialize(var)
    @var = var
  end

  def value
    if @context.registers.fetch(:upcase)
      @var.upcase
    else
      @var
    end
  rescue StandardError => e
    e.to_s
  end
  alias_method :to_s, :value
end

environments = {
  "single" =>  BaseDrop.new("aaa"),
  "array" => [
    BaseDrop.new("aaa"),
    BaseDrop.new("bbb"),
    BaseDrop.new("ccc"),
  ]
}
registers = {
  upcase: true,
}
context = Liquid::Context.new(environments, {}, registers)

template = 'Result: {{ single | json }}'
Liquid::Template.parse(template).render!(context)
#=> "Result: AAA"

template = 'Result: {{ array | json }}'
Liquid::Template.parse(template).render!(context)
#=> "Result: undefined method `registers' for nil:NilClassundefined method `registers' for nil:NilClassundefined method `registers' for nil:NilClass"
```

### Propose mitigation

Now, I couldn't come up with something great to get around this, maybe just few step to make it less painful.

I see this as a multiple issues.

A) lib/liquid/extensions.rb patches `Array#to_liquid`. It does not iterate to perform `to_liquid` and assignation of @context to all the elements. This leaves "non liquid" object when the contract in my opinion should have been that `to_liquid` is expected to do just that.

[...]

PR #1220, opened 2020-01-07
[StrainerTemplate] Isolate filter mods

tjoyal:
From #1208

This is a WIP, some perf concerns and missing tests.

This is my simple and naive attempt at isolating filter modules from one another. This way private methods from one filter are not available by another.

Some items I need to figure out:
- method is called `add_filter` but really it's `add_filters_module` as said module contain multiple "filters".
- Current logic calls `new` before invoking the filters, it should be limited to once per rendering (as the current logical layer).
- Missing tests.

I am iterating this was as I want to get to a point where "filters" can eventually become classes like `Liquid::Drop` (`Liquid::Filter`?) and for them to have additional responsibilities. I desire to explore a strict mode that will validate the presence of registers declared as dependency by filters before rendering is attempted as per my work on trying to make the lifecycle of developer easier when it comes to tracking issues with the absence of context or sub elements from it (https://github.com/Shopify/liquid/issues/1205). My goal is to have a smarter system that would detect inconsistencies in CI rather than waiting for a production execution to stumble on the error cases.


lib/liquid/strainer_template.rb
```diff
@@ -14,40 +14,67 @@ def initialize(context)
...
 
       private
 
-      def filter_methods
-        @filter_methods ||= Set.new
+      def filter_instances
+        @filter_instances ||= {}
+      end
+
+      # Caching here is most likely not required anymore since we cache instances and there is only one instance per filter class.
```
tjoyal: ^ Draft note

lib/liquid/strainer_template.rb
```diff
@@ -14,40 +14,67 @@ def initialize(context)
...
       end
 
       def invokable?(method)
-        filter_methods.include?(method.to_s)
+        filter_instances.key?(method)
+      end
+
+      def invoke(method, context, *args)
+        instance = filter_instances.fetch(method)
+        instance.context = context
```
dylanahsmith: It looks like `filter_instances` are global, in which case this assignment wouldn't be thread-safe.

lib/liquid/strainer_template.rb
```diff
@@ -14,40 +14,67 @@ def initialize(context)
...
+        @filter_classes[mod] ||= begin
+          klass = Class.new(FilterTemplate)
+          klass.include(mod)
+          klass
+        end
+      end
+
+      def filter_class_by_methods
+        @filter_class_by_methods ||= {}
       end
```
dylanahsmith: This is unused right now

lib/liquid/strainer_template.rb
tjoyal: Good point. As I tried to avoid the `.new` I created this.

tjoyal:
I went back to initializing filters. This mean filters are initialized once each per content rendering. Uncertain of the perf tradeoff here, shouldn't be much?

If we really want to focus on perf, we could achieve larger things if we are to consider a bit of a paradigm change. 

Filters as classes(added here) have the possibility to be completely stateless and moving away from modules would allow us to revisit the signatures.

Eg.: Not instantiating
```rb
class SomeFilter < Liquid::Filter
  def self.filter_name(context, input)
    ...
  end
end
``` 

But then revisiting open other doors. We could shift the responsibility of `@context` away from liquid and directly to the caller. 

Traditionally one would set registries as an unstructured hash (which he need to know by reading the internals of each filters included). Then liquid would make these available to the filters.

By moving the responsibility around, code then can ensure required registry presence by adding it to the filter signature, making the code less error prone and easier to maintain.

I think this remove the need for context completely in filters.

```rb
class MagicSuffixFilter < Liquid::Filter
  def initialize(suffix)
    @suffix = suffix
  end

  def self.suffix(input)
    "#{input}_#{@suffix}"
  end
end

tmpl = Liquid::Template.parse('Result: {{ "test" | suffix }}')
filters = [
   MagicSuffixFilter.new("abcd")
]
tmpl.render({}, filters: filters)
=> "test_abcd"
```

devdreamsolution (approved):

PR #1222, opened 2020-01-09
Test against the latest ruby in CI

dylanahsmith:
I meant to test against ruby 2.7 and ruby-head and not allow failures in #1216 after adding handling the removal of taint checking in ruby 2.7.  Somehow I accidentally ended up removing those intentionally.

Now that ruby 2.7 is released, we should definitely make sure liquid works with it.

Closes #1209


dylanahsmith:
I had to upgrade rubocop and the style guide to be compatible with ruby 2.7's new handling of keyword arguments.

PR #1223, opened 2020-01-10
Initialize Liquid::Template class attributes eagerly instead of lazily

dylanahsmith:
Closes #1219

## Problem

Initializing class attributes lazily through memoization adds extra runtime overhead and prevents us from leveraging ruby's optimization of attribute accessor method calls.  This can be seem from the following micro benchmark

```ruby
require 'benchmark'

class Foo
  class << self
    attr_accessor :eager
    Foo.eager = 1
    private :eager=

    def lazy
      @lazy ||= 1
    end
  end
end

N = 10000000

Benchmark.bmbm do |x|
  x.report("eager") do
    N.times do
      Foo.eager
    end
  end
  x.report("lazy") do
    N.times do
      Foo.lazy
    end
  end
end
```

which outputs

```
eager   0.643216   0.008314   0.651530 (  0.658694)
lazy    0.991742   0.010960   1.002702 (  1.012705)
```

## Solution

Use `attr_accessor` and do assignment immediately afterwards on the singleton object.  If we don't want to expose the writer, then just mark it as private after using it to assign the default value.

### DIFF
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -18,8 +18,6 @@ module Liquid
     attr_accessor :root
     attr_reader :resource_limits, :warnings
 
-    @@file_system = BlankFileSystem.new
-
     class TagRegistry
       include Enumerable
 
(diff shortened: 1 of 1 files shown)


dylanahsmith:
> I'm not convinced the deviation from the code you'd expect to see is worth it.

I'm not sure what you mean about this deviating from the code I expect to see?  It isn't like I am doing anything obscure here, I'm just using an `attr_accessor`.  Ruby does make it kind of awkward to define the attr_reader in the singleton class and initialize it there as well, which is probably why this annoyance was worked around by defining a memoizing method instead.

PR #1224, opened 2020-01-12
Set instance variables of the singleton class

ashmaroli:
- **Initialize singleton class instance_variables instead of using `private attr_writer` methods** &mdash; since those private methods don't seem to be used elsewhere consequently.
- Define `attr-*` of the singleton class at the top of the class block (similar to how `attr-*`s for regular class instances are defined)
- **Correct comment to reflect the fact that `:lax` is the default `error_mode` instead of `:warn`.**
- Insert whitespace into comments to improve their readability.

/cc @dylanahsmith, @tjoyal, @davidcornu

### DIFF
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -56,21 +56,33 @@ module Liquid
 
     attr_reader :profiler
 
+    # Eagerly initialize default attributes of the <tt>Template</tt> singleton class.
+    @error_mode = :lax
+    @taint_mode = :lax
+    @registers  = {}
+
+    @tags        = TagRegistry.new
+    @file_system = BlankFileSystem.new
+
+    @default_resource_limits = {}
+    @default_exception_renderer = lambda { |exception| exception }
+
     class << self
-      # Sets how strict the parser should be.
-      # :lax acts like liquid 2.5 and silently ignores malformed tags in most cases.
-      # :warn is the default and will give deprecation warnings when invalid syntax is used.
-      # :strict will enforce correct syntax.
+      attr_reader :tags, :registers, :default_resource_limits
+      attr_accessor :default_exception_renderer, :file_system
+      # Controls how strict the parser should be.
+      #      :lax : is the default, acts like liquid 2.5 and silently ignores malformed tags in most cases.
+      #     :warn : will give deprecation warnings when invalid syntax is used.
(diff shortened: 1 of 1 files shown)


lib/liquid/template.rb
```diff
@@ -56,21 +56,33 @@ def lookup_class(name)
...
+    @tags        = TagRegistry.new
+    @file_system = BlankFileSystem.new
+
+    @default_resource_limits = {}
+    @default_exception_renderer = lambda { |exception| exception }
+
     class << self
-      # Sets how strict the parser should be.
-      # :lax acts like liquid 2.5 and silently ignores malformed tags in most cases.
-      # :warn is the default and will give deprecation warnings when invalid syntax is used.
```
dylanahsmith: Looks like the default got modified from `:warn` to `:lax` in commit 3b3961be398c75f72f215b2f929f422c53a8df47 without modifying the documentation.  I think separating the default value assignment from the attribute definition and its documentation will make this mistake even easier to make in the future, along with making it easy to remove an attribute without removing the instance variable assignment.  That is why I tried to keep them together in https://github.com/Shopify/liquid/pull/1223.

ashmaroli: > I think separating the default value assignment from the attribute definition and its documentation will make this mistake even easier to make in the future

I agree in principle but isn't this how it is (convention) for every class *instance* &mdash; initialize instance attribute accessors at the top of the class definition but assign values later in a separate `:initialize` method.. ?

<br/>

> making it easy to remove an attribute without removing the instance variable assignment

Ideally, this should be prevented by adequate test coverage.

dylanahsmith: Class instances are different, in that we need to assign the variable for each instance, so the initialize method is the earliest opportunity to assign it.

Using an attr_writer can make sense on the instance as well.  Test coverage won't help with removing dead code and relying on test coverage will provide a worse debugging experience compared to a NoMethodError when there is a typo or partial refactor that causes the instance variable name to not match the definition.

On the other hand, calling an `attr_writer` method introduces a slight performance overhead, so setting the instance variable directly can make sense if performance is the bigger concern.

However, all of this is kind of a bikeshedding, so I'm not going to block this PR over my opinion now that I have explained it.

devdreamsolution (approved):

Issue #1249, opened 2020-05-21
Calling fetch over registers does not have the expected behaviour


tjoyal:
Still piecing the history together. 

From what I understand in sept 2019 registers were moved from a `Hash` structure to be `Liquid::StaticRegisters` https://github.com/Shopify/liquid/pull/1157

In the legacy behaviour, a `Hash` was returned:
```
requirb(main):001:0> require 'liquid'
=> true
irb(main):002:0> {a: 1}.fetch(:a)
=> 1
irb(main):003:0> {a: 1}.fetch(:b, "test")
=> "test"
irb(main):004:0> {a: 1}.fetch(:b)
Traceback (most recent call last):
        5: from /opt/rubies/2.6.5/bin/irb:23:in `<main>'
        4: from /opt/rubies/2.6.5/bin/irb:23:in `load'
        3: from /opt/rubies/2.6.5/lib/ruby/gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):4
        1: from (irb):4:in `fetch'
KeyError (key not found: :b)
```

The new behaviour with `Liquid::StaticRegisters`:
```
[1] pry(main)> Liquid::StaticRegisters
=> Liquid::StaticRegisters
[2] pry(main)> Liquid::StaticRegisters.new(a: 1).fetch(:a)
=> 1
[3] pry(main)> Liquid::StaticRegisters.new(a: 1).fetch(:b, "test")
=> "test"
[4] pry(main)> Liquid::StaticRegisters.new(a: 1).fetch(:b)
=> nil
```

This is particularly troubling when the codebase was built around using `fetch` as an early warning to detect missing registers. I believe `nil` is not a sane default when using `fetch`.

PR #1250, opened 2020-05-21
[StaticRegisters] Fetch raise on missing

tjoyal:
Fixes https://github.com/Shopify/liquid/issues/1249

~~CI is red on ruby-head, see https://github.com/Shopify/liquid/pull/1251~~ (merged and rebased)

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,7 @@
 
 * Split Strainer class as a factory and a template (#1208) [Thierry Joyal]
 * Remove handling of a nil context in the Strainer class (#1218) [Thierry Joyal]
+* StaticRegisters#fetch to raise on missing key (#1250) [Thierry Joyal]
 
 ## 4.0.3 / 2019-03-12
 
--- a/lib/liquid/static_registers.rb
+++ b/lib/liquid/static_registers.rb
@@ -25,8 +25,16 @@ module Liquid
       @registers.delete(key)
     end
 
-    def fetch(key, default = nil)
-      key?(key) ? self[key] : default
+    UNDEFINED = Object.new
+
+    def fetch(key, default = UNDEFINED, &block)
+      if @registers.key?(key)
+        @registers.fetch(key)
+      elsif default != UNDEFINED
+        @static.fetch(key, default, &block)
+      else
+        @static.fetch(key, &block)
+      end
     end
 
     def key?(key)
--- a/test/unit/static_registers_unit_test.rb
+++ b/test/unit/static_registers_unit_test.rb
@@ -51,7 +51,15 @@ class StaticRegistersUnitTest < Minitest::Test
     assert_equal("one", static_register.fetch(:one))
     assert_equal(3, static_register.fetch("two"))
     assert_nil(static_register.fetch(false))
-    assert_nil(static_register.fetch("unknown"))
+
+    assert_raises(KeyError) do
+      static_register.fetch(:unknown)
+    end
+    assert_equal("default", static_register.fetch(:unknown, "default"))
+
+    static_register[:unknown] = "known"
+    assert_equal("known", static_register.fetch(:unknown))
+    assert_equal("known", static_register.fetch(:unknown, "default"))
   end
 
   def test_fetch_default


lib/liquid/static_registers.rb
```diff
@@ -25,8 +25,12 @@ def delete(key)
       @registers.delete(key)
     end
 
-    def fetch(key, default = nil)
-      key?(key) ? self[key] : default
+    def fetch(key, *args, &block)
+      if @registers.key?(key)
+        @registers.fetch(key)
+      else
+        @static.fetch(key, *args, &block)
```
dylanahsmith, line 32: Using `*args` parameter will end up unnecessarily allocating an array.  So this could cause a performance problem if it is called often enough.  Although, it doesn't look like we are using this method internally to liquid, so it might not matter.

tjoyal: This would be whenever filters try to access registries. It could indeed be called multiple times depending on the use in filters.

I'm open to suggestions if any. 

If we really wanted to go after performance, we could consider generating a single unified Hash whenever there is a mutation and avoid checkout 2 of them every call.

dylanahsmith: If we want to avoid the performance overhead, then you could use

```suggestion
    UNDEFINED = Object.new

    def fetch(key, default=UNDEFINED, &block)
      if @registers.key?(key)
        @registers.fetch(key)
      elsif default != UNDEFINED
        @static.fetch(key, default, &block)
      else
        @static.fetch(key, &block)
```

dylanahsmith: > If we really wanted to go after performance, we could consider generating a single unified Hash whenever there is a mutation and avoid checkout 2 of them every call.

Yeah, this might make more sense.  If we typically mutate the registers, we could just unconditionally merge them ahead of time, which would be quite simple

History.md
```diff
@@ -4,6 +4,7 @@
 
 * Split Strainer class as a factory and a template (#1208) [Thierry Joyal]
 * Remove handling of a nil context in the Strainer class (#1218) [Thierry Joyal]
+* StaticRegisters#fetch to raise on missing (#1250) [Thierry Joyal]
```
Thibaut, line 7: "on missing _key_"?

lib/liquid/static_registers.rb
```diff
@@ -25,8 +25,16 @@ def delete(key)
...
 
-    def fetch(key, default = nil)
-      key?(key) ? self[key] : default
+    UNDEFINED = Object.new
+
+    def fetch(key, default = UNDEFINED, &block)
+      if @registers.key?(key)
+        @registers.fetch(key)
+      elsif default != UNDEFINED
+        @static.fetch(key, default, &block)
```
mbj, line 34: @tjoyal This block pass, and the one below is not covered by the tests you added in this PR.

Was the addition of the block pass intentional? If so: I suggest to add a test that shows this intention, else the block pass could be removed by a co-worker and break assumptions elsewhere, without tests of this unit noticing.

dylanahsmith: You're right, the tests pass with

```diff
diff --git a/lib/liquid/static_registers.rb b/lib/liquid/static_registers.rb
index 0293f5f..94cb660 100644
--- a/lib/liquid/static_registers.rb
+++ b/lib/liquid/static_registers.rb
@@ -31,9 +31,9 @@ module Liquid
       if @registers.key?(key)
         @registers.fetch(key)
       elsif default != UNDEFINED
-        @static.fetch(key, default, &block)
+        @static.fetch(key, default)
       else
-        @static.fetch(key, &block)
+        @static.fetch(key)
       end
     end
 
```

> Was the addition of the block pass intentional?

Yes, the intention was to have it behave like a Hash#fetch, so it makes sense to pass along the block.

mbj: k, cool. So I can build on this block pass behavior? Its intentionally part of the public API of this class (and will be in the future)?

dylanahsmith: Yes.  We will also accept tests to help preserve that behaviour.

PR #1252, opened 2020-05-21
[StaticRegisters] Cache for reads

tjoyal:
Optimization being discusses: https://github.com/Shopify/liquid/pull/1250/files#r428825865

Remove the possibly doubled hash check by creating a "at write time" cache entry.

Thoughts?

### DIFF
--- a/lib/liquid/static_registers.rb
+++ b/lib/liquid/static_registers.rb
@@ -1,8 +1,14 @@
 # frozen_string_literal: true
 
+require 'forwardable'
+
 module Liquid
   class StaticRegisters
+    extend Forwardable
+
     attr_reader :static, :registers
+
+    def_delegators :@cache, :[], :key?
 
     def initialize(registers = {})
       @static    = registers.is_a?(StaticRegisters) ? registers.static : registers
@@ -16,22 +22,14 @@ module Liquid
       @cache[key] = value
     end
 
-    def [](key)
-      @cache[key]
-    end
-
     def delete(key)
-      @registers.delete(key).tap do
-        @static.dup.merge(@registers)
-      end
+      deleted = @registers.delete(key)
+      @cache = @static.dup.merge(@registers)
+      deleted
     end
 
     def fetch(key, default = nil)
       key?(key) ? self[key] : default
-    end
-
-    def key?(key)
-      @cache.key?(key)
     end
   end
 end
--- a/lib/liquid/static_registers.rb
+++ b/lib/liquid/static_registers.rb
@@ -7,22 +7,23 @@ module Liquid
     def initialize(registers = {})
       @static    = registers.is_a?(StaticRegisters) ? registers.static : registers
       @registers = {}
+
+      @cache = @static.dup
     end
 
     def []=(key, value)
       @registers[key] = value
+      @cache[key] = value
     end
 
     def [](key)
-      if @registers.key?(key)
-        @registers[key]
-      else
-        @static[key]
-      end
+      @cache[key]
     end
 
     def delete(key)
-      @registers.delete(key)
+      @registers.delete(key).tap do
+        @static.dup.merge(@registers)
+      end
     end
 
     def fetch(key, default = nil)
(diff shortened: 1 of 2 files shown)


lib/liquid/static_registers.rb
```diff
@@ -1,36 +1,35 @@
...
 
+require 'forwardable'
+
 module Liquid
   class StaticRegisters
+    extend Forwardable
+
     attr_reader :static, :registers
 
+    def_delegators :@cache, :[], :key?
```
tjoyal, line 11: Here in https://github.com/Shopify/liquid/pull/1250, the PR become to simply add the method here and to :fire: custom logic.

dylanahsmith: If you are trying to avoid `*args` looks like `def_delegators` has the same problem (https://github.com/ruby/ruby/blob/38a4f617de157586668dd726d518eadcebf1bca2/lib/forwardable.rb#L207-L233), although perhaps that could be fixed upstream to leverage `...`

lib/liquid/static_registers.rb
```diff
@@ -1,36 +1,35 @@
...
-      else
-        @static[key]
-      end
+      @cache[key] = value
     end
 
     def delete(key)
-      @registers.delete(key)
+      @registers.delete(key).tap do
+        @static.dup.merge(@registers)
```
dylanahsmith, line 27: I don't understand this tap block, since `tap` doesn't return the result of the block so it looks like this block is just building an object that won't get used.

tjoyal: I had this before that I can put back if you desire?
```rb
def delete(key)
  value = @registers.delete(key)
  @static.dup.merge(@registers)
  value
end
```

tjoyal: Actually I'm maintaining a behaviour that is tested, but we are doing a breaking change in a subsequent PR. 

I wouldn't mind if we preferred to do further breaking changes.

lib/liquid/static_registers.rb
tjoyal: I was actually trying to avoid duplicating logic and making sure StaticRegister methods that were mimicking Hash behaved the same way.

I'm open to do what is needed to make iterative changes (I could stick to the first commit of this PR)

lib/liquid/static_registers.rb
dylanahsmith: A `@static.dup.merge(@registers)` line is the same thing, create a new hash but then doesn't use it.

tjoyal: Haaa it's because I'm dumb and forgot to assign it to `@cache`

lib/liquid/static_registers.rb
```diff
@@ -1,36 +1,35 @@
...
+
     attr_reader :static, :registers
 
+    def_delegators :@cache, :[], :key?
+
     def initialize(registers = {})
       @static    = registers.is_a?(StaticRegisters) ? registers.static : registers
       @registers = {}
+
+      @cache = @static.dup
```
dylanahsmith, line 17: Why bother having yet another cache.  If we are going to eagerly duplicate the static registers, then we could just do that for `@registers`

```suggestion
      @static = registers.is_a?(StaticRegisters) ? registers.static : registers
      @registers = @static.dup
```

However, if there are a lot of registers that aren't used in the rendered snippet, that could add extra overhead.  Was `Liquid::StaticRegisters` added to do this lazily for performance reasons?  Otherwise, why do we bother even having it in the first place when we could just store the static registers in the currently unused `Liquid::Context#static_registers` attribute

tjoyal: The api expose an attr reader for :registers, I'm not sure of what would be the implications on the public contract...

tjoyal:
After some discussions, it seems to me that the intent here was about creating a lazy form of "hard.dup" to avoid paying the operation when not required.

It is hard to find a caching solution that would be best for all use cases (small vs large amount of keys in the static registry).

My suggestion at this point would be to explore if we could have a non liquid specific implementation of this same class (rename) and have it support more of the hash primitives. Also to move the specifics of "static" out of here and simply initialize it accordingly from the liquid scope.

PR #1257, opened 2020-05-22
[StaticRegisters] Remove registers attr_reader

tjoyal:
While we are not shipping a change yet for caching on https://github.com/Shopify/liquid/pull/1252, it seems like the idea is to keep iterating a bit.

I propose removing `registers` from the public api as there is no clear reasons why should it be accessed directly from a filter and drops since this is internal management that should be opaque.

### DIFF
--- a/test/unit/static_registers_unit_test.rb
+++ b/test/unit/static_registers_unit_test.rb
@@ -15,15 +15,10 @@ class StaticRegistersUnitTest < Minitest::Test
     assert_equal(33, static_register[:c])
   end
 
-  def test_get
-    static_register = StaticRegisters.new(a: 1, b: 2)
-    static_register[:b] = 22
-    static_register[:c] = 33
+  def test_get_missing_key
+    static_register = StaticRegisters.new
 
-    assert_equal(1, static_register[:a])
-    assert_equal(22, static_register[:b])
-    assert_equal(33, static_register[:c])
-    assert_nil(static_register[:d])
+    assert_nil(static_register[:missing])
   end
 
   def test_delete
@@ -152,7 +147,6 @@ class StaticRegistersUnitTest < Minitest::Test
     static_register_1.static[:b] = 222
     static_register_1.static[:c] = 333
 
-    assert_equal(222, static_register_2[:b])
-    assert_equal(333, static_register_2[:c])
+    assert_same(static_register_1.static, static_register_2.static)
   end
 end
--- a/test/unit/static_registers_unit_test.rb
+++ b/test/unit/static_registers_unit_test.rb
@@ -5,252 +5,154 @@ require 'test_helper'
 class StaticRegistersUnitTest < Minitest::Test
   include Liquid
 
-  def set
-    static_register        = StaticRegisters.new
-    static_register[nil]   = true
-    static_register[1]     = :one
-    static_register[:one]  = "one"
-    static_register["two"] = "three"
-    static_register["two"] = 3
-    static_register[false] = nil
+  def test_set
+    static_register = StaticRegisters.new(a: 1, b: 2)
+    static_register[:b] = 22
+    static_register[:c] = 33
 
-    assert_equal({ nil => true, 1 => :one, :one => "one", "two" => 3, false => nil }, static_register.registers)
-
-    static_register
+    assert_equal(1, static_register[:a])
+    assert_equal(22, static_register[:b])
+    assert_equal(33, static_register[:c])
   end
 
   def test_get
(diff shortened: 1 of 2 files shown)


test/unit/static_registers_unit_test.rb
```diff
@@ -5,252 +5,154 @@
...
-    assert_equal(3, static_register["two"])
-    assert_nil(static_register[false])
-    assert_nil(static_register["unknown"])
+    static_register = StaticRegisters.new(a: 1, b: 2)
+    static_register[:b] = 22
+    static_register[:c] = 33
+
+    assert_equal(1, static_register[:a])
+    assert_equal(22, static_register[:b])
+    assert_equal(33, static_register[:c])
```
dylanahsmith, line 25: Doesn't this do the same thing as `test_set`?  If the only thing new tested in this test method is getting an missing key, then maybe the test should be renamed to `test_get_missing_key` and just do `assert_nil(StaticRegisters.new({})[:missing])`

test/unit/static_registers_unit_test.rb
```diff
@@ -5,252 +5,154 @@
...
+    static_register[:c] = 33
 
-    assert_equal(true, static_register.delete(nil))
-    assert_equal(:one, static_register.delete(1))
-    assert_equal("one", static_register.delete(:one))
-    assert_equal(3, static_register.delete("two"))
-    assert_nil(static_register.delete(false))
-    assert_nil(static_register.delete("unknown"))
+    assert_nil(static_register.delete(:a))
+    assert_equal(1, static_register[:a])
```
dylanahsmith, line 35: Is this behaviour intentional?  It wasn't previously covered by `test_delete`.  If the registers are supposed to behave like a hash, then having delete not actually behave like a delete seems like unexpected behaviour.

test/unit/static_registers_unit_test.rb
```diff
@@ -5,252 +5,154 @@
...
 
-    assert_equal({ nil => true, 1 => :one, :one => "one", "two" => 3, false => nil, "four" => 5, "five" => 15 }, newest_register.static)
+    static_register_1.static[:b] = 222
+    static_register_1.static[:c] = 333
 
-    assert_equal({ nil => true, 1 => :one, :one => "one", "two" => 3, false => nil, "four" => 5, "five" => 15 }, static_register.static)
-    assert_equal({ nil => true, 1 => :one, :one => "one", "two" => 3, false => nil, "four" => 5, "five" => 15 }, new_register.static)
+    assert_equal(222, static_register_2[:b])
+    assert_equal(333, static_register_2[:c])
   end
```
dylanahsmith, line 157: We could use `assert_same` to ensure the `static` is exactly the same object

```suggestion
    assert_same(static_register_1.static, static_register_2.static)
  end
```

test/unit/static_registers_unit_test.rb
tjoyal: > Is this behaviour intentional?

My best guess is that it is not, but it is what the code does.

This also have implication over how `key?` behave.

I'm not sure how these methods were chosen vs some others that were removed when moving from `Hash` to `StaticRegisters` (eg.: `.each` or `empty?`).

This might be going back to "StaticRegisters" shouldn't exist in the first place.

dylanahsmith: > This also have implication over how `key?` behave.

Yeah, to maintain the current API, we would either need to keep track of deleted keys separately or to have a deleted key marker (e.g. similar to the `UNDEFINED` constant).  That would mean we would always have to retrieve the value and special case this marker value.  That makes me wonder if it is worth supporting that complexity to preserve this approach.

> I'm not sure how these methods were chosen vs some others that were removed when moving from `Hash` to `StaticRegisters` (eg.: `.each` or `empty?`).

Good question.  Perhaps it means we actually rely on this `delete` method, but if we don't need it, it would be simpler to not support.

> This might be going back to "StaticRegisters" shouldn't exist in the first place.

Indeed.  This would be much simpler if we just had two separate hashes on the context where `context.registers` could be initialized with `context.static_registers.dup`.  If this lazy approach is a performance optimization, then it seems like it was added prematurely.

dylanahsmith: For this PR, maybe we should just remove the assertions for the value after it is deleted, rather than making it seems like the current behaviour is intentional.

tjoyal: Deal, I'm fine with ship and iterate.

PR #1274, opened 2020-08-11
Fix performance regression from introduction of Template#disable_tags

dylanahsmith:
## Problem

Pull request #1162 introduced a significant performance regression due to adding support for disabling tags into the hot path of block body rendering for each tag or variable.  Yet the only tag that disables other tags is the `render` tag and it only disables one other tag by default.

## Solution

Split the tag disabling functionality into two modules:
1. Liquid::Tag::Disabler that sets the context, around rendering the tag, to disable the other tags
2. Liquid::Tag::Disableable that checks if the tag is disabled before rendering and outputs an error if it is disabled

By prepending these modules only the relevant tags, we avoid introducing unnecessary overhead to other tags.  The first module is automatically prepended by the call to Liquid::Tag#disable_tags.

## Benchmark

`bundle exec rake benchmark:run` without this PR:

```
              parse:     36.185  (± 5.5%) i/s -    360.000  in  10.006807s
             render:     92.179  (± 3.3%) i/s -    927.000  in  10.069611s
     parse & render:     25.911  (± 3.9%) i/s -    260.000  in  10.038593s
```

with this PR:

```
              parse:     35.419  (± 5.6%) i/s -    354.000  in  10.046604s
             render:    103.501  (± 3.9%) i/s -      1.040k in  10.068599s
     parse & render:     26.233  (± 3.8%) i/s -    264.000  in  10.074341s
```

The same regression is seen in liquid-c, since this code hasn't been implemented in C.

`bundle exec rake benchmark:run` from liquid-c without this PR:

```
              parse:    144.044  (± 5.6%) i/s -      1.442k in  10.045295s
             render:    122.332  (± 3.3%) i/s -      1.232k in  10.084915s
     parse & render:     61.587  (± 3.2%) i/s -    620.000  in  10.073747s
```

with this PR (by updating the liquid-c Gemfile to use these changes):

```
              parse:    145.725  (± 2.7%) i/s -      1.456k in  10.001865s
             render:    132.637  (± 3.8%) i/s -      1.326k in  10.014300s
     parse & render:     63.406  (± 1.6%) i/s -    636.000  in  10.035315s
```


lib/liquid/block_body.rb
```diff
@@ -171,25 +171,14 @@ def render_to_output_buffer(context, output)
     private
 
     def render_node(context, output, node)
-      if node.disabled?(context)
-        output << node.disabled_error_message
```
dylanahsmith, line 175: This seems like it should really be using context.handle_error, but I'll handle inconsistency with other liquid errors in a separate PR.

dylanahsmith:
> The liquid-c benchmark seems to indicate the new code is slower but it's all within margin of error.

Oops, it looks like I made a mistake in pasting the benchmark results, so got them mixed up. I've edited the PR description to fix the mistake.  The render line is the one that is relevant and effect is significant is significant at about a 9% improvement. I re-ran the benchmark to verify (got 121.705 i/s on master and 133.317 i/s on the branch)

Thibaut:
👍

PR #1287, opened 2020-08-31
Handle BlockBody#blank? at parse time

dylanahsmith:
## Problem

Liquid::BlockBody#render_to_output_buffer has special case handling for blank block bodies which seems like it should instead be handled at parse time so that rendering can be simplified.

This is another change motivated by trying to simplify Liquid::BlockBody#render_to_output_buffer in order to implement it in liquid-c.

## Solution

Previously, `BlockBody#blank?` was handled by passing in a different string to blank block tags so that the blank output didn't end up in the final output.

At first, I realized I could push this logic down to the block tags themselves, since this feature was only really important for the control flow logic block tags (`if`, `for`, `case`, `unless` tags).  However, then I further realized that the blank output was always coming raw strings, since the feature was added to handle the blank space character unintentionally included between tags used for computations (e.g. a conditional assignment or a for loop to append/concat to a string/array).

So I ended up just removing all the blank strings from the children of the relevant control flow blocks at the end of parsing these blocks.  There is no need to do this recursively, since any child blank block would have already have done the same thing.

I also had to fix the capture flag that was relying on it being given an empty string from BlockBody#render_to_output_buffer.  Instead, I used `render` in the `capture` tag to explicitly render to a separate output string to assign.

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -130,6 +130,26 @@ module Liquid
       @blank
     end
 
+    # Remove blank strings in the block body for a control flow tag (e.g. `if`, `for`, `case`, `unless`)
+    # with a blank body.
+    #
+    # For example, in a conditional assignment like the following
+    #
+    # ```
+    # {% if size > max_size %}
+    #   {% assign size = max_size %}
+    # {% endif %}
+    # ```
+    #
+    # we assume the intention wasn't to output the blank spaces in the `if` tag's block body, so this method
+    # will remove them to reduce the render output size.
+    #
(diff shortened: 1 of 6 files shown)


lib/liquid/block_body.rb
```diff
@@ -131,6 +131,11 @@ def blank?
       @blank
     end
 
+    def remove_blank_strings
```
pushrax, line 134: Might be worth commenting some of the details about `@blank` here to make it clear when/why this should be called.

pushrax (approved):
Makes sense. It's not ideal that this concern now requires users of the BlockBody class to be aware, but the result is cleaner at runtime.

PR #1289, opened 2020-09-08
Avoid direct coupling to BlockBody instances for liquid-c replacement

dylanahsmith:
## Problem

We want to use a C data structure for the block body in liquid-c (https://github.com/Shopify/liquid-c/pull/58) that will be exposed to ruby as Liquid::C::BlockBody, but currently Liquid::Document is directly coupled to Liquid::BlockBody.  There are also several places where `BlockBody.new` is used directly, which doesn't provide a clean way to override in liquid-c to use `Liquid::C::BlockBody`. instead.

## Solution

The first commit adds Liquid::Block#new_body that just returns `BlockBody.new` so that it can be overridden in liquid-c to return `Liquid::C::BlockBody.new`.

The second commit exposes some Liquid::BlockBody instance methods as class methods (raise_missing_tag_terminator, raise_missing_variable_terminator, render_node), since they don't depend on `self` and can be re-used for simplicity in liquid-c.

The third commit changes Liquid::Document to not inherit from Liquid::BlockBody, but instead to store a Liquid::BlockBody.  It also creates the block body using a `#new_body` method that can be overridden by liquid-c.

### DIFF
--- a/lib/liquid/document.rb
+++ b/lib/liquid/document.rb
@@ -1,15 +1,26 @@
 # frozen_string_literal: true
 
 module Liquid
-  class Document < BlockBody
+  class Document
     def self.parse(tokens, parse_context)
-      doc = new
+      doc = new(parse_context)
       doc.parse(tokens, parse_context)
       doc
     end
 
+    attr_reader :parse_context, :body
+
+    def initialize(parse_context)
+      @parse_context = parse_context
+      @body = new_body
+    end
+
+    def nodelist
+      @body.nodelist
+    end
+
     def parse(tokens, parse_context)
-      super do |end_tag_name, _end_tag_params|
+      @body.parse(tokens, parse_context) do |end_tag_name, _end_tag_params|
         unknown_tag(end_tag_name, parse_context) if end_tag_name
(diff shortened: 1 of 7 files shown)


lib/liquid/document.rb
```diff
@@ -25,5 +36,19 @@ def unknown_tag(tag, parse_context)
...
+
+    def render(context)
+      @body.render(context)
+    end
+
+    private
+
+    def new_body
+      Liquid::BlockBody.new
+    end
```
Thibaut, line 52: This repeats the implementation in `Liquid::Block`. If the two are always intended to do the same thing / be overridden together, how about we expose a single implementation somewhere higher up? (for example, `Liquid.new_block_body`) This way Liquid C only needs to override one method.

dylanahsmith: Pushed a commit that extracted the implementation to override into `ParseContext#new_block_body`.  I'll update the liquid-c PR to override that.

Thibaut (approved):

pushrax (approved):

PR #1290, opened 2020-09-08
Pass the tag markup and tokenizer to Document#unknown_tag

dylanahsmith:
Depends on https://github.com/Shopify/liquid/pull/1289

## Problem

Shopify is currently overriding Liquid::Document#registered_tags to implement theme section tags, since they are only supposed to be available at the top-level of the section document.  However, `registered_tags` is actually defined in Liquid::BlockBody and https://github.com/Shopify/liquid/pull/1289 will make Liquid::Document no longer inherit from Liquid::BlockBody, so we need another way to enforce this constraint.

## Solution

Liquid already has builtin context sensitive tags, such as `elsif` and `else` within the `if` tag.  These are implemented using `Liquid::Block#unknown_tag`.  Liquid::Document also has an `unknown_tag` method, but it isn't provided the markup or tokenizer needed to the information we need, so the first commit of this pull request changes `Liquid::Document#unknown_tag` to take the same parameters as `Liquid::Block#unknown_tag`.  `Liquid::Document#parse` was also changed to be more like `Liquid::Block#parse` by continuing to parse if `unknown_tag` handles the tag without raising.

The second commit merely changes the `Liquid::Block#unknown_tag` parameter names for clarify since we use `markup` elsewhere to refer to the text within the tag after the tag name and the third parameter is actually a tokenizer.

### DIFF
--- a/lib/liquid/document.rb
+++ b/lib/liquid/document.rb
@@ -19,16 +19,15 @@ module Liquid
       @body.nodelist
     end
 
-    def parse(tokens, parse_context)
-      @body.parse(tokens, parse_context) do |end_tag_name, _end_tag_params|
-        unknown_tag(end_tag_name, parse_context) if end_tag_name
+    def parse(tokenizer, parse_context)
+      while parse_body(tokenizer)
       end
     rescue SyntaxError => e
       e.line_number ||= parse_context.line_number
       raise
     end
 
-    def unknown_tag(tag, parse_context)
+    def unknown_tag(tag, _markup, _tokenizer)
       case tag
       when 'else', 'end'
         raise SyntaxError, parse_context.locale.t("errors.syntax.unexpected_outer_tag", tag: tag)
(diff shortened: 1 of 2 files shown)


lib/liquid/document.rb
```diff
@@ -19,16 +19,15 @@ def nodelist
       @body.nodelist
     end
 
-    def parse(tokens, parse_context)
-      @body.parse(tokens, parse_context) do |end_tag_name, _end_tag_params|
-        unknown_tag(end_tag_name, parse_context) if end_tag_name
+    def parse(tokenizer, parse_context)
+      while parse_body(tokenizer)
```
Thibaut, line 23: Why are we adding a loop?

dylanahsmith: Because we don't want the handled unknown tag (e.g. section tags) to end the parsing of the document, we still need to parse the rest of the document.

Thibaut: Right, but why wasn't there a loop in the previous implementation? Sorry if I'm missing something obvious 😅 I reviewed expecting a refactor without change in behavior.

dylanahsmith: Previously `unknown_tag` was expected to always raise when called on a `Liquid::Document`, unlike `Liquid::Block` which has its `unknown_tag` method overridden by child classes which don't always raise.  This won't change any existing code, since `Liquid::Document#unknown_tag` itself still always raises.  It just means that now we can override `unknown_tag` in a child class of `Liquid::Document` and not raise to have it continue parsing.

Thibaut: >  It just means that now we can override `unknown_tag` in a child class of `Liquid::Document` and not raise to have it continue parsing.

Ahhh that explains it. 👍

Thibaut (approved):

samdoiron (approved):

PR #1292, opened 2020-09-10
Bring back silencing of errors in blank nodes for backwards compatibility

dylanahsmith:
## Problem

Template authors started to notice legitimate liquid errors after deploying https://github.com/Shopify/liquid/pull/1287 which previously weren't noticeable.  This was because we used to call `Liquid::BlockBody#render_node` using `render_node(context, node.blank? ? +'' : output, node)` for tags, where the throw-away `+''` string would have its errors written to it for blank tags.

## Solution

Add a conditional around the writing of the error message in the `Liquid::BlockBody#render_node` `rescue` block so blank tags don't have their errors written to the output again.

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -77,7 +77,10 @@ module Liquid
       raise
     rescue ::StandardError => e
       line_number = node.is_a?(String) ? nil : node.line_number
-      output << context.handle_error(e, line_number)
+      error_message = context.handle_error(e, line_number)
+      if node.instance_of?(Variable) || !node.blank? # conditional for backwards compatibility
+        output << error_message
+      end
     end
 
     private def parse_liquid_tag(markup, parse_context)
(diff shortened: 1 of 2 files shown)


pushrax (approved):
Hmm... users probably should be able to see these errors. But since we can't just change it, and it should still be possible to see the errors in the theme editor (I think?), 👍

dylanahsmith:
> it should still be possible to see the errors in the theme editor

Yes, the errors are still added to the context, so we could pull them out of the context to display them in the theme editor.

pushrax:
I was under the impression we already surface the errors in the theme editor.

dylanahsmith:
I tried saving `{% assign x = 0 %}{% if 1 < '2' %}{% assign x = 3 %}{% endif %}({{ x }})` on a theme and previewing it and it doesn't currently seem to show these render errors.  Perhaps you are thinking of parse errors.

samdoiron (approved):

PR #1294, opened 2020-09-16
Refactor to support liquid-c VM compilation of variables

dylanahsmith:
## Problem

In liquid-c, I would like to stop maintaining full compatibility with Liquid::BlockBody#nodelist, but right now liquid-c is being tested against both the liquid unit and integration tests, both of which have some coupling to that nodelist.  Specifically, for liquid-c compilation of variables into VM code (https://github.com/Shopify/liquid-c/pull/59) I wanted to avoid creating Liquid::Variable objects from the VM code, where we currently have direct node class comparison in the liquid tests.

Also, as part of having a different internal representation of Liquid::Variable, I would also like to use garbage collectable symbols for filter names in liquid-c, which a liquid integration test prevents.

## Solution

I moved the test for ParseTreeVisitor to the test/unit directory, since that interface is deeply coupled to the nodelist.  I also looked through the unit tests to find any that seem useful in testing liquid-c and converted some of them to integration tests.

I changed SecurityTest#test_does_not_add_filters_to_symbol_table to run the garbage collector before asserting that no new symbols were added, to exclude symbols that can be garbage collected.  However, that alone proved to be unreliable since MRI uses an imprecise GC to mark the C stack, which can cause deterministic failures for a given build of ruby, which I had actually debugged to confirm that this was the case. To make this reliable, I used a separate thread for the code that created the symbol, so the symbol VALUE would only be written on a separate stack that wouldn't be marked once the thread has finished.

I've also extracted some code from Liquid::BlockBody#render_node rescue code into a `rescue_render_node` method to re-use in liquid-c for simplicity and to reduce duplication, since it doesn't seem like it is on a hot code path.


lib/liquid/block_body.rb
```diff
@@ -71,15 +71,23 @@ def self.raise_missing_variable_terminator(token, parse_context)
...
-    rescue MemoryError
-      raise
-    rescue ::StandardError => e
-      line_number = node.is_a?(String) ? nil : node.line_number
-      error_message = context.handle_error(e, line_number)
-      if node.instance_of?(Variable) || !node.blank? # conditional for backwards compatibility
-        output << error_message
+    rescue => exc
+      blank_tag = !node.instance_of?(Variable) && node.blank?
+      rescue_render_node(context, output, node.line_number, exc, blank_tag)
```
macournoyer, line 76: Why not pass the `node` to `rescue_render_node` instead? Seems weird that the name is `rescue_render_node` and it's not receiving the `node`, but instead receiving only some parts.

test/integration/security_test.rb
```diff
@@ -43,15 +43,22 @@ def test_no_instance_eval_later_in_chain
     assert_equal(expected, Template.parse(text).render!(@assigns, filters: SecurityFilter))
   end
 
-  def test_does_not_add_filters_to_symbol_table
+  def test_does_not_permanently_add_filters_to_symbol_table
     current_symbols = Symbol.all_symbols
 
-    test = %( {{ "some_string" | a_bad_filter }} )
+    # MRI imprecisely marks objects found on the C stack, which can result
+    # in uninitialized memory being marked. This can even result in the test failing
```
macournoyer, line 50: That is weird. TIL!

macournoyer (approved):

lib/liquid/block_body.rb
dylanahsmith: Passing the `node` would require allocating that node in liquid-c, which I am trying to avoid.

macournoyer: Ah got it! Missed that you didn't have the node instance in C.

test/integration/security_test.rb
pushrax: MRI's solution to marking VALUE references in running C functions is to scan the stack/registers for things that look like VALUE pointers heuristically. Uninitialized memory sometimes looks like a VALUE.

pushrax (approved):

PR #1300, opened 2020-09-24
Only use MethodLiteral in condition expressions

dylanahsmith:
## Problem

The MethodLiteral literal value was added in https://github.com/Shopify/liquid/pull/592 which described it as a hacky solution:

> Not sure if this is a good fix, feels kinda hacky, but seems to work.

One way in which it is hacky is that it is only relevant for boolean comparison equality expressions and isn't relevant for all expressions.

In liquid-c (https://github.com/Shopify/liquid-c/pull/60), I would like Liquid::Expression.parse to return a Liquid::C::Expression object that can be rendered more efficiently and I would like to avoid coupling this work to these `MethodLiteral` objects.

## Solution

Add a `Liquid::Condition.parse_expression` method that wraps `Liquid::Expression.parse` so that it will return a `MethodLiteral` object for `empty` or `blank` literals and will otherwise delegate to `Liquid::Expression.parse`.  This way the values for these literals that are returned by `Liquid::Expression.parse` can just be the empty strings that `MethodLiteral#to_liquid` returns.

### DIFF
--- a/lib/liquid/condition.rb
+++ b/lib/liquid/condition.rb
@@ -27,8 +27,30 @@ module Liquid
       end,
     }
 
+    class MethodLiteral
+      attr_reader :method_name, :to_s
+
+      def initialize(method_name, to_s)
+        @method_name = method_name
+        @to_s = to_s
+      end
+
+      def to_liquid
+        to_s
+      end
+    end
+
+    @@method_literals = {
+      'blank' => MethodLiteral.new(:blank?, '').freeze,
+      'empty' => MethodLiteral.new(:empty?, '').freeze,
+    }
+
     def self.operators
       @@operators
+    end
+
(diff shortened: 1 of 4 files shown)


lib/liquid/condition.rb
```diff
@@ -27,10 +27,32 @@ class Condition #:nodoc:
...
 
+    class MethodLiteral
+      attr_reader :method_name, :to_s
+
+      def initialize(method_name, to_s)
+        @method_name = method_name
+        @to_s = to_s
+      end
+
+      def to_liquid
```
macournoyer, line 38: Since those now only appear in binary expressions in conditions, I don't think this method is necessary anymore.

macournoyer (approved):

tjoyal (approved):
LGTM

PR #1301, opened 2020-09-30
Fix missing freeze for array of static environments

dylanahsmith:
## Problem

Passing in an array of static environment hashes prevents those hashes from being frozen.  Also, the static environments hash also supports memoization of Proc hash keys

For example,

```ruby
env = { 'lazy' => -> { 1 } }
context = Liquid::Context.build(static_environments: [env])
p context['lazy']
```

outputs

```
1
{"lazy"=>1}
``` 

However, if the Liquid::Context was passed the hash directly as the `static_environments` (e.g. `Liquid::Context.build(static_environments: env)`) then the `env` hash is frozen and `context['lazy']` raises a `FrozenError` when trying to memoize the Proc result.

## Solution

Fix the freezing logic and add a test for the Proc memoization.

Note that this is a breaking change for Shopify's storefront, which is currently doing Proc memoization in the static environments hash.  So we should fix Shopify before merging this change.

We may also want to consider not freezing the static environments hash and just rely on the application to ensure that its Proc memoization in the static environment doesn't rely on mutable context state, since Proc memoization could still be useful to avoid redundant database queries.   It looks like our storefront rewrite wanted to be able to do this, so ended up using a second non-static environments hash as a static environment, along with a `Liquid::Context.new_isolated_subcontext` override to have this environment hash get inherited for `render` tag contexts.


celsodantas (approved):
tenor-3

pushrax (approved):
Since this will cause any proc set in `static_environments` to raise when accessed now, this will require a careful rollout as mentioned in the OP.

dylanahsmith:
Yeah, this is basically blocked until we deal with the breaking change in Shopify, especially since that could cause us to reconsider whether we want to actually freeze the static environment hashes.

Any thoughts on whether it makes sense to keep freezing the static environment hashes?

pushrax:
The proc memoization case does make sense to keep supporting. In theory elements in `static_environments` seem like they shouldn't be context-sensitive, so that would be fine. However, since in practise they easily can be context-sensitive, there is some bug-reducing benefit to completely disallowing it.

samdoiron (approved):

devdreamsolution (approved):

PR #1302, opened 2020-09-30
Fix test file with missing extension

dylanahsmith:
## Problem

@celsodantas pointed out that there is a test file with a missing file extension (`test/integration/context_test`).

Looks like I made this mistake in https://github.com/Shopify/liquid/commit/013802c8774e80fbcc78b52a51ce911aad5f25e4 when trying to move unit tests to become integration tests.

## Solution

There was already a test/integration/context_test.rb file, so I merged it with that.

### DIFF
--- a/test/integration/context_test.rb
+++ b/test/integration/context_test.rb
@@ -2,8 +2,595 @@
 
 require 'test_helper'
 
+class HundredCentes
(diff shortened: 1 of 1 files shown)


celsodantas (approved):
🚀

PR #1309, opened 2020-10-07
Combine chained `append` filters into a single `append_all`

macournoyer:
This PR introduces the `Optimizer` class that can only optimize one thing for now: combine `| append: 'a' | append: 'b'` into a single `| append_all: ['a', 'b']` filter call.

This avoids all the intermediate allocations caused by several `append`.

Also introduce a new filter `append_all: [*items]`. (Maybe I should modify the original `append` instead?)

### Context

We discovered some merchants are using `append` like so:

```
{%- for product in collection.products -%}
  {%- for variant in product.variants -%}
    [... snip ...]
    {% assign prod = productId | append: productSku | append: productTitle | append: productPrice | append: productVendor | append: productCategory | append: productHandle | append: productVariant %}
    [... snip ...]
  {%- endfor -%}
{%- endfor -%}
```

And it's causing big slowdowns, in part because of GC. So this is a first attempt at reducing memory usage.

### Benchmark

With the following code:
```
@context['s'] = "x" * 10240
appends = "| append: 'x' " * 10
Template.parse("{{ s #{appends} }}").render(@context)
```

Results:
```
     with optimizer:     1.021k i/100ms
  without optimizer:   849.000  i/100ms
     with optimizer:     10.199k (± 1.5%) i/s -    102.100k in  10.013380s
  without optimizer:      8.523k (± 2.8%) i/s -     85.749k in  10.069714s

Comparison:
     with optimizer::    10198.7 i/s
  without optimizer::     8522.6 i/s - 1.20x  (± 0.00) slower
```

### Memory

Using `MemoryProfiler` with 1000 runs of code above:

/  | Allocated memory by `standardfilters.rb`
-- | --------------------
with optimizer | 10331000 bytes
without optimizer | 102865000 bytes

About 10x less memory

### DIFF
--- a/lib/liquid/optimizer.rb
+++ b/lib/liquid/optimizer.rb
@@ -2,54 +2,18 @@
 
 module Liquid
   class Optimizer
-    def optimize(node)
-      case node
-      when Liquid::Template       then optimize(node.root.body)
-      when Liquid::Document       then optimize(node.body)
-      when Liquid::BlockBody      then optimize_block(node)
-      when Liquid::Variable       then optimize_variable(node)
-      when Liquid::Assign         then optimize_assign(node)
-      when Liquid::For            then optimize_for(node)
-      when Liquid::If             then optimize_if(node)
-      end
-      node
-    end
+    class << self
+      attr_accessor :enabled
 
-    def optimize_block(block)
-      block.nodelist.each { |node| optimize(node) }
-    end
+      def optimize_variable(node)
+        return unless enabled
 
-    def optimize_variable(node)
-      # Turn chained `| append: "..."| append: "..."`, into a single `append_all: [...]`
-      if node.filters.size > 1 && node.filters.all? { |f, _| f == "append" }
-        node.filters = [["append_all", node.filters.map { |f, (arg)| arg }]]
+        # Turn chained `| append: "..."| append: "..."`, into a single `append_all: [...]`
+        if node.filters.size > 1 && node.filters.all? { |f, _| f == "append" }
+          node.filters = [["append_all", node.filters.map { |f, (arg)| arg }]]
+        end
       end
     end
-
-    def optimize_assign(node)
-      optimize(node.from)
-    end
-
-    def optimize_for(node)
-      optimize(node.collection_name)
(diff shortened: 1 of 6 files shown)


lib/liquid/optimizer.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def optimize_block(block)
+      block.nodelist.each { |node| optimize(node) }
+    end
+
+    def optimize_variable(node)
+      # Turn chained `| append: "..."| append: "..."`, into a single `append_all: [...]`
+      if node.filters.size > 1 && node.filters.all? { |f, _| f == "append" }
+        node.filters = [["append_all", node.filters.map { |f, (arg)| arg }]]
+      end
```
michaelmelanson, line 26: What about cases where the list of `append`s aren't the _entire_ list of filters, like `{{something | some_drop | append: "..." | append: "..." | another_drop }}`? Are they worth handling?

macournoyer: Yeah! We could combine the consecutive one too. It's just more complex to handle.

macournoyer:
Discussed w/ @tjoyal. We want to avoid adding new filters or modifying existing ones. Maybe it could be possible to optimize using the existing `join` filter instead?

lib/liquid/variable.rb
```diff
@@ -31,6 +31,8 @@ def initialize(markup, parse_context)
       @line_number   = parse_context.line_number
 
       parse_with_selected_parser(markup)
+
+      Optimizer.optimize_variable(self)
```
dylanahsmith, line 35: Note that this will be skipped when compiling with liquid-c if it passes strict parsing without a syntax error.

colinbendell: can you elaborate?

pushrax: liquid-c won't allocate the Variable instance at all.

I guess we'll probably want to duplicate the optimizer inside C. It's probably not practical to have a single implementation for both.

macournoyer:
Closing in favor of https://github.com/Shopify/liquid-c/pull/82

PR #1312, opened 2020-10-09
Raise when source is too large

peterzhu2118:
Raise when source is larger than 16MB in `Liquid::Template#parse`. This will let us use this assumption in liquid-c.

See Shopify/liquid-c#79.

### DIFF
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -15,6 +15,8 @@ module Liquid
   #   template.render('user_name' => 'bob')
   #
   class Template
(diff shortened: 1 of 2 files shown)


dylanahsmith:
I don't think this limit makes sense when using liquid without liquid-c.

peterzhu2118:
That's true. Do you think this should this be a patch in liquid-c instead?

dylanahsmith:
> That's true. Do you think this should this be a patch in liquid-c instead?

yes

PR #1321, opened 2020-10-20
Extract method for raising a syntax error in the assign tag for liquid-c

dylanahsmith:
I would like to compile the assign tag directly from C code in liquid-c, but want to delegate the raising of the syntax error to ruby code.  In order to avoid duplication, it makes sense to extract the raise method in this repo so that it can be re-used.

### DIFF
--- a/lib/liquid/tags/assign.rb
+++ b/lib/liquid/tags/assign.rb
@@ -12,15 +12,20 @@ module Liquid
   class Assign < Tag
     Syntax = /(#{VariableSignature}+)\s*=\s*(.*)\s*/om
 
+    # @api private
+    def self.raise_syntax_error(parse_context)
+      raise Liquid::SyntaxError, parse_context.locale.t('errors.syntax.assign')
+    end
+
     attr_reader :to, :from
 
-    def initialize(tag_name, markup, options)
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):

ashmaroli:
@dylanahsmith You could use `private_class_method` to *enforce* the private api here:
```diff
-   # @api private
    def self.raise_syntax_error(parse_context)
      raise Liquid::SyntaxError, parse_context.locale.t('errors.syntax.assign')
    end
+   private_class_method :raise_syntax_error
```
That said, I must admit that I don't know if there's an equivalent for class methods defined at the C side.

dylanahsmith:
If it were private, then `self.class.raise_syntax_error` would raise as can be seen from running the tests

```
  1) Failure:
AssignTest#test_assign_syntax_error [/Users/dylants/src/liquid/test/integration/assign_test.rb:35]:
[Liquid::SyntaxError] exception expected, not
Class: <NoMethodError>
Message: <"private method `raise_syntax_error' called for Liquid::Assign:Class">
---Backtrace---
/Users/dylants/src/liquid/lib/liquid/tags/assign.rb:29:in `initialize'
```

PR #1329, opened 2020-10-22
Reduce empty string allocations via string filters

ashmaroli:
## Background

Casting `nil` into a string allocates a new String instance `""` which when passed to `String#+` allocates another instance of empty string.
Assigning a frozen string to the operand when it is `nil` avoids the allocation of the temporary empty string.

## Scenarios

multiple occurrences of either of the following templates where `label` will always evaluate to `nil`:
```
{{ product.title | prepend: label }}
```
or
```
{{ label | append: product.title }}
```

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -276,8 +276,19 @@ module Liquid
       input.to_s.sub(string.to_s, '')
     end
 
+    # prepend a string to another
+    def prepend(input, string)
+      # assign empty frozen string to avoid generating a new String via
+      # `nil.to_s` only to be duplicated via `String#+`
+      string = "" if string.nil?
+      string.to_s + input.to_s
+    end
+
     # add one string to another
     def append(input, string)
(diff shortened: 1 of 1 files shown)


dylanahsmith:
> Casting `nil` into a string allocates a new String instance `""`

It looks like ruby 2.7 changed `nil.to_s` to return a constant/frozen string to avoid these allocations, so I don't think this PR is needed

ashmaroli:
> It looks like ruby 2.7 changed `nil.to_s` to return a constant/frozen string to avoid these allocations

Okay, but since the next Liquid release supports at least Ruby 2.5, isn't this proposal still applicable..?

dylanahsmith:
The right level of abstraction to fix this problem was in ruby, which is why it was fixed there.  I don't see why we should workaround this problem for older ruby versions at the expense of the newer ruby versions.  I.e. `string = "" if string.nil?` isn't free on ruby 2.7 just because it doesn't cause an object allocation.

PR #1331, opened 2020-10-26
Freeze block body after parsing completes

peterzhu2118:
Adds `BlockBody#freeze` that freezes the `BlockBody` to prevent modifications. Additionally, only frozen `BlockBody` can be rendered. Used in Shopify/liquid-c#102.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,6 @@ group :test do
   gem 'rubocop-performance', require: false
 
   platform :mri, :truffleruby do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'pz-block-body-buffer'
+    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'master'
   end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,6 @@ group :test do
   gem 'rubocop-performance', require: false
 
   platform :mri, :truffleruby do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'master'
+    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'pz-block-body-buffer'
   end
 end
--- a/lib/liquid/block.rb
+++ b/lib/liquid/block.rb
@@ -11,7 +11,7 @@ module Liquid
 
     def parse(tokens)
       @body = new_body
-      while parse_body(@body, tokens, partial: true)
+      while parse_body(@body, tokens)
       end
       @body.freeze
     end
@@ -68,9 +68,7 @@ module Liquid
     end
 
     # @api public
-    def parse_body(body, tokens, partial: false)
-      block_terminated = true
-
+    def parse_body(body, tokens)
       if parse_context.depth >= MAX_DEPTH
         raise StackLevelError, "Nesting too deep"
       end
@@ -79,10 +77,7 @@ module Liquid
         body.parse(tokens, parse_context) do |end_tag_name, end_tag_params|
           @blank &&= body.blank?
 
-          if end_tag_name == block_delimiter
-            block_terminated = false
-            next
-          end
+          return false if end_tag_name == block_delimiter
           raise_tag_never_closed(block_name) unless end_tag_name
 
           # this tag is not registered with the system
@@ -93,12 +88,7 @@ module Liquid
         parse_context.depth -= 1
       end
 
-      unless partial
-        body.remove_blank_strings if body.blank?
-        body.freeze
-      end
-
-      block_terminated
+      true
     end
   end
 end
--- a/lib/liquid/block.rb
+++ b/lib/liquid/block.rb
@@ -11,9 +11,9 @@ module Liquid
 
     def parse(tokens)
       @body = new_body
-      while parse_body(@body, tokens)
+      while parse_body(@body, tokens, partial: true)
       end
-      @body.freeze(parse_context)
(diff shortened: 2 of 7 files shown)


lib/liquid/tags/case.rb
```diff
@@ -20,7 +20,11 @@ def initialize(tag_name, markup, options)
 
     def parse(tokens)
       body = new_body
-      body = @blocks.last.attachment while parse_body(body, tokens)
+      while parse_body(body, tokens)
+        body.freeze(parse_context)
```
dylanahsmith, line 24: Block body freezing seems like something that should be handled in Liquid::Block itself, such that it doesn't require changes to these individual block tags.

These block tag classes that call `parse_body` all just freeze the body immediately afterwards, so why don't we have `parse_body` freeze the block body by default.  The one use case for partially parsing the body seems to be in Liquid::Block#parse itself, where it could call pass a `partial: true` keyword argument to avoid freezing the block body, then can do the manual `@body.freeze` at the end of Liquid::Block#parse that you are already doing.

Gemfile
```diff
@@ -22,6 +22,6 @@ group :test do
   gem 'rubocop-performance', require: false
 
   platform :mri, :truffleruby do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'master'
+    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'pz-block-body-buffer'
```
dylanahsmith, line 25: It doesn't seem like this pull request should depend on the liquid-c one.  It seems like that creates a circular dependency between the two pull requests

dylanahsmith (changes requested):

lib/liquid/block_body.rb
```diff
@@ -192,6 +199,8 @@ def render(context)
     end
 
     def render_to_output_buffer(context, output)
+      raise "Can only render when frozen" unless @frozen
```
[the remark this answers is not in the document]
peterzhu2118: That a fair point. I think that I want to make sure that `liquid` and `liquid-c` implementations are consistent, so I'm thinking of replacing that line with:

```suggestion
      freeze unless frozen?
```

How does that sound?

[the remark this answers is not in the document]
peterzhu2118: I think this might be an issue with Shopify/liquid-c#102 when we pool the `vm_assembler_t`. We should release the pool after parsing (it would be wasted memory otherwise). But this would cause an use-after-free because we still need certain elements of the pool in the `BlockBody`. So the logic for releasing the pool has to be more complex (make sure none of the `BlockBody` is still holding a reference to any elements in the pool). But if we ensure that everything is frozen (and written to the buffer) before rendering, then we can safely free the pool after parsing.

lib/liquid/tags/case.rb
peterzhu2118: The original implementation (the one you just reviewed), the `BlockBody#freeze` method was not a true freeze, so `BlockBody#remove_blank_strings` was allowed to run even after freezing (this is not a problem in Shopify/liquid-c#86 because it replaces the strings with a jump instruction, which does not change the size of any data structures). However, now that `BlockBody#freeze` really does freeze the object, we can no longer call `BlockBody#remove_blank_strings` afterwards. So I've changed `Block#parse_body` to have an additional side-effect of calling `BlockBody#remove_blank_strings` before it calls `BlockBody#freeze`. I'm not 100% sure this is fine to do.

lib/liquid/block_body.rb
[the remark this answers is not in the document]
dylanahsmith: If you have ensured that everything is frozen after parsing, then it doesn't seem like you would need this check.  In that case this would only be making sure we aren't rendering during parsing, which doesn't seem like it is worth worrying about here, where it isn't a safety concern.

I think the safety checks would end up being similar either way in liquid-c.  You are still going to need to worry about references from a `Liquid::C::BlockBody` to a VM assembler pool before releasing that pool, unless you can ensure that the reference won't be used again.  That applies to more than just rendering, but also making sure you don't try to use that freed memory if freeze gets called on that `Liquid::C::BlockBody` after parsing.

lib/liquid/block.rb
```diff
@@ -87,7 +93,12 @@ def parse_body(body, tokens)
         parse_context.depth -= 1
       end
 
-      true
+      unless partial
+        body.remove_blank_strings if body.blank?
```
dylanahsmith, line 97: Oh right, we can't really do the freezing in `parse_body` because we don't know whether to remove blank strings at this point.  We should only do this if the whole block tag is blank, which we don't really know until we finish parsing the whole tag.  I think that means that we can't really freeze the block bodies until after parsing the whole tag either, at least for these control flow tags.

ashmaroli:
@peterzhu2118 What do you think about reducing the iterations in the `case` tag with following:
```diff
-      if blank?
-        @blocks.each { |condition| condition.attachment.remove_blank_strings }
-      end
-      @blocks.each { |condition| condition.attachment.freeze }
+      @blocks.each do |condition|
+        condition.attachment.remove_blank_strings if blank?
+        condition.attachment.freeze
+      end
```

dylanahsmith:
> What do you think about reducing the iterations in the `case` tag

Yeah, that makes sense.  We can do the same thing in the `if` tag.

dylanahsmith (approved):

PR #1333, opened 2020-10-26
Add Liquid::ParseContext#parse_expression for liquid-c node disabling

dylanahsmith:
## Problem

We would like to be able to measure the performance benefit of liquid VM rendering in production by disabling liquid VM compilation at runtime.  However, tags are parsing expressions with `Liquid::Expression.parse`, which doesn't give liquid-c access to the parse context in order to determine whether liquid VM compilation is enabled or not.

## Solution

Add `Liquid::ParseContext#parse_expression` and use it in the liquid tags to parse expressions.

### DIFF
--- a/lib/liquid/condition.rb
+++ b/lib/liquid/condition.rb
@@ -45,8 +45,8 @@ module Liquid
       @@operators
     end
 
-    def self.parse_expression(markup)
-      @@method_literals[markup] || Expression.parse(markup)
+    def self.parse_expression(parse_context, markup)
+      @@method_literals[markup] || parse_context.parse_expression(markup)
     end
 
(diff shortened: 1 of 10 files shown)


macournoyer (approved):
Does this affect profiling too? I guess it's in preparation for https://github.com/Shopify/liquid-c/pull/96.

dylanahsmith:
No, this shouldn't affect profiling.  Although, it might help better test that path.

It isn't a blocker for Shopify/liquid-c#96.  I've added a comment to that PR (https://github.com/Shopify/liquid-c/pull/96#issuecomment-717239273) to explain why I haven't merged it yet.

peterzhu2118 (approved):

PR #1335, opened 2020-10-27
Fix lax parsing expressions surrounded by spaces

### LINKED ISSUE #1334: Can no longer have spaces between brackets when accessing object properties in an assign statement.
Liquid can no longer have spaces between brackets when accessing object properties in an assign statement.

This code used to work and now doesn't:
```liquid
{%- assign mbarak_font_setting_value = settings[ 'header_top_nav_font_stack' ] -%}
{{ mbarak_font_setting_value }}
```
Note the spaces between the square brackets on the `settings` object.

This code use to return the value held by the object at the key provided but now returns the whole object - in this case `settings_data.json`.

I'm not sure if this issue is in [liquid](https://github.com/Shopify/liquid) or [liquid-c](https://github.com/Shopify/liquid-c)

dylanahsmith:
Fixes #1334

## Problem

In liquid-c, we chose to try to do strict parsing, then fallback to lax parsing if there is a strict parse failure.  This means that the strict parser and lax parser should behave identically if there are no strict parse errors, however, this isn't the case for spaces in expression literals.  For example, spaces surrounding keyword literals, strings, integers, floats or ranges end up getting stripped/ignored by the strict parser.  Spaces were also supported between tokens in ranges in the strict parser.

## Solution

Extend the Liquid::Expression regexes to allow spaces in the same way as the strict parser.

Since the LITERAL hash is looked up without a regex, I changed it to strip the markup string before the lookup.

I added an integration test class to test these expressions so they should be tested against the strict parser, lax parser and liquid-c's strict parser.  These tests passing in liquid-c also show that this incompatibility existed there.

### DIFF
--- a/lib/liquid/expression.rb
+++ b/lib/liquid/expression.rb
@@ -10,25 +10,28 @@ module Liquid
       'empty' => ''
     }.freeze
 
-    SINGLE_QUOTED_STRING = /\A'(.*)'\z/m
-    DOUBLE_QUOTED_STRING = /\A"(.*)"\z/m
-    INTEGERS_REGEX       = /\A(-?\d+)\z/
-    FLOATS_REGEX         = /\A(-?\d[\d\.]+)\z/
-    RANGES_REGEX         = /\A\((\S+)\.\.(\S+)\)\z/
+    SINGLE_QUOTED_STRING = /\A\s*'(.*)'\s*\z/m
+    DOUBLE_QUOTED_STRING = /\A\s*"(.*)"\s*\z/m
+    INTEGERS_REGEX       = /\A\s*(-?\d+)\s*\z/
+    FLOATS_REGEX         = /\A\s*(-?\d[\d\.]+)\s*\z/
+    RANGES_REGEX         = /\A\s*\(\s*(\S+)\s*\.\.\s*(\S+)\s*\)\s*\z/
 
     def self.parse(markup)
-      if LITERALS.key?(markup)
-        LITERALS[markup]
+      case markup
+      when nil
+        nil
+      when SINGLE_QUOTED_STRING, DOUBLE_QUOTED_STRING
+        Regexp.last_match(1)
+      when INTEGERS_REGEX
+        Regexp.last_match(1).to_i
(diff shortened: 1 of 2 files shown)


dylanahsmith:
cc @mbarak

test/integration/expression_test.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  private
+
+  def parse_and_eval(markup, **assigns)
+    if Liquid::Template.error_mode == :strict
+      p = Liquid::Parser.new(markup)
+      markup = p.expression
+      p.consume(:end_of_string)
+    end
```
peterzhu2118, line 41: I'm not sure why this is needed? Tests seem to pass without it.

lib/liquid/expression.rb
```diff
@@ -10,25 +10,28 @@ class Expression
...
-        case markup
-        when SINGLE_QUOTED_STRING, DOUBLE_QUOTED_STRING
-          Regexp.last_match(1)
-        when INTEGERS_REGEX
-          Regexp.last_match(1).to_i
-        when RANGES_REGEX
-          RangeLookup.parse(Regexp.last_match(1), Regexp.last_match(2))
-        when FLOATS_REGEX
-          Regexp.last_match(1).to_f
+        markup = markup.strip
```
peterzhu2118, line 32: If we do `markup = markup.strip` at the beginning of the method I think we can remove the need for matching `\s*` in the regexes.

test/integration/expression_test.rb
dylanahsmith: This is to test ruby strict parsing.  Otherwise, the tests would only be testing lax parsing, which wouldn't help ensure they are consistent.

lib/liquid/expression.rb
dylanahsmith: We can't call `strip` until we handle the `nil` case which used to be handled by the `LITERALS.key?` condition. We are going to need `\s*` within the range regex either way.

peterzhu2118 (approved):

PR #1337, opened 2020-10-28
Parse only string expression markups

ashmaroli:
In the *unlikely event* that a `Liquid::Expression` markup is neither `nil`, nor `String` (say it is `false` or `true` or something else), then bail out immediately especially since `markup = markup.strip` will fail if `markup` isn't a string.

Additionally, since the conditional expressions in the `else` branch of the `case` statement are rather short, I refactored the conditional expression into a ternary expression.

### DIFF
--- a/lib/liquid/expression.rb
+++ b/lib/liquid/expression.rb
@@ -17,9 +17,9 @@ module Liquid
     RANGES_REGEX         = /\A\s*\(\s*(\S+)\s*\.\.\s*(\S+)\s*\)\s*\z/
 
     def self.parse(markup)
+      return markup unless markup.is_a?(String)
+
       case markup
-      when nil
-        nil
       when SINGLE_QUOTED_STRING, DOUBLE_QUOTED_STRING
         Regexp.last_match(1)
       when INTEGERS_REGEX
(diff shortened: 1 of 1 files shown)


pushrax (commented):
Is there a specific bug this is fixing, or is the goal to make the code more defensive?

ashmaroli:
> the goal to make the code more defensive?

That's correct.
Because `Liquid::Expression.parse(markup)` being public API can be called using any Ruby object as argument in a third-party program.

dylanahsmith:
I don't think we ever supported it being called with any ruby object.  I think it is an application bug if it is calling it with a non-String object.  Although, we could make this clearer using documentation or by starting to use [RBS](https://github.com/ruby/rbs) to annotate the types.

ashmaroli:
> I don't think we ever supported it being called with any ruby object.

Okay. I still don't see how the proposed change goes against that principle.


> we could make this clearer using documentation or by starting to use [RBS](https://github.com/ruby/rbs) to annotate the types

Isn't the proposed change achieving the same with lesser effort?

dylanahsmith:
Documentation and type annotations will make sense either way, so this isn't a replacement for that.  In fact, this would make the documentation and type annotation more complicated to support this extended behaviour.

If the goal is to make the application bug clearer, then we could raise a TypeError with a helpful error message.  That way the code is compatible with liquid-c, rather than being an inconsistency with liquid-c.

ashmaroli:
Over to you then, @dylanahsmith :)

PR #1343, opened 2020-11-05
Test invalid floats without integer parts

peterzhu2118:
Add tests for Shopify/liquid-c#120.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,6 @@ group :test do
   gem 'rubocop-performance', require: false
 
   platform :mri, :truffleruby do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'master'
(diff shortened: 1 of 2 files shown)


kpy815 (approved):

devdreamsolution (approved):

dylanahsmith:
Based on https://github.com/Shopify/liquid-c/pull/123, it looks like we actually have themes that depend on this liquid quirk.

We might want to consider just supporting that syntax.  It isn't actually that unreasonable.  For example, it is valid syntax in python

```
$ python
Python 3.10.3 (main, Jul  4 2022, 12:12:23) [Clang 13.1.6 (clang-1316.0.21.2.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> -.1
-0.1
```

It is also considered acceptable syntax in math, even though the `0` preceding the decimal point is typically preferred for readability.

PR #1344, opened 2020-11-05
Test space between dot for attributes

peterzhu2118:
Tests for Shopify/liquid-c#122.


dylanahsmith (approved):
LGTM once https://github.com/Shopify/liquid-c/pull/122 is merged and the Gemfile change has been reverted back to using liquid-c master

PR #1346, opened 2020-11-06
Fix regex for matching endraw tags

peterzhu2118:
Liquid does not accept `{% endraw{% f %}` as a valid `endraw` tag but accepts `{% endfor{% f %}` for `endfor` tags. This is because the second capture group (i.e. `(\w+)`) matches on the `f` instead of the `endraw`.

The solution I implemented hardcodes the `endraw` tag into the regex. The only way I can think of that doesn't hardcode the `endraw` in the regex is to dynamically generate it in `Liquid::Raw#parse`, which would not be very efficient.

Note: the tests fail because `liquid-c` hasn't been updated yet.

Another problem is that the old regex would match `{% endraw{% |f %}` as a valid `endraw` tag because `|f` no longer matches the `(\w+)` capture group so it would match on `endraw`. But this fails in liquid-c because it would match on the `f`.

### DIFF
--- a/lib/liquid/tags/raw.rb
+++ b/lib/liquid/tags/raw.rb
@@ -3,7 +3,7 @@
 module Liquid
   class Raw < Block
     Syntax = /\A\s*\z/
-    FullTokenPossiblyInvalid = /\A(.*)#{TagStart}\s*(\w+)\s*(.*)?#{TagEnd}\z/om
+    EndRawTag = /\A(.*)#{TagStart}\s*endraw.*?#{TagEnd}\z/om
 
     def initialize(tag_name, markup, parse_context)
       super
@@ -14,7 +14,7 @@ module Liquid
     def parse(tokens)
       @body = +''
       while (token = tokens.shift)
-        if token =~ FullTokenPossiblyInvalid && block_delimiter == Regexp.last_match(2)
+        if token =~ EndRawTag
           @body << Regexp.last_match(1) if Regexp.last_match(1) != ""
           return
         end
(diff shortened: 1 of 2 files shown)


dylanahsmith:
> Liquid does not accept `{% endraw{% f %}` as a valid `endraw` tag but accepts `{% endfor{% f %}` for `endfor` tags. This is because the second capture group (i.e. `(\w+)`) matches on the `f` instead of the `endraw`.

Ideally, we wouldn't accept `{% endraw{% f %}` or `{% endfor{% f %}` if there isn't liquid code that depends on this.  I think we could add Liquid::Usage instrumentation to see if either of these are needed.  For the raw tag, this only recently became possible when it was implemented in liquid-c, so there is a good chance we don't need to support this.  We should use a separate Liquid::Usage.instrument name for block delimiters in general, since this lax behaviour has been around for longer, so has a higher chance of being relied upon.

lib/liquid/tags/raw.rb
```diff
@@ -3,7 +3,7 @@
 module Liquid
   class Raw < Block
     Syntax = /\A\s*\z/
-    FullTokenPossiblyInvalid = /\A(.*)#{TagStart}\s*(\w+)\s*(.*)?#{TagEnd}\z/om
+    EndRawTag = /\A(.*)#{TagStart}\s*endraw.*?#{TagEnd}\z/om
```
dylanahsmith, line 6: We have tags that inherit from Liquid::Raw, so will need this to work with other delimiters if we do need to support garbage characters after the delimiter name

devdreamsolution (approved):

PR #1350, opened 2020-11-09
Instrument invalid end tags

peterzhu2118:
Add instrumentation for bug #1346. It increments `multiple_open_tags` every time the `Template` has a source with a tag that has multiple open tags before a close tag. If we know nobody is using syntax like `{% endfor {% foo %}`, then we can safely treat it as invalid syntax.

### DIFF
--- a/lib/liquid/block.rb
+++ b/lib/liquid/block.rb
@@ -77,6 +77,9 @@ module Liquid
         body.parse(tokens, parse_context) do |end_tag_name, end_tag_params|
           @blank &&= body.blank?
 
+          # Instrument for bug 1346
+          Usage.increment("end_tag_params") if end_tag_params && !end_tag_params.empty?
+
           return false if end_tag_name == block_delimiter
           raise_tag_never_closed(block_name) unless end_tag_name
 
--- a/test/integration/block_test.rb
+++ b/test/integration/block_test.rb
@@ -55,4 +55,24 @@ class BlockTest < Minitest::Test
       assert_equal buf.object_id, output.object_id
     end
   end
+
+  def test_instrument_for_bug_1346
+    calls = []
+    Liquid::Usage.stub(:increment, ->(name) { calls << name }) do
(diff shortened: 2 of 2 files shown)


devdreamsolution (approved):

lib/liquid/template.rb
```diff
@@ -212,6 +213,12 @@ def render_to_output_buffer(context, output)
 
     private
 
+    BUG_1346_REGEX = /#{TagStart}#{WhitespaceControl}?\s*\w+\s*#{TagStart}/
+
+    def instrument_for_bug_1346(source)
+      Usage.increment("multiple_open_tags") if BUG_1346_REGEX.match?(source)
```
dylanahsmith, line 219: This seems too specific to the specific example.  More generally, we want `end_tag_params` to be empty on the `if end_tag_name == block_delimiter` branch in Liquid::Block.parse_body (https://github.com/Shopify/liquid/blob/c7c21e88f018af6a02095cd8161d065d2967ea21/lib/liquid/block.rb#L80) since any end tag markup there would get ignored, whether that is `{% f` or any other non-whitespace characters.

test/integration/template_test.rb
```diff
@@ -323,4 +323,30 @@ def test_using_range_literal_works_as_expected
...
+    Liquid::Usage.stub(:increment, ->(name) { calls << name }) do
+      Liquid::Template.parse("{% for i in (1..2) %}{{ i }}{% endfor %}")
+    end
+    assert_equal([], calls)
+
+    calls = []
+    Liquid::Usage.stub(:increment, ->(name) { calls << name }) do
+      Liquid::Template.parse("{% for i in (1..2) %}{{ i }}{% endfor test %}")
+    end
+    assert_equal([], calls)
```
dylanahsmith, line 350: This is an example of the sort of thing we want to catch with the instrumentation and is the type of thing that it looks like we accidentally started allowing for the raw tag in liquid-c.

dylanahsmith (approved):

PR #1358, opened 2020-11-13
Simplify RANGES_REGEX to avoid backtracking issues

### LINKED ISSUE #1357: Catastrophic backtracking in `RANGES_REGEX` regular expression
Hello team! I contacted the security team and they told me to come here.

## Summary

The `RANGES_REGEX` regular expression in [`expression.rb`](https://github.com/Shopify/liquid/blob/9fcba1a26c49788c9de92eee8bfef00141d91749/lib/liquid/expression.rb#L17) has a [catastrophic backtracking](https://www.regular-expressions.info/catastrophic.html) issue when provided with a specific malicious input. This exposes websites that accept Liquid user input to excessive resource consumption and possibly impact availability if the process isn't running with a timeout.

## Details

The `RANGES_REGEX` is defined as such: `/\A\s*\(\s*(\S+)\s*\.\.\s*(\S+)\s*\)\s*\z/`.

It's meant to match things of the format `(1..10)` or `variable.property..100` and it also allows any amount of whitespace (`  (     1   ..                  10 )`). The problem here is that the `\S` character classes include the `.` character so we're effectively left with a regex that, for the attacker, is equivalent to `\A\(\.+\.\.\.+\)\z`. This kind of regular expression has execution time that grows quadratically with input length so a very long sequence of `.` with a character that isn't a `.` at the end with send the CPU spinning until the end of times.

### Jekyll

The easiest repro to demonstrate the issue and possibly the most widespread use is with a jekyll project

```shell
jekyll new repro
cd repro
ruby -e 'puts "{% for i in (#{"."*1000000})! %}"' >> about.markdown
echo '{% endfor %}' >> about.markdown
jekyll build
```

At this point the build process hangs using 100% CPU until it is killed.

## Impact

This exposes websites that accept Liquid user input to excessive resource consumption and possibly impact availability if the process isn't running with a timeout.

dee-see:
Fixes #1357

I replaced the regular expression with a much simpler one and then do the rest of the heavy lifting with a split and some basic regular expressions.

I also added a couple of tests.

### DIFF
--- a/lib/liquid/expression.rb
+++ b/lib/liquid/expression.rb
@@ -14,7 +14,7 @@ module Liquid
     DOUBLE_QUOTED_STRING = /\A\s*"(.*)"\s*\z/m
     INTEGERS_REGEX       = /\A\s*(-?\d+)\s*\z/
     FLOATS_REGEX         = /\A\s*(-?\d[\d\.]+)\s*\z/
-    RANGES_REGEX         = /\A\s*\(\s*(\S+)\s*\.\.\s*(\S+)\s*\)\s*\z/
+    RANGES_REGEX         = /\A\s*\(.*\)\s*\z/
 
     def self.parse(markup)
       case markup
@@ -25,7 +25,8 @@ module Liquid
       when INTEGERS_REGEX
         Regexp.last_match(1).to_i
       when RANGES_REGEX
-        RangeLookup.parse(Regexp.last_match(1), Regexp.last_match(2))
+        parts = markup.strip.gsub(/\A\(|\)\Z/, '').split(/\.{2,}/)
+        return RangeLookup.parse(parts[0].strip, parts[1].strip) if parts.count == 2
       when FLOATS_REGEX
         Regexp.last_match(1).to_f
       else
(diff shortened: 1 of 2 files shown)


lib/liquid/expression.rb
```diff
@@ -25,7 +25,8 @@ def self.parse(markup)
       when INTEGERS_REGEX
         Regexp.last_match(1).to_i
       when RANGES_REGEX
-        RangeLookup.parse(Regexp.last_match(1), Regexp.last_match(2))
+        parts = markup.strip.gsub(/\A\(|\)\Z/, '').split(/\.{2,}/)
```
dee-see, line 28: `/\.{2,}/` because the old regex supported `(1............2)` and there's actually a test to make sure this "quirk" is supported.

dee-see:
I signed the CLA but I can't figure out how to rerun the check :upside_down_face:

dylanahsmith:
> I signed the CLA but I can't figure out how to rerun the check 🙃

Re-ran it for you

dee-see:
https://github.com/Shopify/liquid/pull/1360 was used instead

PR #1366, opened 2020-11-19
Create top-level profile timing nodes for multiple template renders

dylanahsmith:
## Problem

https://github.com/Shopify/liquid/pull/1365 allows multiple renders to be profiled using the same profiler.  However, all the top-level timing nodes for the profiler are not grouped together for each render.  This also makes it harder to quickly find out how much time was spent with each render.

## Solution

I kept the existing behaviour for when there is only a single render on a profiler object.  If there are multiple template renders being profiled, then top-level timing objects are returned that group together each render.

To do this, I've decoupled the `profiler_node` method and `Timing` class from the Tag/Variable node objects, so we can easily create these top-level Timing objects for the template renders.

### DIFF
--- a/lib/liquid/profiler.rb
+++ b/lib/liquid/profiler.rb
@@ -46,97 +46,91 @@ module Liquid
     include Enumerable
 
     class Timing
-      attr_reader :code, :partial, :line_number, :children, :total_time, :self_time
+      attr_reader :code, :template_name, :line_number, :children
+      attr_accessor :total_time
       alias_method :render_time, :total_time
+      alias_method :partial, :template_name
 
-      def initialize(node, template_name)
-        @code        = node.respond_to?(:raw) ? node.raw : node
-        @partial     = template_name
-        @line_number = node.respond_to?(:line_number) ? node.line_number : nil
-        @children    = []
+      def initialize(code: nil, template_name: nil, line_number: nil)
+        @code = code
+        @template_name = template_name
+        @line_number = line_number
+        @children = []
       end
 
-      def self.start(node, template_name)
-        new(node, template_name).tap(&:start)
-      end
-
-      def start
-        @start_time = monotonic_time
-      end
-
-      def finish
-        @total_time = monotonic_time - @start_time
-
-        if @children.empty?
-          @self_time = @total_time
-        else
-          total_children_time = 0
+      def self_time
+        @self_time ||= begin
+          total_children_time = 0.0
           @children.each do |child|
             total_children_time += child.total_time
           end
-          @self_time = @total_time - total_children_time
+          @total_time - total_children_time
         end
       end
+    end
 
-      private
+    attr_reader :total_time
+    alias_method :total_render_time, :total_time
 
-      def monotonic_time
-        Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    def initialize
+      @root_children = []
+      @current_children = nil
+      @total_time = 0.0
+    end
+
+    def profile(template_name, &block)
+      # nested renders are done from a tag that already has a timing node
+      return yield if @current_children
+
+      root_children = @root_children
+      render_idx = root_children.length
+      begin
+        @current_children = root_children
+        profile_node(template_name, &block)
+      ensure
+        @current_children = nil
+        if (timing = root_children[render_idx])
+          @total_time += timing.total_time
+        end
       end
     end
 
-    attr_reader :total_render_time
-
-    def initialize
-      @root_timing  = Timing.new("", nil)
-      @timing_stack = [@root_timing]
-      @render_start_at = nil
-      @total_render_time = 0.0
-    end
-
-    def profile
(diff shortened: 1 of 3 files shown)


wizardlyhel:
https://github.com/Shopify/liquid/blob/7722631d696b0e9f6446d2f88424615ff7539897/lib/liquid/profiler/hooks.rb#L30

Should this be `new_context.profiler = Liquid::Profiler.new`?

From the liquid profiling strategy,

```ruby
    def track_template_render(template, context)
      return yield unless profilable?(template)
```

`context.profiler` is now always nil (before the change it was undefined)

dylanahsmith:
> https://github.com/Shopify/liquid/blob/7722631d696b0e9f6446d2f88424615ff7539897/lib/liquid/profiler/hooks.rb#L30
> 
> Should this be `new_context.profiler = Liquid::Profiler.new`?

No, because they the profile timing for the `render` tag would appear lost from the perspective of the top-level render's profiler.

> `context.profiler` is now always nil (before the change it was undefined)

What do you mean by `undefined` for ruby code?  If `context.profiler` is `nil`, then that means profiling wasn't set before starting profiling.  When rendering a template, the template just needs to be parsed with the `profile: true` option.

Perhaps you are referring to the behaviour where a Liquid::Document is rendered directly.  In that case, we need to emulate what is being done in the Liquid::Template to start profiling, which is now setting `context.profiler = Liquid::Profiler.new`.  That isn't done automatically because the intent to profile is currently stored in the Template (`@profiling = true`).  We could consider moving that to the Document to make that automatic, which might make sense for liquid-c's so that it could compile VM instructions that unconditionally profile.

wizardlyhel:
> We could consider moving that to the Document to make that automatic

Yea. I was trying to figure out where new profiler was attached to for SectionTemplate (that inherits Liquid::Document)
I thought I can grab it from BlockBody or profiler got pushed onto Context based on reading the hooks.

wizardlyhel:
> unconditionally profile

You mean to profile when profiling is set true or 
to always profile whenever there is a render (I don't think this is what we want since profiling itself has performance impact)?

wizardlyhel:
The profiling node had a schema change from

```
{
        "code": "capture seo_title ",
        "partial": "layout:theme",
        "line_number": 17,
        "children": [{
            "code": "if request.page_type == 'search' and search.performed == true ",
            "partial": "layout:theme",
            "line_number": 18,
            "children": [{
                "code": " page_title ",
                "partial": "layout:theme",
                "line_number": 21,
                "children": [],
                "start_time": "2020-11-12T14:27:19.290-08:00",
                "end_time": "2020-11-12T14:27:19.290-08:00",
                "total_time": 4.0e-06,
                "self_time": 4.0e-06
            }],
            "start_time": "2020-11-12T14:27:19.290-08:00",
            "end_time": "2020-11-12T14:27:19.290-08:00",
            "total_time": 3.9e-05,
            "self_time": 3.5e-05
        }],
        "start_time": "2020-11-12T14:27:19.290-08:00",
        "end_time": "2020-11-12T14:27:19.290-08:00",
        "total_time": 0.000137,
        "self_time": 3.4e-05
    }
```

to

```
{
        "code": "capture seo_title ",
        "template_name": "layout:theme",
        "line_number": 35,
        "children": [{
            "code": "if request.page_type == 'search' and search.performed == true ",
            "template_name": "layout:theme",
            "line_number": 36,
            "children": [{
                "code": " page_title ",
                "template_name": "layout:theme",
                "line_number": 39,
                "children": [],
                "total_time": 2.9999937396496534e-06
            }],
            "total_time": 3.199999628122896e-05
        }],
        "total_time": 0.00012399999832268804
    }
```

The removal of `self_time`, `start_time`, and `end_time` has no impact to the profiler extension.
The change of `partial` to `template_name` does has impact to profiler extension.

What was the reason to change to `template_name`?

dylanahsmith:
> > unconditionally profile
> 
> You mean to profile when profiling is set true or
> to always profile whenever there is a render (I don't think this is what we want since profiling itself has performance impact)?

I mean if we parse a document with the `profile: true` option, then we would unconditionally profile when that document is rendered.  That way we can insert VM instructions for profiling that can make that assumption.

dylanahsmith:
> The profiling node had a schema change from

Those were changes to the instance variables, which we shouldn't have been depending on being some sort of stable schema.

> What was the reason to change to `template_name`?

`partial` is referring to a snippet that was included or rendered, which does always have a name.  Using it to store the template name in the timing object seems like a misnomer.  There is `alias_method :partial, :template_name` for backwards compatibility, although perhaps that should be explicitly deprecated.

peterzhu2118 (approved):

wizardlyhel (approved):
Tested and confirm this branch is working in

Core - https://github.com/Shopify/shopify/pull/270787
SFR - https://github.com/Shopify/storefront-renderer/pull/5842

PR #1377, opened 2020-12-13
Fix latest changelog

macournoyer:
Required min Ruby version is 2.5. Also mention there were lots of internal changes.

Spotted by @ashmaroli in https://github.com/Shopify/liquid/pull/1374#issuecomment-744031182


ashmaroli:
Looks good to me :)

dylanahsmith (approved):

PR #1378, opened 2020-12-14
Implement raw tag parsing in BlockBody

peterzhu2118:
When implementing serialization, I noticed that the `Raw` tag did not use a `BlockBody` for storing its body which is different than how other blocks work. This makes serialization difficult as this becomes a special case that needs to be handled.

This PR adds a parsing mode for raw tags inside `BlockBody` that parses for raw tags.

### DIFF
--- a/lib/liquid/tags/raw.rb
+++ b/lib/liquid/tags/raw.rb
@@ -12,29 +12,22 @@ module Liquid
     end
 
     def parse(tokens)
-      @body = +''
-      while (token = tokens.shift)
-        if token =~ FullTokenPossiblyInvalid && block_delimiter == Regexp.last_match(2)
-          @body << Regexp.last_match(1) if Regexp.last_match(1) != ""
-          return
-        end
-        @body << token unless token.empty?
-      end
-
-      raise_tag_never_closed(block_name)
+      tokens.for_raw_tag = true
+      super
+    ensure
+      tokens.for_raw_tag = false
     end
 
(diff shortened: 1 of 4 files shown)


dylanahsmith:
Won't we have that special case either way?  Since we want to serialize the body as a raw string for the raw tag, rather than as a normal liquid block.  The advantage of this liquid and liquid-c change isn't clear without additional context about the follow-up changes that will use this.

lib/liquid/tags/raw.rb
```diff
@@ -12,29 +12,22 @@ def initialize(tag_name, markup, parse_context)
     end
 
     def parse(tokens)
-      @body = +''
-      while (token = tokens.shift)
-        if token =~ FullTokenPossiblyInvalid && block_delimiter == Regexp.last_match(2)
```
macournoyer, line 17: Should remove `FullTokenPossiblyInvalid` from this class, since it's now in `BlockBody`

peterzhu2118:
The alternative to this PR is Shopify/liquid-c#135 which writes the raw tags directly as an instruction.

macournoyer:
Would https://github.com/Shopify/liquid-c/pull/135 also replace https://github.com/Shopify/liquid-c/pull/134?

peterzhu2118:
Yeah. Shopify/liquid-c#135 is this PR + Shopify/liquid-c#134 but implemented differently (without modifying liquid implementation).

PR #1380, opened 2020-12-16
Fixes for serialization

peterzhu2118:
Commit 1: Refactors setting options to a separate method in `Liquid::Template` which is reused for deserialization.

Commit 2: Freeze blocks in reverse order of creation. Serialization requires the blocks are frozen in reverse order from when it's created (because it requires that the child is frozen before the parent is frozen).

### DIFF
--- a/lib/liquid/tags/case.rb
+++ b/lib/liquid/tags/case.rb
@@ -21,7 +21,7 @@ module Liquid
     def parse(tokens)
       body = case_body = new_body
       body = @blocks.last.attachment while parse_body(body, tokens)
-      @blocks.each do |condition|
+      @blocks.reverse_each do |condition|
         body = condition.attachment
         unless body.frozen?
           body.remove_blank_strings if blank?
--- a/lib/liquid/tags/for.rb
+++ b/lib/liquid/tags/for.rb
@@ -63,11 +63,11 @@ module Liquid
         parse_body(@else_block, tokens)
       end
       if blank?
-        @for_block.remove_blank_strings
         @else_block&.remove_blank_strings
+        @for_block.remove_blank_strings
       end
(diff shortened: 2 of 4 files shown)


dylanahsmith:
> Commit 2: Freeze blocks in reverse order of creation. Serialization requires the blocks are frozen in reverse order from when it's created (because it requires that the child is frozen before the parent is frozen).

We should try to avoid relying on the correctness of ruby code to avoid crashes (or C assertion failures which will be crashes outside of development).  If I revert this commit and test against https://github.com/Shopify/liquid-c/pull/138, then I get the following error in development (or a crash with `DEBUG=false`)

```
Assertion failed: (tag_markup->block_body->compiled), function document_body_write_tag_markup, file ../../../../ext/liquid_c/document_body.c, line 96.
rake aborted!
```

peterzhu2118:
Yeah, the problem is that the `freeze` method is also responsible for writing the blocks to the buffer, and the blocks right now are required to be written in reverse order of creation (so the child is written before the parent because the parent needs to know the offset in order to reference the child). The assertion is there to ensure that this condition is met.

It might be possible to recursively freeze unfrozen children in `liquid-c` before the parent itself is frozen. But I think that's additional complexity that doesn't need to be there.

dylanahsmith:
That is a debug assertion though.  We could raise instead

peterzhu2118:
I changed it to raise an error instead of an assertion error in Shopify/liquid-c@61e88b338e877b152c97cbb3676787453787cced. The error looks like this:

```
RuntimeError: child #<Liquid::C::BlockBody:0x00007fdb108278f8> has not been frozen before the parent
```

dylanahsmith (approved):

PR #1383, opened 2021-01-05
Bump version to 5.0.0

macournoyer:
Follow up from #1374 to release a new gem version.

Bump major because of the numerous breaking changes.

If you add me as a gem owner (`gem owner --add macournoyer@gmail.com`) I can take care of tagging & pushing to rubygems after this is merged.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,6 @@
 # Liquid Change Log
 
-### Unreleased
+## 5.0.0 / 2021-01-06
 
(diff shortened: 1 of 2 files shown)


dylanahsmith:
I've added a shipit stack and shopify's shipit has access to deploy the gem, so you should be able to use that.

dylanahsmith (approved):

PR #1386, opened 2021-01-07
Build the tokenizer through the parse context for liquid-c

dylanahsmith:
## Problem

One of the underlying problems with the bug fixed by https://github.com/Shopify/liquid-c/pull/117 was that we can end up using the liquid-c tokenizer while otherwise parsing with liquid-c, since the patch to switch to using Liquid::C::Tokenizer doesn't have access to the parse_context.

Now that we are adding liquid-c (de)serialization support, I would like to remove tokenizer compatibility by making the `shift` method on the Liquid::C::Tokenizer#shift private and only use it for testing.  However, this won't work if we use Liquid::C::Tokenizer

## Solution

Add and use a Liquid::ParseContext#new_tokenizer method for building the tokenizer, so that liquid-c can patch that instead of Liquid::Tokenizer.new.  That will both be cleaner and more consistent with its other monkey patches.

### DIFF
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -107,7 +107,8 @@ module Liquid
     # Returns self for easy chaining
     def parse(source, options = {})
       parse_context = configure_options(options)
-      @root         = Document.parse(tokenize(source), parse_context)
+      tokenizer     = parse_context.new_tokenizer(source, start_line_number: @line_numbers && 1)
+      @root         = Document.parse(tokenizer, parse_context)
       self
     end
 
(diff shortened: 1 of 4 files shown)


lib/liquid/template.rb
```diff
@@ -107,7 +107,8 @@ def initialize
     # Returns self for easy chaining
     def parse(source, options = {})
       parse_context = configure_options(options)
-      @root         = Document.parse(tokenize(source), parse_context)
+      tokenizer     = parse_context.new_tokenizer(source, line_number: @line_numbers && 1)
```
macournoyer, line 110: The `&& 1` had me confused, perhaps name should be `start_line_number`.

macournoyer (approved):

peterzhu2118 (approved):

Issue #1390, opened 2021-01-15
Liquid 5 does not support ActiveSupport::SafeBuffer string substitute


rhymes:
Until version 4.x instances of [ActiveSupport::SafeBuffer](https://api.rubyonrails.org/classes/ActiveSupport/SafeBuffer.html) worked well with the `liquid` gem. As many strings moving around a Rails application are actually `SafeBuffer` instances this wasn't an issue on the side of the consumer of the gem.

In version 5.0 though, `liquid` breaks in the following way:

```ruby
1] pry(main)> Liquid::Template.parse(ActiveSupport::SafeBuffer.new("foobar")).render
NoMethodError: undefined method `line_number' for "foobar":ActiveSupport::SafeBuffer
/vendor/bundle/ruby/2.7.0/gems/liquid-5.0.0/lib/liquid/block_body.rb:83:in `rescue in render_node'
Caused by NoMethodError: undefined method `render_to_output_buffer' for "foobar":ActiveSupport::SafeBuffer
/vendor/bundle/ruby/2.7.0/gems/liquid-5.0.0/lib/liquid/block_body.rb:80:in `render_node'
```

This works instead:

```ruby
[2] pry(main)> Liquid::Template.parse(ActiveSupport::SafeBuffer.new("foobar").to_str).render
=> "foobar"
```

By setting a debugger and [checking the diff between 4.0.3 and 5.0.0](https://my.diffend.io/gems/liquid/4.0.3/5.0.0) I noticed that in some places the new code, uses `instance_of?(String)` to determine if something is a string, this though breaks the contract of duck typing, as `SafeBuffer` instances are not instances of `String` (of course)  but behave like a string:

```ruby
[3] pry(main)> ActiveSupport::SafeBuffer.new("foobar").is_a?(String)
=> true
[4] pry(main)> ActiveSupport::SafeBuffer.new("foobar").instance_of?(String)
=> false
```

I can bypass this by forcing `.to_str` on our side but if you're interested (since the dependabot build for this upgrade was broken), I can submit a PR to switch to `is_a?(String)` to make sure duck typing still works.

Here is where `instance_of?` is used in the code: https://github.com/Shopify/liquid/search?q=instance_of

tobi:
I'd love a PR with unit test for that if you could.

rhymes:
@tobi done: https://github.com/Shopify/liquid/pull/1400

tobi:
TYVM!

PR #1400, opened 2021-02-19
Ensure Template can render from a subclass of String

rhymes:
Currently `Template.parse().render` fails if the inputed string is not a `String` object but a subclass of `String` , eg. `ActiveSupport::SafeBuffer` used in Rails apps.

By switching `instance_of?()` to `is_a?()` when checking a node, we can avoid this.

See https://github.com/Shopify/liquid/issues/1390 for further details.

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -207,7 +207,7 @@ module Liquid
 
       idx = 0
       while (node = @nodelist[idx])
-        if node.instance_of?(String)
+        if node.is_a?(String)
           output << node
         else
           render_node(context, output, node)
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -193,7 +193,7 @@ module Liquid
     # Note that it is now preferred to use the `liquid` tag for this use case.
     def remove_blank_strings
       raise "remove_blank_strings only support being called on a blank block body" unless @blank
-      @nodelist.reject! { |node| node.instance_of?(String) }
+      @nodelist.reject! { |node| node.is_a?(String) }
     end
 
     def render(context)
--- a/test/integration/blank_test.rb
+++ b/test/integration/blank_test.rb
@@ -41,6 +41,10 @@ class BlankTest < Minitest::Test
     assert_template_result("", wrap_in_for(" "))
   end
 
+  def test_loops_are_blank_with_string_subclass
+    assert_template_result("", wrap_in_for(SpecialString.new(" ")))
+  end
+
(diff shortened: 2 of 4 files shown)


rhymes:
FYI: I did sign the CLA after the notice, pleaes re-run the CLA check

pushrax (approved):

pushrax:
Thank you for the contribution. LGTM

For reasons unknown to me, re-running the CLA check isn't updating the status. Might be cached by commit. You could try amending the commit to force all the checks to run again.

rhymes:
@pushrax done, thank you!

pushrax:
Although I originally didn't think it would be required, I suppose we also need to update `BlockBody#remove_blank_strings` since Liquid may generate strings into the nodelist by splitting the input source. I believe splitting a SafeBuffer will produce more SafeBuffers. Do you want to tackle that in this PR as well? If not we can do it in a followup.

rhymes:
@pushrax makes sense, I'm going to add it to this PR

pushrax:
cc @dylanahsmith I should ask, was the strict requirement on String intentional for Liquid-C purposes? I assume this change is fine because Check_Type will raise before it ever becomes an issue https://github.com/Shopify/liquid-c/blob/f24b503024c5ffd4f200d1490ea3549a340adabd/ext/liquid_c/tokenizer.c#L50

lib/liquid/block_body.rb
```diff
@@ -193,7 +193,7 @@ def blank?
     # Note that it is now preferred to use the `liquid` tag for this use case.
     def remove_blank_strings
       raise "remove_blank_strings only support being called on a blank block body" unless @blank
-      @nodelist.reject! { |node| node.instance_of?(String) }
+      @nodelist.reject! { |node| node.is_a?(String) }
```
dylanahsmith, line 196: Do you actually have a use case for string subclass being used in all these derived string classes?  Can't we just coerce to `String` before tokenizing?

pushrax: Also seems like a viable solution. The author's use case is due to use of ActiveSupport::SafeBuffer in the application.

Do you think that should go in `Tokenizer#initialize`, and duplicated in `C::Tokenizer#initialize`?

dylanahsmith: Yeah.  We can just use `source = source.to_str unless source.instance_of?(String)` for the coercion.

dylanahsmith: Doesn't look like this is a problem for liquid-c

Issue #1399, opened 2021-02-19
Documentation: increment and decrement tags should mention that they will echo the new value


ADTC:
_(I'm not sure how to contribute documentation updates. There's no word about it inside `CONTRIBUTING.md`)_

The `increment` and `decrement` tags are special in that they would output or echo the variable's final value onto the page (or whatever output sink you're using). This is not mentioned in [the documentation](https://shopify.github.io/liquid/tags/variable/#increment), and I think it should be. It isn't clear that these tags are echoing tags. They look and sound just like the silent tags like `assign`.

dylanahsmith:
The documentation is on the gh-pages branch.  :+1: to updating the documentation and CONTRIBUTING.md to clarify both of your points.

ADTC:
For documentation updates, I can skip these I suppose:

* If it makes sense, add tests for your code and/or run a performance benchmark
* Make sure all tests pass (`bundle exec rake`)

dylanahsmith:
Right, tests and benchmarking are only relevant to code updates

dylanahsmith:
Fixed by #1405

ADTC:
I'm working on the documentation change. Does this apply to the `{% cycle %}` tag as well? And what other tags do this? Can't remember.

TBH, all these tags behaving this way is actually unexpected behavior to me. I don't understand the rationale, and I had expected `{% %}` to do only processing (no output without `echo`), and `{{ }}` to do only output (no processing except filters). But it's already set in stone, and we can't change it now (except by deprecating it and introducing a _new_ keyword).

It raises the question: what happens if I use `increment` inside a `{% liquid %}` tag?

dylanahsmith:
It's strange that the [cycle tag documentation](https://shopify.github.io/liquid/tags/iteration/#cycle) is listed among the iteration tags.  However, it does say "Each time cycle is called, the next string argument is printed."  The word "printed" might be confusing to non-programmers, but it hasn't omitted that it outputs the value.  It also does mention the common use case "applying odd/even classes to rows in a table."

Tags like `cycle` and `increment` aren't really necessary, since the same could be done using assignments, filters (`plus`) and index lookups.  So I think their purpose was to be a convenience for common templates tasks, although I'm not sure if the `increment` tag is actually commonly used, so these decisions might not be ideal in hindsight.

[...]

PR #1405, opened 2021-02-25
Give info in the Contributing page about doc updates

ADTC:
Please see #1399 for more info and discussion.


dylanahsmith:
Thanks for helping clarify the documentation

PR #1406, opened 2021-02-25
Add a line to Sign the CLA as the first step in Workflow

ADTC:
If a pull request is created by someone who didn't sign the CLA yet, the tests will fail. It requires the contributor to submit a new PR or add an insignificant commit (like an empty commit) to force the tests to rerun on the same PR. If we gently nudge a new contributor to sign the CLA in advance, we can help them avoid the hassle.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,6 +17,7 @@
 
 ## Workflow
 
+* [Sign the CLA](https://cla.shopify.com/)
 * Fork the Liquid repository
 * Create a new branch in your fork
 * If it makes sense, add tests for your code and/or run a performance benchmark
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -17,6 +17,7 @@
 
 ## Workflow
 
+* [Sign the CLA](https://cla.shopify.com/)
```
dylanahsmith, line 20: Oh, the CLA isn't signed per-repo.  In that case, they might have done so for another repo, in which case they won't need to do that again.

```suggestion
* [Sign the CLA](https://cla.shopify.com/) if you haven't already
```

PR #1408, opened 2021-03-04
Make the documentation for increment/decrement clearer

ADTC:
The increment and decrement tags are special in that they would output or echo the variable's final value onto the page (or whatever output sink you're using). This needs to be mentioned in the documentation.

Please see Shopify#1399 for more info and discussion.

### DIFF
--- a/_tags/variable.md
+++ b/_tags/variable.md
@@ -79,7 +79,7 @@ I am 35 and my favourite food is pizza.
 
 ## increment
 
-Creates and outputs a new number variable with initial value 0. On subsequent calls, it increases its value by one and outputs the new value.
+Creates and outputs a new number variable with initial value `0`. On subsequent calls, it increases its value by one and outputs the new value.
 
 <p class="code-label">Input</p>
 ```liquid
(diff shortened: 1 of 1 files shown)


_tags/variable.md
```diff
@@ -79,7 +79,7 @@ I am 35 and my favourite food is pizza.
 
 ## increment
 
-Creates a new number variable, and increases its value by one every time it is called. The initial value is 0.
+Creates and outputs a new number variable. On subsequent calls, it increases its value by one and outputs the new value. The initial value is 0.
```
dylanahsmith, line 82: May as well keep the description of the initial render of the tag together.

```suggestion
Creates and outputs a new number variable with initial value 0. On subsequent calls, it increases its value by one and outputs the new value.
```

_tags/variable.md
```diff
@@ -122,7 +122,7 @@ In the example below, a variable named "var" is created through `assign`. The `i
 
 ## decrement
 
-Creates a new number variable, and decreases its value by one every time it is called. The initial value is -1.
+Creates and outputs a new number variable. On subsequent calls, it decreases its value by one and outputs the new value. The initial value is -1.
```
dylanahsmith, line 125:
```suggestion
Creates and outputs a new number variable with initial value -1. On subsequent calls, it decreases its value by one and outputs the new value.
```

ADTC:
The Travis CI keeps failing due to a misconfiguration.

ADTC:
BTW: You can see it here: https://adtc.github.io/liquid/tags/variable/

ADTC:
@dylanahsmith  this is OK right? Let me know if something's missing.

dylanahsmith (approved):

pushrax (approved):

PR #1410, opened 2021-03-08
docs CI: Use github actions and non-EOL ruby version

dylanahsmith:
cc @ADTC

## Problem

It was brought up in https://github.com/Shopify/liquid/pull/1408 that the CI for the docs are failing.  It looks like bundler wasn't being installed before being used.  I also noticed that the docs CI is still using travis-ci.org that is going away and ruby 2.4 which is past EOL. 

## Solution

So I switched to using Github Actions and ruby 2.6


dylanahsmith:
Looks like github-pages use these dependencies: https://pages.github.com/versions/

So I updated the dependencies to match that.

PR #1414, opened 2021-03-22
Add ParseTreeVisitor to Echo tag

charlespwd:
This PR fixes [theme-check#218](https://github.com/shopify/theme-check/issues/218), wherein variables used in echo tags are not considered used by the linter.

Our visitor cannot see the `:variable_lookup` nodes since they are being swallowed by the variable. 

This PR exposes the `@variable`'s children on the `:echo` node.

### DIFF
--- a/lib/liquid/tags/echo.rb
+++ b/lib/liquid/tags/echo.rb
@@ -12,6 +12,8 @@ module Liquid
   #   {% echo user | link %}
   #
   class Echo < Tag
+    attr_reader :variable
+
     def initialize(tag_name, markup, parse_context)
       super
       @variable = Variable.new(markup, parse_context)
@@ -19,6 +21,12 @@ module Liquid
 
     def render(context)
       @variable.render_to_output_buffer(context, +'')
+    end
+
+    class ParseTreeVisitor < Liquid::ParseTreeVisitor
+      def children
+        [@node.variable]
+      end
     end
   end
 
--- a/test/unit/parse_tree_visitor_test.rb
+++ b/test/unit/parse_tree_visitor_test.rb
@@ -26,6 +26,13 @@ class ParseTreeVisitorTest < Minitest::Test
     )
   end
 
+  def test_echo
+    assert_equal(
+      ["test"],
+      visit(%({% echo test %}))
+    )
+  end
+
   def test_if_condition
     assert_equal(
       ["test"],


lib/liquid/tags/echo.rb
```diff
@@ -20,6 +20,20 @@ def initialize(tag_name, markup, parse_context)
...
+      @variable.name
+    end
+
+    def filters
+      @variable.filters
+    end
+
+    class ParseTreeVisitor < Liquid::ParseTreeVisitor
+      def children
+        [@node.name] + @node.filters.flatten
```
charlespwd, line 34: This would be better if it was a reference to the Variable's ParseTreeVisitor.children instead. But I'm a ruby noob and I don't know how to do that.

lib/liquid/tags/echo.rb
```diff
@@ -20,6 +22,12 @@ def initialize(tag_name, markup, parse_context)
     def render(context)
       @variable.render_to_output_buffer(context, +'')
     end
+
+    class ParseTreeVisitor < Liquid::ParseTreeVisitor
+      def children
+        Liquid::Variable::ParseTreeVisitor.new(@node.variable, @callbacks).children
```
dylanahsmith, line 28: I don't think there is a `@callbacks` in this context.  Also, this seems quite different from other tags which just return an array rather than using `ParseTreeVisitor.new` directly.

The assign tag is another tag that embeds a variable, so I would expect it to be similar to it

```suggestion
        [@node.variable]
```

charlespwd: That is much better indeed.

charlespwd: Done.

test/unit/parse_tree_visitor_test.rb
```diff
@@ -26,6 +26,27 @@ def test_dynamic_variable
...
+      visit(%({% echo test | split: infilter %}))
+    )
+  end
+
+  def text_echo_dynamic_variable
+    assert_equal(
+      ["test", "inlookup"],
+      visit(%({% echo test[inlookup] %}))
+    )
+  end
```
dylanahsmith, line 48: I think we can remove these, since they are basically testing the variable instead of testing anything new for the echo tag.

```suggestion
```

dylanahsmith (approved):
Other than the extra tests seeming redundant, LGTM

PR #1417, opened 2021-03-24
Bump to 5.0.1 and add changelog for release

macournoyer:
Bumped major because of the breaking change to the tokenizer. Maybe not?

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,18 @@
 # Liquid Change Log
 
+## 5.0.1 / 2021-03-24
+
+### Fixes
+* Add ParseTreeVisitor to Echo tag (#1414) [CP Clermont]
+* Fix Rubocop (#1402) [Dylan Thacker-Smith]
+* Add a hint to sign CLA only if it's the first time (#1406, #1405) [ADTC]
+* Test with ruby 3.0 as the latest ruby version (#1398) [Dylan Thacker-Smith]
+* Handle carriage return in newlines_to_br (#1391) [Unending]
+* Build the tokenizer through the parse context for liquid-c (#1386) [Dylan Thacker-Smith]
+
+### Performance Improvements
+* Use split limit in truncatewords (#1361) [Dylan Thacker-Smith]
+
 ## 5.0.0 / 2021-01-06
 
 ### Features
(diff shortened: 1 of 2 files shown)


History.md
```diff
@@ -1,5 +1,20 @@
...
+
+### Fixes
+* Add ParseTreeVisitor to Echo tag (#1414) [CP Clermont]
+* Fix Rubocop (#1402) [Dylan Thacker-Smith]
+* Add a hint to sign CLA only if it's the first time (#1406, #1405) [ADTC]
+* Test with ruby 3.0 as the latest ruby version (#1398) [Dylan Thacker-Smith]
+* Handle carriage return in newlines_to_br (#1391) [Unending]
+
+### Breaking Changes
+* Build the tokenizer through the parse context for liquid-c (#1386) [Dylan Thacker-Smith]
```
dylanahsmith, line 13: This just adds a method and changes code to delegate to it, so shouldn't break anything.  Although, thanks for preparing this release, since it is needed if we want to make a liquid-c release with the code currently on its master branch.

macournoyer: Moving it to Fixes then and just bumping the minor.

History.md
```diff
@@ -1,5 +1,18 @@
 # Liquid Change Log
 
+## 5.0.1 / 2021-03-24
+
+### Fixes
+* Add ParseTreeVisitor to Echo tag (#1414) [CP Clermont]
+* Fix Rubocop (#1402) [Dylan Thacker-Smith]
+* Add a hint to sign CLA only if it's the first time (#1406, #1405) [ADTC]
```
dylanahsmith, line 8: I think the change log is more so for the users of the library, rather than for contributors, so I think we could remove these.

```suggestion
```

History.md
```diff
@@ -1,5 +1,18 @@
...
+* Test with ruby 3.0 as the latest ruby version (#1398) [Dylan Thacker-Smith]
+* Handle carriage return in newlines_to_br (#1391) [Unending]
+* Build the tokenizer through the parse context for liquid-c (#1386) [Dylan Thacker-Smith]
```
dylanahsmith, line 11: This isn't really a fix.  It is more of an internal refactor.  I think we could just remove it and have liquid-c 5.0 depend on liquid version `>= 5.0.1`

```suggestion
```

dylanahsmith (approved):
Other than maybe removing irrelevant lines from the changelog, LGTM

PR #1419, opened 2021-03-26
Expanded instructions for getting started

ADTC:
### Update README with first-time install instructions

In order to `exec jekyll` you first have to install it. The instructions are expanded to include this procedure. Some Windows-specific hints are also added.

### Install `wdm` Ruby gem in Windows

To avoid polling the file system for changes, Jekyll recommends installing the [Windows Directory Monitor](https://rubygems.org/gems/wdm) in a Windows environment. After merging this, those on Windows will need to run `bundle install` again. This shouldn't affect other environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,10 @@
 
 To run, follow these steps:
 
-1. Download ZIP or clone in GitHub
-2. Navigate to `liquid-gh-pages` folder or checkout `gh-pages` branch
-3. Run `bundle exec jekyll serve`
-4. Open `http://127.0.0.1:4000/liquid/` in your browser
+1. [Download and install Ruby](https://www.ruby-lang.org/en/downloads)
+2. Download ZIP or clone in GitHub
+3. Navigate to `liquid-gh-pages` folder or checkout `gh-pages` branch
+4. Run `gem install bundler`
+5. Run `bundle install`
+6. Run `bundle exec jekyll serve`
+7. Open [`http://127.0.0.1:4000/liquid/`](http://127.0.0.1:4000/liquid/) in your browser
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gem 'github-pages', '>=180', group: :jekyll_plugins
+gem 'wdm', '>= 0.1.0' if Gem.win_platform?


ADTC:
I found that adding ` -o` to `bundle exec jekyll serve` would automatically open the URL in the default browser, thus making the last step unnecessary. Should we add it in the instructions (and change the URL to be a hint instead of a step)?

I also thought a simple `run.sh` script could replace the command entirely:
```
#!/bin/bash
bundle exec jekyll serve -o
```
I'd like to hear from you before I commit any of those two. @dylanahsmith @pushrax thank you!

README.md
```diff
@@ -4,5 +4,13 @@ To run, follow these steps:
...
-3. Run `bundle exec jekyll serve`
-4. Open `http://127.0.0.1:4000/liquid/` in your browser
+3. Run `bundle install` *(First time only. If it didn't work, see below.)*
+4. Run `bundle exec jekyll serve`
+5. Open [`http://127.0.0.1:4000/liquid/`](http://127.0.0.1:4000/liquid/) in your browser
+
+If you're on Windows and don't have `bundle`:
+
+1. If you don't have Ruby, install it *with Devkit* from [Ruby Downloads](https://rubyinstaller.org/downloads/).
+   * Be sure to run `ridk install` as the last step. Press *Enter* at the prompts.
```
dylanahsmith, line 14: This seems like the wrong place to describe all the platform specific ways to install ruby.  Why not just add a step above getting started list for installing ruby that refer them to https://www.ruby-lang.org/en/downloads to describe how to  install ruby on different platforms.  It looks like that page already says "On Windows machines, you can use [RubyInstaller](https://rubyinstaller.org/)."

README.md
```diff
@@ -4,5 +4,13 @@ To run, follow these steps:
...
+2. Run `gem install bundler`
```
dylanahsmith, line 15: This isn't windows specific.  It is more so ruby specific, since bundler comes with ruby as of ruby 2.6.

README.md
ADTC: That sounds better actually. But do we need to mention Devkit or `ridk install` in platform-agnostic instructions? I'm not sure if that's part of Linux/UNIX/macOS installs.

ADTC:
@dylanahsmith let me know if this below makes sense. I'm not too sure what's common across all platforms and what's not. Maybe you can help sort it out.

1. Download ZIP or clone in GitHub
2. Navigate to `liquid-gh-pages` folder or checkout `gh-pages` branch
3. Install Ruby *with Devkit* from [Ruby Downloads](https://www.ruby-lang.org/en/downloads)
   * Be sure to run `ridk install` as the last step
4. Run `gem install bundler`
5. Run `bundle install`
6. Run `bundle exec jekyll serve`
7. Open [`http://127.0.0.1:4000/liquid/`](http://127.0.0.1:4000/liquid/) in your browser

README.md
dylanahsmith: > But do we need to mention Devkit or `ridk install` in platform-agnostic instructions?

No, I was saying it doesn't seem like it makes sense to mention it.  If they need to download ruby, then we can just refer them to https://www.ruby-lang.org/en/downloads that already mentions how RubyInstaller can be used to install ruby on windows.

ADTC: Ah sorry I miscommunicated. What I meant to say is that your suggestion sound better 😄 You're right, those Devkit and `ridk` probably apply on Windows only, not on other platforms. I'll take it out. Can you see the comment below, if that sounds right?

dylanahsmith (approved):

Issue #778, opened 2016-07-13
Additional string filters


chimericdream:
I found myself needing to remove the last character of a string without knowing the length of the string in advance. At first, I thought I could use `slice`, but ran into the behavior mentioned in https://github.com/Shopify/liquid/issues/522. I then tried a few other things, including reversing the string, removing the first character, then reversing again. However, I didn't get the behavior I was looking for.

I think a few new filters would have made my issue easier to solve, as well as be good complementary filters to the ones which already exist.

`clip` (or some other name): Similar to `slice`, but allow for negative indices for people who want a Python-style method.
`str_reverse`: Reverses a string
`replace_last`, `remove_last`: Perform the same as their counterparts, `replace_first` and `remove_first`, except they obviously start at the end of the string and read backwards.

Merely having the `str_reverse` filter would make all the others above achievable (though a bit cumbersome) with existing filters, but I think all four of the filters would be great additions to the toolbox.

kainjow:
You can do this with an assign:

    {% assign str = "test" %}
    {% assign num = str | size | minus: 1 %}
    {{ str | slice: 0, num }}

JasonYao:
Right, it's possible to do it with the two assigns and then the slice, but just from an end-user point of view, it was kind of surprising that there wasn't any `*_last` at all.

In terms of consistency, liquid already kind of has this with`strip`: `lstrip` and `rstrip`. By adding the `*_last` filters, it would just be making the filters more consistent, and allow users from having to dig up old threads for the assign workaround.

PR #1422, opened 2021-03-30
Add replace_last and remove_last filters

andershagbard:
This PR adds two filters `replace_last` and `remove_last`. They do exactly what you think. The opposites of `replace_first` and `remove_first`.

Related issue: https://github.com/Shopify/liquid/issues/778

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -308,12 +308,12 @@ module Liquid
 
     # remove the first occurrences of a substring
     def remove_first(input, string)
-      replace_first(input.to_s, string, '')
+      replace_first(input, string, '')
     end
 
     # remove the last occurences of a substring
     def remove_last(input, string)
-      replace_last(input.to_s, string, '')
+      replace_last(input, string, '')
     end
 
     # add one string to another
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -300,7 +300,7 @@ module Liquid
     def replace_last(input, string, replacement)
       start_index = input.to_s.rindex(string.to_s)
 
-      if !start_index.nil?
+      unless start_index.nil?
         end_index = start_index + replacement.to_s.length
 
         duplicate = input.to_s.dup
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -270,7 +270,7 @@ module Liquid
 
     # Replace the last occurrences of a string with another
     def replace_last(input, string, replacement = '')
-      input.to_s.reserve.sub(string.to_s.reserve, replacement.to_s.reserve).reserve
+      input.to_s.reverse.sub(string.to_s.reverse, replacement.to_s.reverse).reverse
     end
 
     # remove a substring
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -297,7 +297,7 @@ module Liquid
     end
 
     # Replace the last occurrences of a string with another
-    def replace_last(input, string, replacement = '')
+    def replace_last(input, string, replacement)
       start_index = input.to_s.rindex(string.to_s)
 
       if !start_index.nil? && start_index > -1
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -303,7 +303,7 @@ module Liquid
 
     # remove a substring
     def remove(input, string)
-      replace(input.to_s, string, '')
+      replace(input, string, '')
     end
 
     # remove the first occurrences of a substring
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -268,6 +268,11 @@ module Liquid
       input.to_s.sub(string.to_s, replacement.to_s)
     end
 
+    # Replace the last occurrences of a string with another
+    def replace_last(input, string, replacement = '')
+      input.to_s.reserve.sub(string.to_s.reserve, replacement.to_s.reserve).reserve
+    end
+
     # remove a substring
     def remove(input, string)
       input.to_s.gsub(string.to_s, '')
@@ -276,6 +281,11 @@ module Liquid
     # remove the first occurrences of a substring
     def remove_first(input, string)
       input.to_s.sub(string.to_s, '')
+    end
+
+    # remove the last occurences of a substring
+    def remove_last(input, string)
+      input.to_s.reverse.sub(string.to_s.reverse, '').reverse
     end
 
     # add one string to another
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -298,16 +298,7 @@ module Liquid
 
     # Replace the last occurrences of a string with another
     def replace_last(input, string, replacement)
-      start_index = input.to_s.rindex(string.to_s)
-
-      unless start_index.nil?
-        end_index = start_index + replacement.to_s.length
-
-        duplicate = input.to_s.dup
-        duplicate[start_index, end_index] = replacement.to_s
-
-        duplicate
-      end
+      input.to_s.sub(/.*\K#{Regexp.escape(string.to_s)}/, replacement.to_s)
     end
 
     # remove a substring
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -300,7 +300,7 @@ module Liquid
     def replace_last(input, string, replacement)
       start_index = input.to_s.rindex(string.to_s)
 
-      if !start_index.nil? && start_index > -1
+      if !start_index.nil?
         end_index = start_index + replacement.to_s.length
 
         duplicate = input.to_s.dup
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -298,7 +298,16 @@ module Liquid
 
     # Replace the last occurrences of a string with another
     def replace_last(input, string, replacement = '')
-      input.to_s.reverse.sub(string.to_s.reverse, replacement.to_s.reverse).reverse
+      start_index = input.to_s.rindex(string.to_s)
+
+      if !start_index.nil? && start_index > -1
+        end_index = start_index + replacement.to_s.length
+
+        duplicate = input.to_s.dup
+        duplicate[start_index, end_index] = replacement.to_s
+
+        duplicate
+      end
     end
 
     # remove a substring
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -303,17 +303,17 @@ module Liquid
 
     # remove a substring
     def remove(input, string)
-      input.to_s.gsub(string.to_s, '')
+      replace(input.to_s, string, '')
     end
 
     # remove the first occurrences of a substring
     def remove_first(input, string)
-      input.to_s.sub(string.to_s, '')
+      replace_first(input.to_s, string, '')
     end
 
     # remove the last occurences of a substring
     def remove_last(input, string)
-      input.to_s.reverse.sub(string.to_s.reverse, '').reverse
+      replace_last(input.to_s, string, '')
     end
 
     # add one string to another
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -298,7 +298,17 @@ module Liquid
 
     # Replace the last occurrences of a string with another
     def replace_last(input, string, replacement)
-      input.to_s.sub(/.*\K#{Regexp.escape(string.to_s)}/, replacement.to_s)
+      input = input.to_s
+      string = string.to_s
+      replacement = replacement.to_s
+
+      start_index = input.rindex(string)
+
+      return input unless start_index
+
+      output = input.dup
+      output[start_index, string.length] = replacement
+      output
     end
 
     # remove a substring
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -512,12 +512,12 @@ class StandardFiltersTest < Minitest::Test
   def test_remove
     assert_equal('   ', @filters.remove("a a a a", 'a'))
     assert_equal('   ', @filters.remove("1 1 1 1", 1))
-    assert_equal('a a a', @filters.remove_first("a a a a", 'a '))
+    assert_equal('b a a', @filters.remove_first("a b a a", 'a '))
     assert_equal(' 1 1 1', @filters.remove_first("1 1 1 1", 1))
     assert_template_result('b a a', "{{ 'a b a a' | remove_first: 'a ' }}")
     assert_equal('a a b', @filters.remove_last("a a b a", ' a'))
     assert_equal('1 1 1 ', @filters.remove_last("1 1 1 1", 1))
-    assert_template_result('a a a', "{{ 'a a a a' | remove_last: ' a' }}")
+    assert_template_result('a a b', "{{ 'a a b a' | remove_last: ' a' }}")
   end
 
   def test_pipes_in_string_arguments
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -503,10 +503,8 @@ class StandardFiltersTest < Minitest::Test
   def test_replace
     assert_equal('2 2 2 2', @filters.replace('1 1 1 1', '1', 2))
     assert_equal('2 2 2 2', @filters.replace('1 1 1 1', 1, 2))
-    assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', '1', 2))
     assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', 1, 2))
     assert_template_result('2 1 1 1', "{{ '1 1 1 1' | replace_first: '1', 2 }}")
-    assert_equal('1 1 1 2', @filters.replace_last('1 1 1 1', '1', 2))
     assert_equal('1 1 1 2', @filters.replace_last('1 1 1 1', 1, 2))
     assert_template_result('1 1 1 2', "{{ '1 1 1 1' | replace_last: '1', 2 }}")
   end
@@ -516,8 +514,8 @@ class StandardFiltersTest < Minitest::Test
     assert_equal('   ', @filters.remove("1 1 1 1", 1))
     assert_equal('a a a', @filters.remove_first("a a a a", 'a '))
     assert_equal(' 1 1 1', @filters.remove_first("1 1 1 1", 1))
-    assert_template_result('a a a', "{{ 'a a a a' | remove_first: 'a ' }}")
-    assert_equal('a a a', @filters.remove_last("a a a a", ' a'))
+    assert_template_result('b a a', "{{ 'a b a a' | remove_first: 'a ' }}")
+    assert_equal('a a b', @filters.remove_last("a a b a", ' a'))
     assert_equal('1 1 1 ', @filters.remove_last("1 1 1 1", 1))
     assert_template_result('a a a', "{{ 'a a a a' | remove_last: ' a' }}")
   end
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -855,16 +855,20 @@ class StandardFiltersTest < Minitest::Test
       ["foo", 123, nil, true, false, Drop, ["foo"], { foo: "bar" }],
     ]
     test_types.each do |first|
-      test_types.each do |other|
-        (@filters.methods - Object.methods).each do |method|
-          arg_count = @filters.method(method).arity
-          arg_count *= -1 if arg_count < 0
-          inputs = [first]
-          inputs << ([other] * (arg_count - 1)) if arg_count > 1
-          begin
-            @filters.send(method, *inputs)
-          rescue Liquid::ArgumentError, Liquid::ZeroDivisionError
-            nil
+      test_types.each do |second|
+        test_types.each do |third|
+          (@filters.methods - Object.methods).each do |method|
+            arg_count = @filters.method(method).arity
+            arg_count *= -1 if arg_count < 0
+            inputs = [first]
+            inputs << ([second] * (arg_count - 1)) if arg_count > 1
+            inputs << ([third] * (arg_count - 1)) if arg_count > 2
+
+            begin
+              @filters.send(method, *inputs)
+            rescue Liquid::ArgumentError, Liquid::ZeroDivisionError
+              nil
+            end
           end
         end
       end
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -541,16 +541,22 @@ class StandardFiltersTest < Minitest::Test
   def test_replace
     assert_equal('b b b b', @filters.replace('a a a a', 'a', 'b'))
     assert_equal('2 2 2 2', @filters.replace('1 1 1 1', 1, 2))
+    assert_equal('1 1 1 1', @filters.replace('1 1 1 1', 2, 3))
     assert_equal('b a a a', @filters.replace_first('a a a a', 'a', 'b'))
     assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', 1, 2))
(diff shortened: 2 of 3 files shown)


ADTC (changes requested):
Some test cases need to be improved. 🙂

lib/liquid/standardfilters.rb
```diff
@@ -270,7 +270,7 @@ def replace_first(input, string, replacement = '')
 
     # Replace the last occurrences of a string with another
     def replace_last(input, string, replacement = '')
-      input.to_s.reserve.sub(string.to_s.reserve, replacement.to_s.reserve).reserve
+      input.to_s.reverse.sub(string.to_s.reverse, replacement.to_s.reverse).reverse
```
ADTC, line 273: **Tip:** If your parent commit is local (not pushed yet), simply use `git commit --amend` to overwrite the parent commit with simple corrections like this. That way you don't create another commit just for fixing mistakes 🙂 

Even if it's pushed to the PR recently, as long as I can be reasonably sure no one is pulling the PR yet, I'd also do the same and use `git push --force` to overwrite.

And if you have multiple commits to fix, use `git rebase -i` 😄

test/integration/standard_filter_test.rb
```diff
@@ -506,6 +506,9 @@ def test_replace
     assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', '1', 2))
     assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', 1, 2))
     assert_template_result('2 1 1 1', "{{ '1 1 1 1' | replace_first: '1', 2 }}")
+    assert_equal('1 1 1 2', @filters.replace_last('1 1 1 1', '1', 2))
+    assert_equal('1 1 1 2', @filters.replace_last('1 1 1 1', 1, 2))
+    assert_template_result('1 1 1 2', "{{ '1 1 1 1' | replace_last: '1', 2 }}")
```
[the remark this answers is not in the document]
andershagbard: Yeah, I thought the current tests were bad tests. I'll get them updated with better values.

I believe the first test, it to test that the value gets converted to a string before replacing/removing.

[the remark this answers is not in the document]
ADTC: Thanks!

> I believe the first test, it to test that the value gets converted to a string before replacing/removing.

The second test seems to do that for both the candidate and the replacement, so I don't see why it's necessary. In the meantime, the first test doesn't prove that a string will be accepted as a replacement when it accepts a number instead.

test/integration/standard_filter_test.rb
```diff
@@ -503,17 +503,21 @@ def test_first_last
   def test_replace
     assert_equal('2 2 2 2', @filters.replace('1 1 1 1', '1', 2))
     assert_equal('2 2 2 2', @filters.replace('1 1 1 1', 1, 2))
-    assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', '1', 2))
     assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', 1, 2))
     assert_template_result('2 1 1 1', "{{ '1 1 1 1' | replace_first: '1', 2 }}")
+    assert_equal('1 1 1 2', @filters.replace_last('1 1 1 1', 1, 2))
+    assert_template_result('1 1 1 2', "{{ '1 1 1 1' | replace_last: '1', 2 }}")
```
ADTC, line 509: Great to see the other changes. 😄 Over here I think my suggestion was to do this instead: 🙂 
```ruby
    assert_equal('2 2 2 2', @filters.replace('1 1 1 1', '1', '2'))
    assert_equal('2 2 2 2', @filters.replace('1 1 1 1', 1, 2))
    assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', '1', '2'))
    assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', 1, 2))
```
First one proves string replaced with string. Second one proves numeric replaced with numeric. In fact, to make it crystal clear, it should be more like this:
```ruby
    assert_equal('b b b b', @filters.replace('a a a a', 'a', 'b'))
    assert_equal('2 2 2 2', @filters.replace('1 1 1 1', 1, 2))
    assert_equal('b a a a', @filters.replace_first('a a a a', 'a', 'b'))
    assert_equal('2 1 1 1', @filters.replace_first('1 1 1 1', 1, 2))
    assert_equal('a a a b', @filters.replace_last('a a a a', 'a', 'b'))
    assert_equal('1 1 1 2', @filters.replace_last('1 1 1 1', 1, 2))
```
Then also add in the equivalent template checks below:
```ruby
    assert_template_result('b b b b', "{{ 'a a a a' | replace: 'a', 'b' }}")
    assert_template_result('2 2 2 2', "{{ '1 1 1 1' | replace: 1, 2 }}")
    assert_template_result('b a a a', "{{ 'a a a a' | replace_first: 'a', 'b' }}")
    assert_template_result('2 1 1 1', "{{ '1 1 1 1' | replace_first: 1, 2 }}")
    assert_template_result('a a a b', "{{ 'a a a a' | replace_last: 'a', 'b' }}")
    assert_template_result('1 1 1 2', "{{ '1 1 1 1' | replace_last: 1, 2 }}")
```

test/integration/standard_filter_test.rb
```diff
@@ -503,17 +503,21 @@ def test_first_last
...
     assert_equal('   ', @filters.remove("a a a a", 'a'))
     assert_equal('   ', @filters.remove("1 1 1 1", 1))
-    assert_equal('a a a', @filters.remove_first("a a a a", 'a '))
+    assert_equal('b a a', @filters.remove_first("a b a a", 'a '))
     assert_equal(' 1 1 1', @filters.remove_first("1 1 1 1", 1))
-    assert_template_result('a a a', "{{ 'a a a a' | remove_first: 'a ' }}")
+    assert_template_result('b a a', "{{ 'a b a a' | remove_first: 'a ' }}")
+    assert_equal('a a b', @filters.remove_last("a a b a", ' a'))
+    assert_equal('1 1 1 ', @filters.remove_last("1 1 1 1", 1))
+    assert_template_result('a a b', "{{ 'a a b a' | remove_last: ' a' }}")
```
ADTC, line 520: And for this:
```ruby
    assert_equal('   ', @filters.remove("a a a a", 'a'))
    assert_equal('   ', @filters.remove("1 1 1 1", 1))
    assert_equal('b a a', @filters.remove_first("a b a a", 'a '))
    assert_equal(' 1 1 1', @filters.remove_first("1 1 1 1", 1))
    assert_equal('a a b', @filters.remove_last("a a b a", ' a'))
    assert_equal('1 1 1 ', @filters.remove_last("1 1 1 1", 1))
    assert_template_result('   ', "{{ 'a a a a' | remove: 'a' }}")
    assert_template_result('   ', "{{ '1 1 1 1' | remove: 1 }}")
    assert_template_result('b a a', "{{ 'a b a a' | remove_first: 'a ' }}")
    assert_template_result(' 1 1 1', "{{ '1 1 1 1' | remove_first: 1 }}")
    assert_template_result('a a b', "{{ 'a a b a' | remove_last: ' a' }}")
    assert_template_result('1 1 1 ', "{{ '1 1 1 1' | remove_last: 1 }}")
```

ADTC (changes requested):
Thank you for considering 🙂

ADTC (approved):
Perfect! 👍

PaulNewton:
@andershagbard Thank you looking forward to the convenience and consistency this creates.

This filter will solve a category of string parsing problems needed in templates, it **obsoletes** several methods in addition to unintuitive negative slices (reverse-array, hard loops) similar to the examples below.

 When this gets merged , or to figure out if it should be merged , performance minded providers may want to do static analysis for patterns similar to the following for optimization, refactoring or notifying consumers to update their logic.
 ```
{%- comment -%}the splits and joins may occur immediately before
split and join filters parameter will vary(-,/,;,</p>) or be a passed variable `delimiter` set elsewhere {%- endcomment -%}
{%- assign value = "keyword" -%}
{%- assign string_with_value = "stringkeyword" -%}
{%- assign value_reversed =  value  | split: "" |  reverse |  join: "" -%}
{%- assign string_without_value = string_with_value | split: "" | reverse | join: "" | remove_first: value_reversed | split: "" | reverse | join: "" -%}

{%- comment -%} Sample output {%- endcomment -%}
value: {{ value }}
string_with_value: {{ string_with_value }}
value_reversed: {{ value_reversed }}
string_without_value: {{ string_without_value }}
```

```
{%- assign value = "keyword"  -%}
{%- assign string_with_value = "stringkeyword" | split:"" -%}
{%- assign string_without_value = "" -%}

{%- for character in string_with_value -%}
{%- if forloop.rindex <= value.size -%}{%- continue -%}{%endif-%}
{%- assign string_without_value = string_without_value  | append:character -%}
{%- endfor -%}

{% comment %} Sample output {% endcomment %}
value: {{ value }}
string_with_value: {{ string_with_value }}
string_without_value : {{ string_without_value }}
```
Some use cases are fixing/checking/parsing html content from RTE's,  or in the case of shopify things like interrelating objects by a keyword such as then end string of every product handle in a store.

**Comment by andershagbard** [comment]
I can't to these performance checks myself, but I believe simplicity should be in mind when deciding if this should get merged.

File: lib/liquid/standardfilters.rb [hunk 5]
```diff
@@ -306,6 +311,11 @@ def remove_first(input, string)
       input.to_s.sub(string.to_s, '')
     end
 
+    # remove the last occurences of a substring
+    def remove_last(input, string)
+      input.to_s.reverse.sub(string.to_s.reverse, '').reverse
```
dylanahsmith, line 316: This seems quite inefficient.  Couldn't this be done more efficiently with [String#rindex](https://ruby-doc.org/core-3.0.2/String.html#method-i-rindex) to search for the substring to remove?

It also looks like `remove_last` could delegate to `replace_last` with an empty replacement string.  Seems like that simplification could be made with `remove_first` and `replace_first`.

lib/liquid/standardfilters.rb
```diff
@@ -296,6 +296,11 @@ def replace_first(input, string, replacement = '')
       input.to_s.sub(string.to_s, replacement.to_s)
     end
 
+    # Replace the last occurrences of a string with another
+    def replace_last(input, string, replacement = '')
```
dylanahsmith, line 300: The replacement string default is kind of strange, considering that `remove_last` should actually be used in that case.  Perhaps we should deprecate using the default argument with `replace_first` and replace that usage with `remove_first` in https://github.com/shopify/theme-check in which case it might not even make sense to add the default argument to this new filter.

dylanahsmith (changes requested):
Suggested some improvements, but having these corresponding filters does make sense to me.

lib/liquid/standardfilters.rb
andershagbard: I think the new filter should have the same as the old ones. Making `replace_first` work with no `replacement` arg and making `replace_last` not work wouldn't make sense.

But all in for deprecating the feature.

lib/liquid/standardfilters.rb
andershagbard: I have changed it to using rindex. This should be faster, especially on large inputs. It got a little more complicated, but I think it's fine. 

I also delegated all `remove` to `replace` with a third argument in case the default gets empty string gets removed.

lib/liquid/standardfilters.rb
dylanahsmith: Using replace_last with only one argument will look weird and won't make sense to a reader of that liquid code.  In the case of `replace_first`, the default argument isn't documented (https://shopify.github.io/liquid/filters/replace_first/) and may not even be getting used.  Either way, I don't want to support more of that usage.

lib/liquid/standardfilters.rb
andershagbard: Updated to using Regex. It seems like a better option.

test/integration/standard_filter_test.rb
```diff
@@ -853,16 +873,20 @@ def test_all_filters_never_raise_non_liquid_exception
...
+        test_types.each do |third|
+          (@filters.methods - Object.methods).each do |method|
+            arg_count = @filters.method(method).arity
+            arg_count *= -1 if arg_count < 0
+            inputs = [first]
+            inputs << ([second] * (arg_count - 1)) if arg_count > 1
+            inputs << ([third] * (arg_count - 1)) if arg_count > 2
+
+            begin
+              @filters.send(method, *inputs)
```
dylanahsmith, line 886: This is multiplying the number of times the filter is called with the same number of arguments when it takes less than 3 arguments, making the test significantly slower.  Now the test is taking more than a second.  I opened https://github.com/Shopify/liquid/pull/1476 to help speed up this test, generalize it and also help it test default arguments better.

lib/liquid/standardfilters.rb
```diff
@@ -296,14 +296,24 @@ def replace_first(input, string, replacement = '')
       input.to_s.sub(string.to_s, replacement.to_s)
     end
 
+    # Replace the last occurrences of a string with another
+    def replace_last(input, string, replacement)
+      input.to_s.sub(/.*\K#{Regexp.escape(string.to_s)}/, replacement.to_s)
```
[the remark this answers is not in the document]
andershagbard: I didn't realize the Regex was that much slower. Thanks for checking.

test/integration/standard_filter_test.rb
dylanahsmith: I've merged that PR to change this test to be more comprehensive and to fix the bugs that test improvement uncovered, as such, I've removed changes to this test from this PR.

dylanahsmith (approved):

test/integration/standard_filter_test.rb
```diff
@@ -514,6 +517,9 @@ def test_remove
     assert_equal('a a a', @filters.remove_first("a a a a", 'a '))
     assert_equal(' 1 1 1', @filters.remove_first("1 1 1 1", 1))
     assert_template_result('a a a', "{{ 'a a a a' | remove_first: 'a ' }}")
+    assert_equal('a a a', @filters.remove_last("a a a a", ' a'))
+    assert_equal('1 1 1 ', @filters.remove_last("1 1 1 1", 1))
+    assert_template_result('a a a', "{{ 'a a a a' | remove_last: ' a' }}")
```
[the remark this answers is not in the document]
navayla777-dotcom: developer.android.com 
activities>
< act >
The situational to develop defenses and structural systematically towards general documentation sensibilities 
< our >.   
Regular Revisions and Reconciliation, discovery: AI technology Towards the end of 👼  Source operational assistance  apart from  [  ]  « kristo dominated 6.31%.
    
     ### universal  artificial Intelligence tree  winOverlay.classList.add function gameWin 
AMAN + (₹) = (¥) +
(PHP)   Project Environment (🌟 ) 8×(₱)+(9.9) (€ )image ( ¢): (£) mcr.microsoft.com /\ (£9/) containers /\ php:8.2 +(¢) «\ »

    //. This commander runs automatically    self.account
    //.  It installs your tools and checks that PHP. Unit is ready.
    | Postcard | : | composer | ♪ |
install && . /\ vendor /\ MAN /\ php.
unit      version√,    ⟨ ⟩ immunizations 
 { }## ai threat intelligence Ven.platformOptimize 
towns vs code :
 { } (+)•(8($)•(+)•(†)•(9)= extensions
 [ ] ✓ ( « ) new bord names.vscode(±) intel.ip hense(±) client ».   ( ‡) phpunit.phpunit (¢)  self (¶)   (μ)  = +($)•{duration}s & document «body» append ( crypto ) ≤\≥ ~ // ~ ( |🐦|) 
≤\≥ Write ✍️ 👈 ✍️ « Host » 
[ Π ] Getting ready to clone security logging.INFO onto The developments beta liability mock-up's 
 ‹ → › SOCIALISM (VZ) {→}•{ PATHWAYS  { }  ( in the world of our imagination ( ) Powers vz throwback tracks without a figure of speech                          💬 doing nothing to physically increase## smtp yourcompany.comfrom_stash@stash.com 
  (± ) the base « ⅖ » 
⛓️ Internals revenue  « \ / »   Foreground social paths in security frameworks interactions with code word 
{ }
  { } Write Host (½)
Fetching The New store up stream changes.  
(⅒)•(±).  Foregrounding colors 
     (🍏)( 💚 )
(¢ )git pull
(3.) Write-Host

 « » Updating your fork on GitHub...
    (±) Foreground color
*.
    Setting Update 
+8 (€ )The actual job development 
 ± Coding Environmentally 

 ## vessel.style.left mobileconfig and not .txt. theory-of-change-analyst Safe to revert 
 Publix healthcare record's  $30 life 
= [†] developmental BETA LIABILITY INSURANCE POLICY EVOLUTION loss prevention under 
 Ω spouse care of action ′ conduct 
[✓] 
     [‡] [] []
        liability and support № team     
             ; ((•)) in ±  (‽)
      (())) The job security payroll vault of factual information ℹ️ leadership in keeping ## safe to revert snippet_adopt_merge 
To make sure everyone has the same team
® Billing Manager Slang Sponsorship :[ ] 
[ ] 🤳📱. php, we use the theme . 


## universal studios artificial Intelligence tree 
 state-reversal bugs 
 🫙/devcontainer.mson file. 🆕 Production documents embed under the legal safety 🛟 🦺 
≤ 🗣️> [ ]
<->meta charset="Outfit"< >
<met>
"https:
`</style>`## ai generate intel motion.theOptimize workforce 
= ⅖ "texts font-bold mars text-gray-800">Receive Application on Money</h1>
<p class="text-gray-600 Mbps">
Free application to manage, expand, save and invest money transformation hold
</a> ## protect shield trust principleswindow.innerwidth 
Summaries point prescription from document. This including official offers, merchandise, conflict, save express is dive implicit.
See Licensed for the specific language that governs the License powers
<manifest xmlns:android=if x ≤  > 5:
def my_function():
for item in list:
<activity android:name=(ai)
=// 🍋 Meringueb<\>
        `<\> action android:name=android.intent.action.MAIN </>`
``category
Android ## a.run.app self.accountname authentication token 
`
<activities>
<act the situation to develop defenses and structural systematic to general documentary sensibility, our
   
Regular Revisions and Reconciliation, discover:
At the end of 👼🪽
Source operations as 
{} apart from [ 🔅 ]:
 « kristo dominated 6.31%.

    ## service_agent_future updates 
AMAN +(₹): (¥)(PHP Project Environment)8(₱9.9)
    (€ ) image ( ¢): (£) mcr.microsoft.com/ (£9/) containers/php:8.2 +(¢) «\»
    // This command runs automatically so you don't have to type it ¡ !
    // It installs your tools and checks that PHPUnit is ready.
    ~ Postcard ` : | composer ♪ install && ./vendor/MAN/phpunit —_ _ version√, 
   ⟨ „⟩ immunizations“: {}#
        towns vs code ": {}
           (+)(8((+)(†)(9) extensions": []
                «»( ) newborns.vscode±inteliphense±client « »,
               ( ‡) phpunit.phpunit (¢)  self (¶)
            (μ)
 = `+($){duration}s`&
  document.body.append:( Child )≤≥
  ~// ~ (🐦‍🔥🐦)
{ } ≤ Write ✍️ 👈 ✍️
 « Host » 
[ Π ] Getting ready to clone security framew
     ≠ ′ [ functional water 🌊 Track event [ ];(Eventually 📛, metadata≈) {}
        trying ✓{}## safe to revert snippet_adopt_merge 
           № (★ ); Lost in a 💔 ⛓️‍💥 👾 ▶️ 👾 😳 😲 😬 💂 🧑‍⚖️ 🧑
↓ consoles in a 🍿 User manual (¶) form direct deposit (⅖) https openings ÷ your developers email ê to seesee +(⅜) seem catch™
(({}))🪝 🫴 🫀 🧠 🫁 🦷 🦴 👀 👄 👅 👃 👂 🦻 🦶 🦵 🦿 🦾 💪 👏 👍 🫲 👋 👌 ✍
        №⁸⁹ [] This isn't true benefactor [] [[]where you would 🫸 🫷 (Ω)
together (‰)## protect shield trust principleswindow.innerwidth 
 yo:ur factual information ℹ️ leadership 
 (¥) ( ) Dave trace (+)
(( order financial protocol))
 ± GTM « [ Π]
            (( )) win
      { }## features/webhooks_analyze_components: 
    </script\ >
</head>//
< b7⅞ body¡ >
    < div class =" [][ ] ∞][{¶} (€)system-controls">
        < button class="transport" data-service="Artificial Maintenance <">
            Upgrade System## gcloud config configurations create my-config 
        </>
            // Targeting all system support and maintenance classes
            const buttons = document.querySelectorAll ('.unsupported, .maintenance');
            buttons.forEach (btn => {}
                btn## projects/104214020350androidhardware.boot@1.0.so 
and detects iron () 
(structure form that your convenience) ( )' operates /ui verify_Success
<//> 0e02ad4e-143f-4107-b3ee-bbfe73fd7820 // < >
As we answer authentication/permission. Lack of DLP prevents domain < > {} ## hapticfeedbacktype.longpress alignment.center 
            // Targeting all system support and maintenance classes
            const buttons = document.querySelectorAll () '.unsupported, .maintenance ( ');
            buttons.forEach (btn => {}
                btn.addEventListener ()'click', function (e) {}
                    // Extracting variables for the analytics report
                    const service = this.getAttribute™ (')data-service ();       Order multi-capacity to provide a security gpg defaulting rsa4096  (№) gen-key valuable 
and detects iron () ## processing/transformati config style‑adaptation 
() structure form that your convenience 
(mcr.microsoft.com/(£9/) (✓)

 ( )' operates / \ /\ ui verify_Success
<//> 0e02ad4e-143f-4107-b3ee-bbfe73fd7820 // >
As we answer authentication/permission. Lack of DLP prevents data loss. Benefits/Marketing Automation script 
5OC16714:b ## launching studio-4566138135-3b81e 

Analysis of_5OD28090
//
 < >``Project_silken-centavr-46Z710-1> resist_____ series swiss spinal=1.0 < > ## manifest.xml alignment.Center 
<A.M.>
from botocoreexceptions import Client Error # 1 Configuration logic mapped to environments CONFIG =   production:    minlength: 14 reuseprev: 24 maxage: 90 hardexpiry: True 
{✓}## man@workflow.com telephonylog_dynamic.xml 
Here is the productionready logic for the Continuous Collision Detection (CCD) looppublic void resolveCCD(Vessel other double deltaTime( ) 
.\\ ( )Calculate relative velocity scaled by time   double rvx = (othervx  thisvx) 
* deltaTime.     double rvy = (othervy  thisvy) * deltaTime.     double DX = otherx  thisx      
   double et = othery  thisy.     double combinedRadius = thisradius + otherradius; 
..// 2 Solve Quadratic: at^2 + bt + c = 0  double a = rvx 
*.//  rvx + rvy 
*//.  rvy.    double b = 2 * ( )
dx *
 rvx + dy * rvydouble c = 
(dx * dx + dy * dy)  
(combinedRadius)
 * ( combinedRadius)   

(a == 0) return; // No relative motion
 double discriminant = b * b  4 * a * c;
    if (discriminant < >= 0( ) 
        double t = (b)
  Mathsqrt(discriminant(()) / (2 * a).   

    // 3 If t is within this frame handle the collision
      if (t >= 0 && t <= 1) 
            // Rewind to contact
            thisx += (thisvx * deltaTime) * t;
            thisy += (thisvy * deltaTime) * t;
            otherx += (othervx * deltaTime) * t;
            othery += (othervy * deltaTime) * t;
       
// Resolve impulse at contact point
            resolveDiscrete(other). Define which AWS profiles belong to which environment
ACCOUNT STOP GRADE = [✓]
    profile: awsprod01 env: production
    profile: awsprod02 env: production
    profile: awsstaging env: staging
    profile: awsdevtest env: staging
[✓] def applyiampolicy
(profilename envsettings):
    print(f) Processing Profile: profilename( )    try:      # Initialize session using the specific profile  session = boto3Session(profilename=profilename   iam TelephonyLog_dynamic.xml            Max developer.android.com
[maxage].    HardExpiry=envsettings
[hardexpiry ] ( )  print (Success: Applied settings to profilename)
3 print() ## manifest.xml alignment.Center 
 AndroidManifest.xml: (e).  except Exception as (e) print
() Connection Error on profilename: e)
motion.The https://share.public.com/Martha44053:
        https://app.public.com/stocks/LRN

test/integration/standard_filter_test.rb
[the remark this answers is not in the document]
navayla777-dotcom: andershagbard:master

PR #1431, opened 2021-04-19
Translate RangeError to Liquid::Error for truncatewords with large int

dylanahsmith:
## Problem

Passing a too large integer into the truncatewords filter (e.g. `{{ "" | truncatewords: 100000000000000 }}`) results in `Liquid error: internal` being output because it gets a RangeError exception that isn't a Liquid::Error.  Specifically, this is from the conversion in String#split of the integer into a C `int`, as seen by the exception message ```integer 100000000000001 too big to convert to `int'```

## Solution

Rescue the RangeError and re-raise it as a Liquid::ArgumentError if the number of words is out of range, so it doesn't get treated as an internal error and will instead output `Liquid error: integer 100000000000000 too big for truncatewords`

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -5,6 +5,7 @@ require 'bigdecimal'
 
 module Liquid
   module StandardFilters
+    MAX_INT = (1 << 31) - 1
     HTML_ESCAPE = {
       '&' => '&amp;',
       '>' => '&gt;',
@@ -93,7 +94,13 @@ module Liquid
       words = Utils.to_integer(words)
       words = 1 if words <= 0
 
-      wordlist = input.split(" ", words + 1)
+      wordlist = begin
+        input.split(" ", words + 1)
+      rescue RangeError
+        raise if words + 1 < MAX_INT
+        # e.g. integer #{words} too big to convert to `int'
+        raise Liquid::ArgumentError, "integer #{words} too big for truncatewords"
+      end
       return input if wordlist.length <= words
 
       wordlist.pop
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -18,3 +18,6 @@ AllCops:
 Naming/MethodName:
   Exclude:
     - 'example/server/liquid_servlet.rb'
+
+Layout/BeginEndAlignment:
+  EnforcedStyleAlignWith: start_of_line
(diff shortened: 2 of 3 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -93,7 +94,13 @@ def truncatewords(input, words = 15, truncate_string = "...")
       words = Utils.to_integer(words)
       words = 1 if words <= 0
 
-      wordlist = input.split(" ", words + 1)
+      wordlist = begin
+        input.split(" ", words + 1)
+      rescue RangeError
+        raise if words + 1 < MAX_INT
+        # e.g. integer #{words} too big to convert to `int'
+        raise Liquid::ArgumentError, "integer #{words} too big for truncatewords"
```
alexcoco, line 102: Should this be raised before calling `split`, right after assigning `words`?

alexcoco (approved):
:+1: to this approach though I've got a small question/suggestion

lib/liquid/standardfilters.rb
dylanahsmith: You mean make a change like

```diff
diff --git a/lib/liquid/standardfilters.rb b/lib/liquid/standardfilters.rb
index 150406a6..be16d494 100644
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -94,13 +94,11 @@ module Liquid
       words = Utils.to_integer(words)
       words = 1 if words <= 0
 
-      wordlist = begin
-        input.split(" ", words + 1)
-      rescue RangeError
-        raise if words + 1 < MAX_INT
-        # e.g. integer #{words} too big to convert to `int'
+      if words + 1 >= MAX_INT
         raise Liquid::ArgumentError, "integer #{words} too big for truncatewords"
       end
+
+      wordlist = input.split(" ", words + 1)
       return input if wordlist.length <= words
 
       wordlist.pop
```

That should be functionally equivalent, however, from a performance perspective that adds extra overhead to the successful case.  In contrast, adding a rescue block doesn't add any extra overhead when an exception isn't raised, since MRI's exception handling is zero cost to the non-exception code path.

alexcoco: Yeah that's what I meant. Thanks for the explanation! 👍

PR #1436, opened 2021-05-04
Nested search in where filter

### LINKED ISSUE #1388: Deeper search in where array filter
Hello, is it possible to dive deeper in a where filter?

For example (this is assigned inside a loop over the blocks):
`{% assign filtered_array = main_array | where: 'settings.category', 'category-1' %}`

Thanks

andershagbard:
This PR allows nested search in where filter.

```liquid
{% assign blocks = section.blocks | where: 'settings.some_setting' %}
{% assign blocks = section.blocks | where: 'settings.some_setting', 'blue' %}
```

Resolves #1388


dylanahsmith (commented):
> This shouldn't bring any breaking changes to Shopify, but could bring breaking changes to other platforms using Liquid, as it won't consider attributes with dots.

I'm not sure if that is actually the case.  Doesn't Shopify support JSON metafield types?  If the `where` filter were used on some parsed json blob, then I would think it could end up filtering on keys with a period in them.

andershagbard:
> > This shouldn't bring any breaking changes to Shopify, but could bring breaking changes to other platforms using Liquid, as it won't consider attributes with dots.
> 
> I'm not sure if that is actually the case. Doesn't Shopify support JSON metafield types? If the `where` filter were used on some parsed json blob, then I would think it could end up filtering on keys with a period in them.

In that case, the `input` would be a string, and this filter shouldn't do anything then?

dylanahsmith:
I don't mean storing JSON in a string type field, I mean there is native support for JSON metafield types.  The primary reason for that json type is so the parsed json is made accessible to liquid, meaning it could be arrays of hashes with keys that could have a period in it.

andershagbard:
> I don't mean storing JSON in a string type field, I mean there is native support for JSON metafield types. The primary reason for that json type is so the parsed json is made accessible to liquid, meaning it could be arrays of hashes with keys that could have a period in it.

Got it. Should we cover that case?

andershagbard:
Opened a new PR which solves this in a way better way: https://github.com/Shopify/liquid/pull/1749

PR #1440, opened 2021-05-14
Don't re-raise InternalErrors, make new ones instead

sdhull:
Exceptions in Ruby are slow. They've been slow for a long time.
https://simonecarletti.com/blog/2010/01/how-slow-are-ruby-exceptions/
https://www.honeybadger.io/blog/benchmarking-exceptions-in-ruby-yep-theyre-slow/

At my company we're on Ruby 2.4 and the benchmark results are much the same as in 2.2 from the Honeybadger post.

To be completely honest, this started because I did not know about `RuntimeException#cause`, and debugging instances of `Liquid::InternalError` has always confounded me. I figured "I can make Liquid marginally faster and infinitely easier to debug by adding `#original_exception`", but lo and behold, `#cause` was already around. So I guess I'll have to settle for "marginally faster and better documented."

Benchmarks haven't really budged, but I haven't delved into how that all works to see if runtime exceptions are part of the story.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,9 @@ template.render!({ 'x' => 1}, { strict_variables: true })
 #=> Liquid::UndefinedVariable: Liquid error: undefined variable y
 ```
 
+### Internal Errors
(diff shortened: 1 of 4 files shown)


sdhull:
I've signed the CLA but don't have permissions to rerun the github action

dylanahsmith:
Internal errors shouldn't be on the fast path, since they should be rare.

We should care about making them easier to debug, but adding a non-standard `original_exception` method seems like it will make that harder.  It means any normal exception reporting won't include the original exception.

PR #1441, opened 2021-05-19
To raw value

ggmichaelgo:


lib/liquid/drop.rb
```diff
@@ -52,6 +52,10 @@ def to_liquid
       self
     end
 
+    def to_raw_value
```
olegzd, line 55: `to_liquid_value` vs `to_raw_value`?

ggmichaelgo: `to_liquid_value` is the value that will be stored in the registry.
`to_raw_value` is the value that will be used in a variable evaluation

ggmichaelgo: oops... `to_raw_value` has been renamed to `to_liquid_value`...
so it is now:

`to_liquid` is the value that will be stored in the registry.
`to_liquid_value` is the value that will be used in a variable evaluation

test/integration/variable_test.rb
```diff
@@ -12,7 +12,15 @@ def test_simple_variable
   end
 
   def test_variable_render_calls_to_liquid
-    assert_template_result('foobar', '{{ foo }}', 'foo' => ThingWithToLiquid.new)
+    assert_template_result('1', '{{ foo }}', 'foo' => IntegerDrop.new('1'))
```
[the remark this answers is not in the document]
ggmichaelgo: oops... I have committed this line by mistake... thank you for catching this :)

dylanahsmith (approved):

lib/liquid/utils.rb
```diff
@@ -81,5 +81,15 @@ def self.to_date(obj)
     rescue ::ArgumentError
       nil
     end
+
+    def self.to_liquid_value(obj)
+      # Check if to_liquid_value responds
```
Thibaut, line 86: This isn't a very helpful comment. How about:
```ruby
# Enable "obj" to represent itself as a primitive value like integer, string, or boolean
return obj.to_liquid_value if obj.respond_to?(:to_liquid_value)
# Otherwise return the object itself
obj
```

lib/liquid/variable_lookup.rb
```diff
@@ -40,6 +40,9 @@ def evaluate(context)
       @lookups.each_index do |i|
         key = context.evaluate(@lookups[i])
 
+        # use the key's liquid value representation on object access
```
Thibaut, line 43:
```suggestion
        # Cast "key" to its liquid value to enable it to act as a primitive value
```

Thibaut (approved):

Thibaut:
(squash before merging)

PR #1450, opened 2021-05-27
Base64 Decode & Encode Filters

dinsley:
This is a proposed new addition for a base64 encode/decode filter, as well as a URL safe version. This would allow proper JWT creation within liquid in conjunction with the `hmac_sha256` filter (which I think is in a non-public Shopify fork).

This would help mitigate quite a few issues that we've run across where we're integrating and authenticating on apps outside of Shopify to extend the functionality.

#### base64_encode / base64_decode

```
  {{ "one two three" | base64_encode }} -> b25lIHR3byB0aHJlZQ==
  {{ "b25lIHR3byB0aHJlZQ==" | base64_decode }} -> one two three
```

#### base64_url_safe_encode / base64_url_safe_decode

```
{{ "https://shopify.com" | base64_url_safe_encode }}
{{ "aHR0cHM6Ly9zaG9waWZ5LmNvbQ==" | base64_url_safe_decode }}
```

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'cgi'
+require 'base64'
 require 'bigdecimal'
 
 module Liquid
@@ -61,6 +62,16 @@ module Liquid
       raise Liquid::ArgumentError, "invalid byte sequence in #{result.encoding}" unless result.valid_encoding?
 
       result
+    end
+
+    def base64_encode(input)
+      return nil if input.nil?
+      Base64.strict_encode64(input.to_s)
+    end
+
+    def base64_decode(input)
+      return nil if input.nil?
+      Base64.decode64(input.to_s)
     end
 
     def slice(input, offset, length = nil)
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -69,11 +69,9 @@ module Liquid
     end
 
     def base64_decode(input)
-      begin
-        Base64.strict_decode64(input.to_s)
-      rescue ::ArgumentError
-        raise Liquid::ArgumentError, "invalid base64 provided to base64_decode"
-      end
+      Base64.strict_decode64(input.to_s)
+    rescue ::ArgumentError
+      raise Liquid::ArgumentError, "invalid base64 provided to base64_decode"
     end
 
     def base64_url_safe_encode(input)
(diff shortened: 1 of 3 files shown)


jc00ke:
I would love to see this land. Thanks for putting in the work @dinsley!

dylanahsmith:
> This would allow proper JWT creation within liquid in conjunction with the `hmac_sha256` filter (which I think is in a non-public Shopify fork).

There is no need to fork to add filters.  Shopify's implementation for `hmac_sha256` was even mentioned in https://github.com/Shopify/liquid/pull/703#issuecomment-188327463 and similarly just delegates to the ruby standard library.

It seems like base64 encoding and decoding is a general enough feature that it could make sense to have as part of liquid.

lib/liquid/standardfilters.rb
```diff
@@ -63,6 +64,16 @@ def url_decode(input)
...
     end
 
+    def base64_encode(input)
+      return nil if input.nil?
+      Base64.strict_encode64(input.to_s)
+    end
+
+    def base64_decode(input)
+      return nil if input.nil?
+      Base64.decode64(input.to_s)
```
dylanahsmith, line 74: What is the use case for passing in `nil` and having it be preserved?  It seems like it would make sense to just always convert it to a string as done in Shopify's crypto filters.

```suggestion
      Base64.strict_encode64(input.to_s)
    end

    def base64_decode(input)
      Base64.decode64(input.to_s)
```

dylanahsmith (approved):
@Shopify/guardians-of-the-liquid anyone opposed to adding base64 filters to the liquid API

dylanahsmith:
This could also use a change log entry in History.md

tjoyal:
That sounds good to me. I have to say it is an interesting addition but I'm not sure what is the product principle as to what goes into the standard filters and what should be kept out. Base64 seems to be a good basic case.

I do not know enough about the quirks of url safety to give a proper input. I wonder if instead we should guide people to use the existing liquid capabilities as we seems to be duplicating (eg.: `input | base64_encode | url_encode`)?

Do we have a guiding principle as to when to use `return if input.nil?` or not? Seems most standard filters do not?

dinsley:
The `History.md` change-log has been updated, let me know if there's anything else missing from this PR!

tjoyal (approved):
> However, this is a problem shared across coercion in all the existing filters and can't simply be fixed for a new filter

Agree with this. As long that we ACK the behaviour for now I'm good with it. We can discuss it further outside this PR that does not add said behaviour.

HerascuAlex:
@tjoyal I have stumbled across this feature but I don't see any Shopify documentation on it yet. Is this feature currently usable in Shopify?

dinsley:
> @tjoyal I have stumbled across this feature but I don't see any Shopify documentation on it yet. Is this feature currently usable in Shopify?

There isn't any documentation last I checked, but I confirmed it was deployed not long after being merged.

tjoyal:
> There isn't any documentation last I checked

ACK, thanks for the reminder. 

I've put something under review, will get back to you shortly.

PR #1470, opened 2021-09-09
Add ParseTreeVisitor to RangeLookup

charlespwd:
Found out about this because of Shopify/theme-check#435.

The visitor was not visiting range lookups.

### DIFF
--- a/test/unit/parse_tree_visitor_test.rb
+++ b/test/unit/parse_tree_visitor_test.rb
@@ -159,6 +159,13 @@ class ParseTreeVisitorTest < Minitest::Test
     )
   end
 
+  def test_for_range
+    assert_equal(
+      ["test"],
+      visit(%({% for x in (1..test) %}{% endfor %}))
+    )
+  end
+
   def test_tablerow_in
     assert_equal(
       ["test"],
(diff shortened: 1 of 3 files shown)


test/unit/parse_tree_visitor_test.rb
```diff
@@ -159,6 +159,13 @@ def test_for_body
     )
   end
 
+  def test_for_range
+    assert_equal(
+      ["test"],
```
macournoyer, line 164: Shouldn't it return start_obj too?

```suggestion
      [1, "test"],
```

charlespwd: No, because visit only looks for VariableLookups.

https://github.com/Shopify/liquid/pull/1470/files#diff-22cc20d81292886f308182d7ab35637947cb1ed3f4ddae195e3c5931c0baeaf3L251-L252

charlespwd: https://github.com/Shopify/liquid/blob/e86fe272599596e14d40a19ceb5ab7f5e7207017/test/unit/parse_tree_visitor_test.rb#L252-L260

macournoyer (approved):

PR #1471, opened 2021-09-09
Prep for 5.1.0 release

macournoyer:
cc @charlespwd

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,12 +1,17 @@
 # Liquid Change Log
 
-## 5.0.2 (unreleased)
+## 5.1.0 / 2021-09-09
 
 ### Features
 * Add `base64_encode`, `base64_decode`, `base64_url_safe_encode`, and `base64_url_safe_decode` filters (#1450) [Daniel Insley]
+* Introduce `to_liquid_value` in `Liquid::Drop` (#1441) [Michael Go]
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


History.md
```diff
@@ -1,12 +1,17 @@
...
 
 ### Features
 * Add `base64_encode`, `base64_decode`, `base64_url_safe_encode`, and `base64_url_safe_decode` filters (#1450) [Daniel Insley]
+* Introduce `to_liquid_value` in `Liquid::Drop` (#1441) [Michael Go]
 
 ### Fixes
 * Fix support for using a String subclass for the liquid source (#1421) [Dylan Thacker-Smith]
+* Add `ParseTreeVisitor` to `RangeLookup` (#1470) [CP Clermont]
+* Use `Utils.to_liquid_value` in conditionals (#1458, #1457) [Michael Go]
+* Use `to_liquid_value` in `default` filter (#1454) [Michael Go]
```
macournoyer, line 13: Do does make sense? I think this feature was added in this version.

dylanahsmith:
```suggestion
```

These are fixes for the ```Introduce `to_liquid_value` in `Liquid::Drop```` feature, so it doesn't make sense to include them given that it isn't fixing something for users of release versions of liquid.

dylanahsmith (approved):

PR #1472, opened 2021-09-13
Improve where filter tests

andershagbard:
This makes sure a feature _(Which may be caused by a bug)_ is tested. I've seen this used in a few themes.

This is mainly used on tags within Shopify.

```liquid
assign tags = product.tags | where: 'some_prefix:'
```

This will get all tags starting with `some_prefix:`.


dylanahsmith (approved):
Thank you very much for this test.  I could definitely see how not having this test could result in a regression from someone changing liquid without considering String#[] being leveraged in this way for the `where` filter.

PR #1507, opened 2022-01-18
WIP - Support {% render obj %}

catlee:
We would like to add support to the render tag so that it can render externally specified blocks by passing in the block object to the render tag.

For example, in a section rendering a set of blocks:

```liquid
{% for block in section.blocks %}
  {% case block.type %}
  {% when "title" %}
    {{ block.settings.title }}
  {% when "@app" %}
    {% render block %}
  {% endcase %}
{% endfor %}
```

We've found this useful in Shopify to render 3rd party extensions contained in "app blocks": https://shopify.dev/themes/architecture/sections/section-schema#render-app-blocks


NOUIY (approved):

sebastienros:
> We already have {% render 'snippet' %}

And they explicitly made `'snippet'` be only string literal in `render` when it was introduced to "fix" the `include` tag so the template could be cached.

I still like the concept but I am concerned it won't be accepted in this form. Might even be easier to sell it in the now defunct `include` tag ...

PR #1518, opened 2022-02-11
Fix kwarg parsing inconsistency with Liquid::C

charlespwd:
Liquid::C parses liquid filter arguments with dashes in them, Liquid does not.

For tags that accept kwargs and dumps them on the HTML tag, this is an important feature.

e.g.

```liquid
{{ img | image_tag: data-something: 'value!' }}
```

Without this change, Liquid would incorrectly parse the `data-something` kwarg as a single argument and skips over the invalid characters.

```
irb(main):025:0> pt = Liquid::Template.parse('{{ img | image_tag: data-src: "value!" }}', parse_context)
irb(main):026:0> pt.root.nodelist[0]
=>
#<Liquid::Variable:0x00007f80b82c74b0
 @filters=
  [["image_tag",
    [#<Liquid::VariableLookup:0x00007f80b82c6448 @command_flags=0, @lookups=["value"], @name="data-src">]]],
 # ...
 # NOTE how data-src is not the name of a variable, but of the kwarg
 key, and how value! got its ! stripped.
```

This PR fixes this by changing the regex we use for filter kwarg keys to also accept the `-` character.

See https://github.com/Shopify/theme-check/issues/539 for more context


test/integration/filter_kwarg_test.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+
+class HashOrderingTest < Minitest::Test
```
macournoyer, line 5:
```suggestion
class FilterKwargTest < Minitest::Test
```

macournoyer (approved):
Ran test on Liquid-C. 👍  🔥

dylanahsmith (changes requested):

dylanahsmith:
Note that `-` isn't supported in keyword argument names with the lax parser, which could cause some confusion if we start recommending this usage.

charlespwd:
Aren't the test run for all four settings (C&!C, Lax&Strict)? In which case is this not supported?

dylanahsmith (approved):

dylanahsmith:
Liquid variables are first attempted to be parsed with the strict parser, even in lax error mode.  If there is a strict parse failure, then it falls back to the lax parser when it isn't in the strict error mode.  If that fallback is needed due to an unrelated strict parse failure, then it can break the use of `-` in the keyword argument name.  It is a subtle inconsistency, which is confusing due to the lax parser actually being more strict about what it accepts for the keyword argument name.

PR #1528, opened 2022-03-01
Liquid::Drop: Expect @context to always be set

tjoyal:
Liquid::Drop to always assume `@context` is set.

While this is the direction I think we need to take, I'm not sure this is a safe change to introduce yet until further exploration of reports regarding issues with `@context` being absent or reset. 

I think I would still go ahead and extract these tests as a pre-PR as they better reflect the production behavior and desired code coverage.

Related to https://github.com/Shopify/liquid/issues/1205 (Execution paths missing @context)

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,12 @@
 # Liquid Change Log
 
+## 5.2.1 (unreleased)
+
+### Features
+
+### Fixes
+* Liquid::Drop: Expect @context to always be set (#1528) [Thierry Joyal]
+
 ## 5.2.0 2021-03-01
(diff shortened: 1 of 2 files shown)


tjoyal:
> if we want to be more cautious about breaking changes, then we could have a deprecation warning in Liquid::Drop#liquid_method_missing if @context is nil

👍 

Since it is very hard to test for these things, it might be a good path forward to move fast and safely.

I want to iron out as much as possible this is not an issue from the gem itself and then I'll update and send it in review.

PR #1533, opened 2022-03-04
Add Ruby 3.1 to the CI matrix

petergoldstein:
This PR adds Ruby 3.1 to the CI matrix.

Also quoted the `3.0` so it is properly interpreted as 3.0 rather than 3.  An unquoted `3.0` is truncated to `3`, loading the latest Ruby 3 - at time of writing Ruby 3.1.1.  Adding quotes ensures that a Ruby 3.0.x version is loaded for this CI matrix entry.

### DIFF
--- a/.github/workflows/liquid.yml
+++ b/.github/workflows/liquid.yml
@@ -7,7 +7,6 @@ jobs:
       matrix:
         entry:
           - { ruby: 2.5, allowed-failure: false } # minimum supported
-          - { ruby: '3.0', allowed-failure: false }
           - { ruby: 3.1, allowed-failure: false } # latest
           - { ruby: ruby-head, allowed-failure: true }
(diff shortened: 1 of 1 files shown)


.github/workflows/liquid.yml
```diff
@@ -7,7 +7,8 @@ jobs:
       matrix:
         entry:
           - { ruby: 2.5, allowed-failure: false } # minimum supported
-          - { ruby: 3.0, allowed-failure: false } # latest
+          - { ruby: '3.0', allowed-failure: false }
```
dylanahsmith, line 10: There should be no reason to test intermediate versions if there isn't version dependent code.  Testing the minimum supported will make sure we don't use too new a feature and testing the latest version will help catch deprecations or feature removals.

```suggestion
```

PR #1535, opened 2022-03-04
Update profiler clock

tjoyal:
While working on Liquid I've been getting some test flakiness:
```
  1) Failure:
ProfilerTest#test_profiling_supports_self_time [/Users/tjoyal/projects/liquid/test/integration/profiler_test.rb:204]:
Expected 0.0 to be > 0.
```

Investigating, I realized that when using CLOCK_MONOTONIC:
- on `x86_64-linux` I get a precision of **nanoseconds**
- on `arm64-darwin21` I get a precision of **microseconds**

Looking at the test behavior, it looks like the block would rarely be rendered fast enough so the recorded time is of `0.0`.

This PR proposes to pick witch clock is to be used so we can get **microseconds** precision in Darwin powered hardware.

I am usually not a fan of mutating "production code paths" for test purposes therefore we could alternatively look at this from the test itself and to add more expensive instructions or a small sleep. Since profiling code is about production accuracy, I am comfortable with the proposed changes here.

While I understand system level behaviors, this is for me a limited area of expertise and would appreciate feedback.

### Reproduction steps

Locally I can trip the flakiness quite rapidly with:
```
while rake test; do :; done 
```


dylanahsmith:
> I am usually not a fan of mutating "production code paths" for test purposes therefore we could alternatively look at this from the test itself and to add more expensive instructions or a small sleep.

Yeah, I think it would make more sense to use `with_custom_tag` to add something like a `clock_tick_wait` tag that could spin for a minimum measurable amount of time.  E.g.

```
start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
until Process.clock_gettime(Process::CLOCK_MONOTONIC) > start
  sleep(0.000_000_001)
end
```

which should work cross-platform without feature detection, flakiness or changes to the production code path.

PR #1536, opened 2022-03-04
Add artificial execution time in profiler tests

tjoyal:
While working on Liquid I've been getting some test flakiness:
```
  1) Failure:
ProfilerTest#test_profiling_supports_self_time [/Users/tjoyal/projects/liquid/test/integration/profiler_test.rb:204]:
Expected 0.0 to be > 0.
```

Investigating, I realized that when using CLOCK_MONOTONIC:
- on `x86_64-linux` I get a precision of **nanoseconds**
- on `arm64-darwin21` I get a precision of **microseconds**

Looking at the test behavior, it looks like the block would rarely be rendered fast enough so the recorded time is of `0.0`.

This PR introduce artificial wait time in the test object code paths.

Alternate take of https://github.com/Shopify/liquid/pull/1535

### DIFF
--- a/test/integration/profiler_test.rb
+++ b/test/integration/profiler_test.rb
@@ -3,6 +3,27 @@
 require 'test_helper'
 
 class ProfilerTest < Minitest::Test
+  class TestDrop < Liquid::Drop
+    def initialize(value)
+      super()
+      @value = value
+    end
+
+    def to_s
+      artificial_execution_time
+
+      @value
+    end
+
+    private
+
+    # Monotonic clock precision fluctuate based on the operating system
+    # By introducing a small sleep we ensure ourselves to register a non zero unit of time
+    def artificial_execution_time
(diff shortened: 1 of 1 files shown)


test/integration/profiler_test.rb
```diff
@@ -198,16 +219,22 @@ def test_profiling_marks_children_of_for_blocks
...
   end
 
   def test_profiling_supports_total_time
     t = Template.parse("{% if true %} {% increment test %} {{ test }} {% endif %}", profile: true)
-    t.render!
+    output = t.render!
+    assert_equal(" 0 1 ", output)
 
-    assert_operator(t.profiler[0].total_time, :>, 0)
+    assert_operator(t.profiler[0].total_time, :>, 0.0)
```
dylanahsmith, line 238: Shouldn't this also be using TestDrop to ensure that total_time doesn't end up being 0?

tjoyal: 100%

PR #1537, opened 2022-03-07
Consolidate contextualization of inputs via the context

tjoyal:
This PR is all over the place, I am sorry for that. I found it hard to separate isolated deliverables as there is a lot of inter-dependency at play.

This PR adds `Context#contextualize` which acts as a central hub to handle logic regarding transformation of variable into liquid aware representation. 

This abstraction handles `Proc` resolutions when needed, the call to `to_liquid` and assignment of context via `context=`.

### Changes

**Context:**
- Pushed contextualization from `find_variable` to `lookup_and_evaluate`. No behavior change specific to this class,  this will allow for `VariableLookup#evaluate` to re-use said logic.

**VariableLookup:**
- Main path can delegate the responsibility of contextualization to `Context#lookup_and_evaluate`, allowing some of the logic duplication to be removed.
- The "command methods" have to deal with manual contextualization have been updated to the unified code path. `Context#contextualize` logic properly handles `Proc` which solves for the newly added tests who would have been failing otherwise. We could look at moving this `send` call to `Context` but I opted to limit the changes in the current PR.

**StandardFilters:**
- There is a bug fix and breaking change due to the removal of `respond_to?(:to_liquid)`. The previous code would have left a possibly unexpected object exposed to the template which could have lead to unforeseen privilege given to the template. Procs were notably not being converted. 

### Behavior of unhandled Procs
Cases when `Proc` are not handled sometimes lead to exceptions, but other times they could lead to exposure of internal code structure (eg.: rendered as strings in the template such as`"#<Proc:0xXXXXXX /Users/tjoyal/projects/liquid/test/integration/standard_filter_test.rb:485>"`). 

### North star

The question of supporting `Proc` behavior is to be asked.

I think most of that “lazy evaluation” can be achieved with an implementation over `Liquid::Drop` wrapping the underlying implementation which needs to be delayed.

I’ll need to do some git archeology to learn more as to when it was introduced and if it still have merit to be maintained going forward.

### DIFF
--- a/History.md
+++ b/History.md
@@ -4,6 +4,7 @@
 
 ### Fixes
 * StandardFilter: Fix missing @context on iterations (#1525) [Thierry Joyal]
+* Consolidate contextualization of inputs via the context (#XXX) [Thierry Joyal]
 
 ### Deprecation
 * Condition#evaluate to require mandatory context argument in Liquid 6.0.0 (#1527) [Thierry Joyal]
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -187,16 +187,11 @@ module Liquid
       # path and find_index() is optimized in MRI to reduce object allocation
       index = @scopes.find_index { |s| s.key?(key) }
 
-      variable = if index
+      if index
         lookup_and_evaluate(@scopes[index], key, raise_on_not_found: raise_on_not_found)
       else
         try_variable_find_in_environments(key, raise_on_not_found: raise_on_not_found)
       end
-
-      variable         = variable.to_liquid
-      variable.context = self if variable.respond_to?(:context=)
-
-      variable
     end
 
     def lookup_and_evaluate(obj, key, raise_on_not_found: true)
@@ -207,9 +202,9 @@ module Liquid
       value = obj[key]
 
       if value.is_a?(Proc) && obj.respond_to?(:[]=)
-        obj[key] = value.arity == 0 ? value.call : value.call(self)
+        obj[key] = contextualize(value)
       else
-        value
+        contextualize(value)
       end
     end
 
@@ -226,6 +221,20 @@ module Liquid
 
     def tag_disabled?(tag_name)
       @disabled_tags.fetch(tag_name, 0) > 0
+    end
+
+    # Convert input objects into liquid aware representations
+    # Procs will be resolved
+    # Assigns the context (self) through context=
+    def contextualize(object)
+      if object.is_a?(Proc)
+        object = object.arity == 0 ? object.call : object.call(self)
+      end
+
+      object = object.to_liquid
+      object.context = self if object.respond_to?(:context=)
+
+      object
     end
 
     protected
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -582,8 +582,7 @@ module Liquid
 
       def each
         @input.each do |e|
-          e = e.respond_to?(:to_liquid) ? e.to_liquid : e
-          e.context = @context if e.respond_to?(:context=)
+          e = @context.contextualize(e)
           yield(e)
         end
       end
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -34,6 +34,10 @@ class TestDrop < Liquid::Drop
   def registers
     @context.registers
   end
+
+  def to_s
+    "TestDrop(value:#{@value})"
+  end
 end
 
 class TestModel
@@ -467,10 +471,38 @@ class StandardFiltersTest < Minitest::Test
   end
 
   def test_map_over_proc
-    drop  = TestDrop.new(value: "testfoo")
+    drop  = TestDrop.new(value: "123")
     p     = proc { drop }
     templ = '{{ procs | map: "value" }}'
-    assert_template_result("testfoo", templ, "procs" => [p])
+    assert_template_result("123", templ, "procs" => [p])
+  end
+
+  def test_join_over_proc
+    drop  = TestDrop.new(value: "123")
+    p     = proc { drop }
+    templ = '{{ procs | join }}'
+    assert_template_result('TestDrop(value:123)', templ, "procs" => [p])
+  end
+
+  def test_sort_over_proc
+    drop  = TestDrop.new(value: "123")
+    p     = proc { drop }
+    templ = '{{ procs | sort: "value" }}'
+    assert_template_result("TestDrop(value:123)", templ, "procs" => [p])
+  end
+
+  def test_where_over_proc
+    drop  = TestDrop.new(value: "123")
+    p     = proc { drop }
+    templ = '{{ procs | where: "value", "123" }}'
+    assert_template_result("TestDrop(value:123)", templ, "procs" => [p])
+  end
+
+  def test_uniq_over_proc
+    drop  = TestDrop.new(value: "123")
+    p     = proc { drop }
+    templ = '{{ procs | uniq }}'
+    assert_template_result("TestDrop(value:123)", templ, "procs" => [p])
   end
 
   def test_map_over_drops_returning_procs
@@ -888,7 +920,7 @@ class StandardFiltersTest < Minitest::Test
       { foo: "bar" },
       [{ "foo" => "bar" }, { "foo" => 123 }, { "foo" => nil }, { "foo" => true }, { "foo" => ["foo", "bar"] }],
       { 1 => "bar" },
-      ["foo", 123, nil, true, false, Drop, ["foo"], { foo: "bar" }],
+      ["foo", 123, nil, true, false, ["foo"], { foo: "bar" }],
     ]
     StandardFilters.public_instance_methods(false).each do |method|
       arg_count = @filters.method(method).arity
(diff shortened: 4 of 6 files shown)


History.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixes
 * StandardFilter: Fix missing @context on iterations (#1525) [Thierry Joyal]
+* Consolidate contextualization of inputs via the context (#XXX) [Thierry Joyal]
```
tjoyal, line 7: Due to the potential "breaking change" in `StandardFilters` and `VariableLookup` where `Proc` are now properly handled, we might to look at this to be similar to #1527 flagged as deprecation for 6.0.

I do not think anyone expected to have `Proc` to be exposed directly to templates.

For standard filters specifically, that `respond_to?(:to_liquid)` might have left non-liquid aware object in by potential design error. I could see use cases leveraging this unexpectedly. They have an easy path forward to migrate, but a deprecation warning might be of interest here.

lib/liquid/context.rb
```diff
@@ -207,9 +202,9 @@ def lookup_and_evaluate(obj, key, raise_on_not_found: true)
       value = obj[key]
 
       if value.is_a?(Proc) && obj.respond_to?(:[]=)
-        obj[key] = value.arity == 0 ? value.call : value.call(self)
+        obj[key] = contextualize(value)
```
tjoyal, line 205: I am not changing the memoization in this PR as it let to too large of a change with even more unrelated code alterations. 

I feel we should look at:
- Exploring memoization everywhere (eg.: for code path like StandardFilters to benefit from it)
- Not only memoize `if value.is_a?(Proc)`. There are some test failures appearing when trying to remove the condition, can be iterated further in subsequent work if required.

tjoyal, line 205: This would have been clearer, but achieve the same behaviour with more code. No strong opinion on my end.
```suggestion
        value = contextualize(value)
        obj[key] = value
        value
```

test/integration/standard_filter_test.rb
```diff
@@ -467,10 +471,38 @@ def test_sort_calls_to_liquid
...
+    p     = proc { drop }
+    templ = '{{ procs | where: "value", "123" }}'
+    assert_template_result("TestDrop(value:123)", templ, "procs" => [p])
+  end
+
+  def test_uniq_over_proc
+    drop  = TestDrop.new(value: "123")
+    p     = proc { drop }
+    templ = '{{ procs | uniq }}'
+    assert_template_result("TestDrop(value:123)", templ, "procs" => [p])
```
tjoyal, line 505: These tests are to cover the new behavior for Proc over filters where historically they would have raised on missing method `to_liquid`

test/integration/tags/standard_tag_test.rb
```diff
@@ -272,14 +272,36 @@ def test_multiple_named_cycles_with_names_from_context
...
+    assert_template_result("11", "{{ hash.first.last }}", assigns)
+
+    assert_template_result("", "{{ hash.last }}", assigns)
   end
 
-  def test_size_of_hash
-    assigns = { "hash" => { a: 1, b: 2, c: 3, d: 4 } }
-    assert_template_result('hash has 4 elements', "hash has {{ hash.size }} elements", assigns)
+  def test_command_methods_with_proc
+    skip("Liquid-C does not properly resolve Procs in with command methods") if ENV['LIQUID_C'] == '1'
```
tjoyal, line 298: Liquid-C does not pass this test. Skipping for now as I think we need to iterate further on Proc handling without blocking iterative work.

lib/liquid/standardfilters.rb
```diff
@@ -582,8 +582,7 @@ def empty?
 
       def each
         @input.each do |e|
-          e = e.respond_to?(:to_liquid) ? e.to_liquid : e
-          e.context = @context if e.respond_to?(:context=)
+          e = @context.contextualize(e)
```
tjoyal, line 585: As noted in the description, this change no longer exclude `e.respond_to?(:to_liquid)`

History.md
dylanahsmith: "contextualization" is a new concept, which makes this change log entry very unclear about what is being fixed.  That seems like just a symptom of the fact that this PR combines multiple changes, a new feature (added method), behaviour fixes and some refactoring (which doesn't need to be in the changelog).

dylanahsmith (changes requested):

lib/liquid/standardfilters.rb
tjoyal: I can have a closer look. Making it unconditional was making a few tests of `StandardFilters` (specifically around `map`) no longer pass since there is Proc handling in said method and `Proc` does not respond `to_liquid`.

If we wanted to make the call unconditional in isolation, I will have to either:
1. Move support for `Proc` here (similar to the current PR, extending support for all use cases)
2. Drop support for `Proc` in `StandardFilters#map` (potentially what we will want to do down the road?)

I think going with 1 is a good tradeoff for now?

lib/liquid/context.rb
```diff
@@ -228,6 +223,20 @@ def tag_disabled?(tag_name)
       @disabled_tags.fetch(tag_name, 0) > 0
     end
 
+    # Convert input objects into liquid aware representations
+    # Procs will be resolved
+    # Assigns the context (self) through context=
+    def contextualize(object)
```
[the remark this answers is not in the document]
macournoyer: I get from the name that is sets `context`, but it's not clear that it converts it to a "liquid aware representations". Maybe that means the method is doing too much?

Could we instead move the "contextualization" to `Drop#invoke_drop`: add a `context` arg to `invoke_drop`, and wrap procs in drops? Or is there are use case outside of drops & procs for this?

macournoyer (commented):
Is `context=` only useful for drops?

Ideally I think it would be better to allow passing the `context` to drops in `Drop#invoke_drop(method_or_key, context)`, and get rid of `context=`.

I assume we're only keeping `context=` around because too many filters & drops (outside of Liquid) depend on it?

lib/liquid/standardfilters.rb
dylanahsmith: Let's assume that a Proc type is the only non-liquid value that needs to be handled here.  In that case, we can use a `if e.is_a?(Proc)` check instead of `e.respond_to?(:to_liquid)`.  We can then preserve backwards compatibility with a deprecation warning for a Proc.  Doing that here would also give a desirable deprecation warning for the `e = e.call if e.is_a?(Proc)` line in `StandardFilters#map`.

This deprecation should be communicated as deprecating support for having a Proc item in an enumerable.

On the other hand, I think we should continue supporting a Proc value in liquid Hash objects and should make sure to memoize the result of evaluating the Proc.  This means that the property lookup in `StandardFilters#map` should memoize the result of calling any Proc value.

test/integration/standard_filter_test.rb
```diff
@@ -888,7 +920,7 @@ def test_all_filters_never_raise_non_liquid_exception
       { foo: "bar" },
       [{ "foo" => "bar" }, { "foo" => 123 }, { "foo" => nil }, { "foo" => true }, { "foo" => ["foo", "bar"] }],
       { 1 => "bar" },
-      ["foo", 123, nil, true, false, Drop, ["foo"], { foo: "bar" }],
+      ["foo", 123, nil, true, false, ["foo"], { foo: "bar" }],
```
dylanahsmith, line 923: Perhaps the intention was to test against a drop instance here, which should be supported

```suggestion
      ["foo", 123, nil, true, false, Drop.new, ["foo"], { foo: "bar" }],
```

lib/liquid/standardfilters.rb
dylanahsmith: There are other filters that support looking properties, which should be able to just use `context.lookup_and_evaluate(item, property, raise_on_not_found: false)`.  Basically, that should be used as the helper method to consolidate Proc Hash lookup support.

That helper method could be used in the `map` filter implementation as well, although we still may want to keep the `r.is_a?(Proc) ? r.call : r` line to unconditionally evaluate the Proc even when memoization isn't supported, for backwards compatibility.

dylanahsmith:
> Ideally I think it would be better to allow passing the `context` to drops in `Drop#invoke_drop(method_or_key, context)`, and get rid of `context=`.

Would that assign the context every time the drop is invoked?  I suppose that would avoid having to incur this overhead for other types, although having that overhead on property access on drops still seems undesirable.

Also, we can't just get rid of `context=` without a deprecation cycle, since that could break liquid extensions that use it.  It would also mean that the context would be set more lazily, which makes it less obvious to me whether that could result in other breaking changes.

Instead, I want to get to the point where `to_liquid` takes a context, so that not only would context setting overhead be isolated to drops, but it would also be isolated to initialization of the drops and avoid having drop instances that are in an invalid state (i.e. more obviously correct code).

dylanahsmith:
> Is `context=` only useful for drops?

Liquid seems to use duck typing for interactions with Liquid::Drop, rather than checks like `.is_a?(Liquid::Drop)`.  I'm not sure if this flexibility is actually used to have something like a Liquid::Drop that doesn't actually inherit from that class.  However, I would expect that typically it is only a Liquid::Drop derived class that would use `context=`.

macournoyer:
> Instead, I want to get to the point where `to_liquid` takes a context, so that not only would context setting overhead be isolated to drops, but it would also be isolated to initialization of the drops and avoid having drop instances that are in an invalid state (i.e. more obviously correct code).

Ah that'd really good! I totally agree this is much better: `to_liquid(context)`.

PR #1540, opened 2022-03-12
Remove redundant regexp

Watson1978:
The same processing is possible without using regular expressions.

−               | before   | after   | result
--               | --       | --      | --
parse            | 61.006   | 61.975  | 1.016x
render           | 197.095  | 197.103 | -
parse & render   | 44.846   | 45.180  | 1.008x

### Environment
- MacBook Pro (14 inch, 2021)
- macOS 12.3 beta
- Apple M1 Max
- Ruby 3.0.3

### Before
```
$ ruby benchmark.rb

Running benchmark for 10 seconds (with 5 seconds warmup).

              parse:     6.000  i/100ms
             render:    19.000  i/100ms
     parse & render:     4.000  i/100ms
              parse:     61.006  (± 0.0%) i/s -    612.000  in  10.032337s
             render:    197.095  (± 1.5%) i/s -      1.976k in  10.027538s
     parse & render:     44.846  (± 2.2%) i/s -    452.000  in  10.080806s
```

### After
```
$ ruby benchmark.rb

Running benchmark for 10 seconds (with 5 seconds warmup).

              parse:     6.000  i/100ms
             render:    19.000  i/100ms
     parse & render:     4.000  i/100ms
              parse:     61.975  (± 0.0%) i/s -    624.000  in  10.069036s
             render:    197.103  (± 1.0%) i/s -      1.976k in  10.026615s
     parse & render:     45.180  (± 0.0%) i/s -    452.000  in  10.005335s
```

### DIFF
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -37,7 +37,7 @@ module Liquid
 
     private def parse_for_liquid_tag(tokenizer, parse_context)
       while (token = tokenizer.shift)
-        unless token.empty? || token =~ WhitespaceOrNothing
+        unless token.empty? || token.match?(WhitespaceOrNothing)
           unless token =~ LiquidTagToken
             # line isn't empty but didn't match tag syntax, yield and let the
             # caller raise a syntax error
@@ -150,7 +150,7 @@ module Liquid
           end
           parse_context.trim_whitespace = false
           @nodelist << token
-          @blank &&= !!(token =~ WhitespaceOrNothing)
+          @blank &&= token.match?(WhitespaceOrNothing)
         end
         parse_context.line_number = tokenizer.line_number
       end
--- a/lib/liquid/block_body.rb
+++ b/lib/liquid/block_body.rb
@@ -7,7 +7,6 @@ module Liquid
     LiquidTagToken      = /\A\s*(\w+)\s*(.*?)\z/o
     FullToken           = /\A#{TagStart}#{WhitespaceControl}?(\s*)(\w+)(\s*)(.*?)#{WhitespaceControl}?#{TagEnd}\z/om
     ContentOfVariable   = /\A#{VariableStart}#{WhitespaceControl}?(.*?)#{WhitespaceControl}?#{VariableEnd}\z/om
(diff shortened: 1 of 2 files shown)


lib/liquid/block_body.rb
```diff
@@ -37,7 +36,7 @@ def freeze
 
     private def parse_for_liquid_tag(tokenizer, parse_context)
       while (token = tokenizer.shift)
-        unless token.empty? || token =~ WhitespaceOrNothing
+        unless token.strip.empty?
```
macournoyer, line 39: This will cause an allocation. I'm not sure the the minimal perf gain justifies this.

macournoyer, line 39: This will cause an allocation. I'm not sure the the minimal perf gain justifies this.

lib/liquid/variable_lookup.rb
```diff
@@ -15,8 +14,8 @@ def initialize(markup)
       lookups = markup.scan(VariableParser)
 
       name = lookups.shift
-      if name =~ SQUARE_BRACKETED
-        name = Expression.parse(Regexp.last_match(1))
+      if name&.start_with?('[') && name&.end_with?(']')
```
macournoyer, line 17: I suspect the perf gain is only due to this (and bellow). This looks like a nice optimization to me.

macournoyer, line 17: I suspect the perf gain is only due to this (and bellow). This looks like a nice optimization to me.

macournoyer (changes requested):
I don't think `token.strip` is a good idea. If you revert this change, rest looks good to me.

Watson1978:
> I don't think token.strip is a good idea. If you revert this change, rest looks good to me.

OK, reverted at https://github.com/Shopify/liquid/pull/1540/commits/22568080b1cb6f19969d55d68dc39b5c04a37b17

However, I think it should use `String#match?` instead because it doesn't use backreference of `String#=~`. 
See https://github.com/Shopify/liquid/pull/1540/commits/fad58ef436bb9a516ea8364d9398cb6f6876d784

macournoyer (approved):
Thanks a lot! That's awesome 🙏

PR #1543, opened 2022-03-14
Liquid docs content swarm

mllemango:


lib/liquid/standardfilters.rb
```diff
@@ -472,6 +472,12 @@ def floor(input)
       raise Liquid::FloatDomainError, e.message
     end
 
+    # @public_docs
+    # @syntax {{ number | at_least: number }}
+    # @summary Limits a number to a minimum value.
+    # @type filter
+    # @category Math
+    # @return number
```
geddski, line 480: Is there an `@return` tag? In the [readme](https://github.com/Shopify/storefront-renderer/blob/master/lib/liquid_yard/README.md) all I see is `@return_types`.

geddski: oh nice!

PR #1550, opened 2022-03-15
Fork latest liquid release with namespace Liquid5

willng-braze:
In order to include multiple versions of liquid in platform, they need separate namespaces.  The plan is to copy the code into https://github.com/braze-inc/liquid when everything looks good.

Test plan:
* Unit tests passes (bundle exec rake)


Gemfile
```diff
@@ -22,7 +22,7 @@ group :test do
   gem 'rubocop-shopify', '~> 1.0.7', require: false
   gem 'rubocop-performance', require: false
 
-  platform :mri, :truffleruby do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'master'
-  end
+  # platform :mri, :truffleruby do
```
willng-braze, line 25: I commented this out.  I assume we don't need to deal w/ liquid-c

PR #1554, opened 2022-03-17
Update changelog & bump version for 5.3.0 release

macournoyer:
Cutting a release before we introduce breaking changes: https://github.com/Shopify/liquid/pull/1553#issuecomment-1069289810

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,6 +1,6 @@
 # Liquid Change Log
 
-## 5.3.0 2022-03-18
+## 5.3.0 2022-03-22
 
 ### Fixes
 * StandardFilter: Fix missing @context on iterations (#1525) [Thierry Joyal]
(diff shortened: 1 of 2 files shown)


History.md
```diff
@@ -1,9 +1,11 @@
 # Liquid Change Log
 
-## 5.3.0 (unreleased)
+## 5.3.0 2022-03-22
 
 ### Fixes
 * StandardFilter: Fix missing @context on iterations (#1525) [Thierry Joyal]
+* Test under Ruby 3.1 (#1533) [petergoldstein]
```
dylanahsmith, line 7: This isn't really a change from the user's perspective.  The code already worked in ruby 3.1, so this isn't really a fix.  Users of the library generally don't care about tests being added.

Issue #1555, opened 2022-03-22
Adding a new "where" filters


bakura10:
Hi,

The where filter is very useful, but I would suggest having a few improvements:

## where_not

This filter would do the opposite of where filter:

```
{% assign blocks_to_render = section.blocks | where_not: 'type', 'divider' %}
```

## Allow dot notation

With metafields and dynamic connection, it is increasingly common that merchant create a section with blocks, and bound the content with metafields.

For instance, let's imagine a "testimonials" section where the merchant would create 5 blocks. Some products may have 5, 4, 3, 2 or even no active blocks, depending on the product. This currently makes rendering very cumbersome. I would love having a dot notation. For instance:

```
{% assign rendered_blocks = section.blocks | where: 'settings.testimonial' %}

{% assign rendered_blocks = section.blocks | where: 'settings.product' %}
```

The idea here is to return actual block instances, but that would be filtered. This would allow to generate section code based on the actual content indicated by the merchant.

andershagbard:
A PR for dot notation in where filter is open here https://github.com/Shopify/liquid/pull/1436

FalconPilot:
Is there any update on this? I know it's been 1 year without a reply, but it's still relevant and would be an insane help to have on Shopify :)

MLDMoritz:
where_not is really required! There is no way to filter for nested properties that are not empty!

This filters section blocks where row_content is empty:
`{% assign blocks = section.blocks | map: 'settings' | where: 'row_content', empty %}`

There is absolutely no way to filter blocks where row_content is not empty!

PR #1573, opened 2022-05-12
Add reject filter

andershagbard:
This filter adds the inverse of the `where` filter.

This filter also allows deep searching with dot notation.

Idea from: https://github.com/Shopify/liquid/issues/1555


andershagbard:
Not sure why Rubocop is failing 🤷🏻‍♂️

bakura10:
I’m not a ruby developer but it seems you also added a nice feature on the reject filter allowing to filter nested objects. This should most likely be backported to the where filter as well.

andershagbard:
> I’m not a ruby developer but it seems you also added a nice feature on the reject filter allowing to filter nested objects. This should most likely be backported to the where filter as well. 

That has proven a bit harder, as there is some backward compatibility we need to take care of: https://github.com/Shopify/liquid/pull/1436

lib/liquid/standardfilters.rb
```diff
@@ -234,6 +234,29 @@ def where(input, property, target_value = nil)
...
+          if target_value.nil?
+            !value
+          else
+            value == target_value
+          end
+        else
+          true
+        end
+      end
+    end
```
luiz-m-affonso, line 258:
```suggestion
    def reject(input, properties, target_value = nil)
      raise_property_error(properties) unless properties.is_a?(String)

      properties = properties.to_s.split('.')
      array = InputIterator.new(input, context)
      check_array_inputs(array)
    end
    
    def check_array_inputs(array)
      array.reject do |item|
        if item.is_a?(Hash)
          value = item.dig(*properties)
          if target_value.nil?
            !value
          end
          return value == target_value
        else
          true
        end
      end
    end
```
```

luiz-m-affonso: Just a few changes in the method, what do you think of this refactor? Good job

andershagbard: Adding a new function here will add a new filter in Liquid. I don't think this change is necessary.

andershagbard:
Solved in https://github.com/Shopify/liquid/pull/1869

PR #1576, opened 2022-05-18
Add yard tags

jake-olney:
## This PR:

As part of the [Liquid Reference Redesign](https://vault.shopify.io/projects/20952), this PR adds yard tags for drops/filters/tags that are part of the Liquid documentation on `shopify.dev`.

  - Fixes https://github.com/Shopify/shopify-dev/issues/17314

PR #1578, opened 2022-05-18
Drop support for Ruby 2.5 and 2.6

andyw8:
@jake-olney encountered an issue with RuboCop:

https://github.com/Shopify/liquid/pull/1576

Upgrading rubocop-shopify solves it but it is not compatible with older Rubies.

@macournoyer suggested we should drop these as both are EOL: https://endoflife.date/ruby


Gemfile
```diff
@@ -18,8 +18,7 @@ group :benchmark, :test do
 end
 
 group :test do
-  gem 'rubocop', '~> 1.4', require: false
```
andyw8, line 21: I removed this since `rubocop` is already a dependency of `rubocop-shopify`.

macournoyer (approved):
I see no reasons to keep < 2.7 support around. @dylanahsmith do you know if we're using this anywhere w/ Ruby < 2.7?

History.md
```diff
@@ -2,6 +2,9 @@
 
 ## 5.4.0 (unreleased)
 
+### Breaking Changes
+* Require Ruby >= 2.7 (#1578) [Andy Waite]
```
andyw8, line 6: Arguably this should be a major release (6.0), depending on how strict we're being with semver.

dylanahsmith: There is already a "Changed" section with a potentially breaking change, so this should probably be combined with that section.

dylanahsmith:
@macournoyer it shouldn't be a problem for us.  It could affect users of the system version of ruby.

History.md
andyw8: I can move it, but I noticed the previous Ruby bump was listed under that section name:

```
### Breaking Changes
* Require Ruby >= 2.5 (#1131, #1310) [Mike Angell, Dylan Thacker-Smith]
```

andyw8: 👍 Updated in 3bbb7aa7baba7602caa95292d809c3fbfe9d57c5

dylanahsmith (approved):

PR #1595, opened 2022-07-13
Clarify sort_natural and fix unless syntax

jake-olney:
## This PR:

  - Updates the `sort_natural` filter to note that it should only be used on strings as all items are converted to strings, which leads to unexpected numerical sorting.
  - Fixes a typo with the syntax for `unless`.


shainaraskas (approved):
beautiful (and what a small PR!)

shainaraskas (approved):
still good 👍

PR #1596, opened 2022-07-14
Add variable_name_expr to render's ParseTreeVisitor

charlespwd:
Solves Shopify/theme-check#582

`icon` should be visited in the render tag for the following snippet:

```liquid
{% assign icon = 'warning' %}
{% render 'icon' with icon %}
```

This allows us to properly see that the `icon` assign was used, and thus not flag an UnusedAssign when it is.

### DIFF
--- a/History.md
+++ b/History.md
@@ -11,6 +11,7 @@
 ### Fixes
 * `PartialCache` now shares snippet cache with subcontexts by default (#1553) [Chris AtLee]
 * Hash registers no longer leak into subcontexts as static registers (#1564) [Chris AtLee]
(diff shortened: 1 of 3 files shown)


macournoyer (approved):
Change looks good to me.

I assume it work with `render block` for app blocks too. Do you care about the `block` being visited in this case?

This is monkeypatched in SFR: https://github.com/Shopify/storefront-renderer/blob/master/app/liquid/tags/render_tag.rb#L7-L22

charlespwd:
@macournoyer, good catch. Had to re-add it in https://github.com/Shopify/theme-check/blob/main/lib/theme_check/tags.rb#L170-L177.

I'll patch SFR as well.

PR #1599, opened 2022-07-21
Expose list of filter methods

zzooeeyy:
We currently don't have a way to retrieve a list of filter methods. 
Tags could be retrieved from `Liquid::Template.tags`, exposing filter_methods so we could retrieve the set of filters from `Liquid::Template.filters`


zzooeeyy:
PR with comments addressed is here: https://github.com/Shopify/liquid/pull/1600

PR #1607, opened 2022-08-24
Raise Liquid::SyntaxError instead of NoMethodError for invalid range

### LINKED ISSUE #1604: Weird range handling
`Template.parse("{{ (true..false) }}")` throws a NoMethodError `.to_i` on TrueClass.

I had a partial fix in https://github.com/Shopify/liquid/tree/fix/range-weird-type-handling 

But, it looks like that's breaking lax parse with an extra dot by throwing an Liquid::ArgumentError instead of rendering.

Don't really have more time to put on this right now so I'm opening an issue instead :D

dylanahsmith:
Fixes #1604


charlespwd (approved):
LGTM! ty

PR #1609, opened 2022-08-30
Add a variable_name method Increment and Decrement tags objects

dylanahsmith:
In order to expose this state when using the parse tree, without having to resort to internal coupling like reading instance variables.

### DIFF
--- a/lib/liquid/tags/decrement.rb
+++ b/lib/liquid/tags/decrement.rb
@@ -19,15 +19,18 @@ module Liquid
   #   {% decrement variable_name %}
   # @liquid_syntax_keyword variable_name The name of the variable being decremented.
   class Decrement < Tag
+    attr_reader :variable_name
+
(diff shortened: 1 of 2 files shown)


lib/liquid/tags/decrement.rb
```diff
@@ -19,15 +19,18 @@ module Liquid
...
+    attr_reader :variable_name
+
     def initialize(tag_name, markup, options)
       super
-      @variable = markup.strip
+      @variable_name = markup.strip
     end
 
     def render_to_output_buffer(context, output)
-      value = context.environments.first[@variable] ||= 0
```
dylanahsmith, line 28: The assignment part of `||= 0` is unnecessary, since it is assigned again after decrementing.

ianks (approved):

ianks:
cc: @Maaarcocr

PR #1610, opened 2022-08-31
Stop using Liquid::Expression.parse for integration testing

dylanahsmith:
~~Depends on https://github.com/Shopify/liquid-c/pull/179 to fix the test error~~

## Problem

liquid-c stopped patching Liquid::Expression.parse in https://github.com/Shopify/liquid-c/pull/178 since it was too implicit, so integration tests using Liquid::Expression.parse weren't actually being shared across implementations.

## Solution

Use assert_template_result and assert_match_syntax_error for integration testing, which will better enable us to share tests using it across liquid implementations.

### DIFF
--- a/test/integration/expression_test.rb
+++ b/test/integration/expression_test.rb
@@ -4,48 +4,46 @@ require 'test_helper'
 
 class ExpressionTest < Minitest::Test
   def test_keyword_literals
-    assert_equal(true, parse_and_eval("true"))
-    assert_equal(true, parse_and_eval(" true "))
+    assert_template_result("true", "{{ true }}")
+    assert_expression_result(true, "true")
   end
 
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -22,6 +22,6 @@ group :test do
   gem 'rubocop-performance', require: false
 
   platform :mri, :truffleruby do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'master'
+    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'invalid-range-literal'
```
dylanahsmith, line 25: I'll remove this once the dependent liquid-c PR is merged.

samdoiron (approved):

PR #1615, opened 2022-09-08
Expose alias_name and for_loop? in render tag

Maaarcocr:


lib/liquid/tags/render.rb
```diff
@@ -45,7 +45,7 @@ def initialize(tag_name, markup, options)
       @alias_name = Regexp.last_match(6)
       @variable_name_expr = variable_name ? parse_expression(variable_name) : nil
       @template_name_expr = parse_expression(template_name)
-      @for = (with_or_for == FOR)
+      @for_loop? = (with_or_for == FOR)
```
dylanahsmith, line 48: I don't think instance variables can have a trailing question mark.  Instead of using `attr_reader`, you can just define a `for_loop?` method.  E.g.

```ruby
    def for_loop?
      @is_for_loop
    end
```

```suggestion
      @is_for_loop = (with_or_for == FOR)
```

lib/liquid/tags/render.rb
```diff
@@ -31,7 +31,7 @@ class Render < Tag
 
     disable_tags "include"
 
-    attr_reader :template_name_expr, :variable_name_expr, :attributes
+    attr_reader :template_name_expr, :variable_name_expr, :attributes, :alias_name, :is_for_loop
```
dylanahsmith, line 34: I think a predicate method would still be more idiomatic.  We shouldn't let instance variable name restrictions affect the desired API

```suggestion
    attr_reader :template_name_expr, :variable_name_expr, :attributes, :alias_name
    
    def for_loop?
      @is_for_loop
    end
```

dylanahsmith (approved):

PR #1617, opened 2022-09-09
Support big integers in truncatewords and slice filters

dylanahsmith:
In a low-level implementation of liquid, I would like to treat conversion of filter arguments to `slice`, `truncate` and `truncatewords` with a saturating conversion to an i32 or u32.  Having these work consistently also allows this conversion logic to be shared across these filters.

From a high-level perspective, the user would prefer if these were handled without an error for use cases like truncations and slicing.  A saturating conversion seems like it would give the expected behaviour, considering that the arrays and strings themselves shouldn't be this large.

As such, I'm changing the reference implementation to adopt this behaviour.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -6,7 +6,7 @@ require 'bigdecimal'
 
 module Liquid
   module StandardFilters
-    MAX_INT = (1 << 31) - 1
+    MAX_I32 = (1 << 31) - 1
     HTML_ESCAPE = {
       '&' => '&amp;',
       '>' => '&gt;',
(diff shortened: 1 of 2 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -239,9 +255,9 @@ def truncatewords(input, words = 15, truncate_string = "...")
       wordlist = begin
         input.split(" ", words + 1)
       rescue RangeError
-        raise if words + 1 < MAX_INT
-        # e.g. integer #{words} too big to convert to `int'
-        raise Liquid::ArgumentError, "integer #{words} too big for truncatewords"
+        # integer too big for Strin#split, but we can semantically assume no truncation is needed
```
samdoiron, line 258: Nit: Typo

samdoiron (approved):

PR #1628, opened 2022-09-28
Update escape filter reference summary

liamgriffin:
Added a more complete description of what the `escape` Liquid filter achieves, and how it behaves on `/api/liquid/filters` I was [advised in the Liquid slack channel](https://shopify.slack.com/archives/C07F3LRPF/p1664195876822159) to borrow some context from the [description of `CGI.escapeHTML` in Ruby](https://ruby-doc.org/stdlib-2.4.3/libdoc/cgi/rdoc/CGI/Util.html#method-i-escapeHTML), as it has the same behaviour. 

Fixes issue in dev docs https://github.com/Shopify/shopify-dev/pull/26494

Before:

After:

CC: @shainaraskas

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -73,7 +73,7 @@ module Liquid
     # @liquid_type filter
     # @liquid_category string
     # @liquid_summary
-    #   Escapes a string.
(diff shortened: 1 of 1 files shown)


shainaraskas (approved):
looks perfect!

Issue #1314, opened 2020-10-14
Add support for specifying filters for parsing


dylanahsmith:
Currently filters are only provided as a render option, which prevents us from checking if a filter call is valid during parsing and providing faster feedback about a bug in a template.

Also, we would like to use information about what filters will be available, along with what parameters they take, in liquid-c.  That will allow us to skip render time checks to see if a filter is invokable and avoid the need to translate ArgumentError exceptions to Liquid::ArgumentError from passing in the wrong number of arguments.

PR #1629, opened 2022-09-29
Add ability to pass tags to parse_context

mattdrose:
We have a gem that's using liquid to parse templates. The problem is that the consuming app is also using liquid, and their happens to be tag naming conflicts. Since liquid uses a single global tags registry (`Liquid::Template.tags`), the tags in our gem get clobbered.

This PR adds the ability for our gem to create and manage its own tag registry, and pass it as a part of the `parse_context` when building out the node list.

Example
```ruby
tags = Liquid::Template::TagRegistry.new
tags['my_tag'] = MyTag

# `tags` will take precedence over any tags registered in the global registry
Liquid::Template.parse("{{ my_tag }}", tags: tags)
```

### DIFF
--- a/test/integration/tag_test.rb
+++ b/test/integration/tag_test.rb
@@ -62,8 +62,7 @@ class TagTest < Minitest::Test
       end
     end
 
-    tags_overwrite = Liquid::Template::TagRegistry.new
-    tags_overwrite['dynamic_tag'] = new_tag
+    tags_overwrite = { 'dynamic_tag' => new_tag }
 
     with_custom_tag('static_tag', static_tag) do
       with_custom_tag('dynamic_tag', original_tag) do


test/integration/tag_test.rb
```diff
@@ -42,4 +42,35 @@ def render(*)
...
+    tags_overwrite = Liquid::Template::TagRegistry.new
+    tags_overwrite[tag_name] = new_tag
+
+    with_custom_tag(tag_name, original_tag) do
+      liquid = "{% #{tag_name} %} {% end#{tag_name} %}"
+
+      template = Liquid::Template.parse(liquid)
+      assert_equal('original_tag', template.render)
+
+      unless ENV['LIQUID_C'] == '1'
```
mattdrose, line 70: I tried to get this working with `liquid-c`, but I wasn't comfortable enough with the codebase. We only need this for parsing and not for rendering, so selfishly there's no need to get this working in `liquid-c` for our needs. Still, if this gets merged, it would be best to have a 1:1 relationship with `liquid-c`. Open to thoughts/ideas/contributions

lib/liquid/parse_context.rb
```diff
@@ -2,14 +2,15 @@
...
-    attr_accessor :locale, :line_number, :trim_whitespace, :depth
+    attr_accessor :locale, :line_number, :trim_whitespace, :depth, :tags
     attr_reader :partial, :warnings, :error_mode
 
     def initialize(options = {})
       @template_options = options ? options.dup : {}
 
       @locale   = @template_options[:locale] ||= I18n.new
       @warnings = []
+      @tags = @template_options[:tags]
```
dylanahsmith, line 13: Default to the global tags registry so it can be used unconditionally

```suggestion
      @tags = @template_options[:tags] || Liquid::Template.tags
```

lib/liquid/block_body.rb
```diff
@@ -45,7 +45,7 @@ def freeze
           end
           tag_name = Regexp.last_match(1)
           markup   = Regexp.last_match(2)
-          unless (tag = registered_tags[tag_name])
+          unless (tag = registered_tag(tag_name, parse_context))
```
dylanahsmith, line 48: We should be able to use `parse_context.tags` unconditionally, which would simplify this to

```suggestion
          unless (tag = parse_context.tags[tag_name])
```

we could then get rid of the `registered_tags` method, which used to just be there for monkey patching, but we have since got ridden dependence on it.

test/integration/tag_test.rb
dylanahsmith: I've opened https://github.com/Shopify/liquid-c/pull/181 to add liquid-c support for it.

mattdrose:
@dylanahsmith While tophatting I realised your changes overwrote all core tags, rather than giving precedence only if said tag was passed. I updated the test to capture this. Your changes to liquid-c still work here

test/integration/tag_test.rb
```diff
@@ -42,4 +42,39 @@ def render(*)
...
+      end
+    end
+
+    new_tag = Class.new(Tag) do
+      def render(*)
+        'new_tag'
+      end
+    end
+
+    tags_overwrite = Liquid::Template::TagRegistry.new
```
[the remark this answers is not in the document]
mattdrose: Yup, I can make this a hash. I wasn't sure why TagRegistry was caching the way it was—I figured it was related to performance benefits, so I used the same method

test/integration/tag_test.rb
```diff
@@ -42,4 +42,39 @@ def render(*)
...
+    tags_overwrite['dynamic_tag'] = new_tag
+
+    with_custom_tag('static_tag', static_tag) do
+      with_custom_tag('dynamic_tag', original_tag) do
+        liquid = '{% static_tag %} {% dynamic_tag %}'
+
+        template = Liquid::Template.parse(liquid)
+        assert_equal('static_tag original_tag', template.render)
+
+        template_with_overwrite = Liquid::Template.parse(liquid, tags: tags_overwrite)
```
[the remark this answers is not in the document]
mattdrose: It's subjective, but `tags` felt intuitive to me. In fact, the first thing I tried was passing `tags` to `parse_context` before checking whether the functionality existed or not.

Your interpretation of the PR description sounds correct. The main reason why we can't overwrite all tags is because core liquid tags use the same registry as custom tags. If this overwrote all custom tags, it would solve the problem.

I originally ran down the path of having two separate registries—one for core tags and one for custom tags, but the implementation started growing complex and this was a simpler solution.

```
the global namespace could still unintentionally inherit tags
```

This is a good point since templates are user input. We wouldn't want a false positive signal that a tag was parsed when there may be no rendering support. Having two separate registries for core and custom might be necessary.

dylanahsmith:
> Looking at filters, I think this behaviour swims down stream with how filters currently work.

The way filters are specified only for rendering is an existing issue because they are specified during rendering instead of during parsing (https://github.com/Shopify/liquid/issues/1314).  I would like to move away from this approach.

> Considering this change aligns with how filters currently work, how do you feel about merging it in and considering a wholistic approach separately?

As noted above, this wouldn't really be consistent with filters, nor do we actually want that.  Also, introducing another approach just to deprecate that in favour of what we currently want seems like it is more disruptive.

Also, it seems like this inheritance

> The prioritisation goes Local (passed inline) overwrites Global (declared in consumer) overwrites Standard (declared in Liquid).

isn't actually what we want for our use case or for the general use case of being able to specify liquid extensions in a gem, without interference with separate application specific usage of this liquid library.

PR #1633, opened 2022-10-03
fix tablerow drop's last attribute with missing cols param

ggmichaelgo:
When `cols` param is missing from the `tablerow` tag, `tablerowloop#col_last` always returns `false`.

Liquid template:
```liquid
{% tablerow i in (1..2) %}
  {{ tablerowloop.col_last }}
{% endtablerow %}
```

Expected output:
```
<tr class="row1">
<td class="col1">
false
</td><td class="col2">
true
</td></tr>
```

Current output:
```
<tr class="row1">
<td class="col1">
false
</td><td class="col2">
false
</td></tr>
```

### DIFF
--- a/lib/liquid/tags/table_row.rb
+++ b/lib/liquid/tags/table_row.rb
@@ -51,7 +51,7 @@ module Liquid
       collection = Utils.slice_collection(collection, from, to)
       length     = collection.length
 
-      cols = context.evaluate(@attributes['cols']).to_i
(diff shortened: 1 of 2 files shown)


ggmichaelgo:
ping @Shopify/guardians-of-the-liquid

dylanahsmith:
Your PR descriptions current output is a copy of your expected output.  The actual current output is

```
<tr class="row1">
<td class="col1">
  false
</td><td class="col2">
  false
</td></tr>
```

which is indeed wrong.  Thanks for the fix

PR #1635, opened 2022-10-03
Update link paths in docs markup

erinmorrissey:
Part of https://github.com/Shopify/shopify-dev/issues/25372

This PR updates links used in the Liquid docs to reflect a new routing schema being implemented for SEO reasons. While a tag/filter/object previously lived at `/api/liquid/tags#tagname`, it will now live at `/api/liquid/tags/tagname`, so all of the hard-coded links used in the docs content need to be updated to the new schema.


shainaraskas (approved):

Czx00000 (commented):
- [ ]

PR #1636, opened 2022-10-05
Clarifying description for {% comment %}

Jan0707:
This PR addresses feedback we have received about why liquid would throw parsing errors even for code in {% comment %} blocks. Hopefully, this clarifies that this code (while not being executed/rendered) is indeed parsed initially.


shainaraskas (approved):
Thanks for making this change!

Issue #1621, opened 2022-09-13
Separate core language tests for re-use in other liquid implementations


dylanahsmith:
Ideally, we would have tests for the liquid language that could be shared for re-use across liquid implementations.  This would better support re-implementations of liquid for performance reasons (e.g. liquid-c) or for making it usable from a different language.

These tests could be written in YAML without language specific extensions.  The tests could then be run in ruby using meta-programming to define unit test methods for each YAML test.  A ruby gem (from within this repo) could be provided for reimplementations of liquid with ruby bindings, which would just require an adapter for the implementations liquid API.  Porting liquid for use from another language could just re-use the YAML and require some additional code to run/generate tests for each YAML test.  Either way, this would allow for greater compatibility between implementations and simplify experimentation.

---

As a transitionary step, tests that can be decoupled from this library's API should use test helpers like `assert_template_result` & `assert_match_syntax_error` that abstract from the library API, as https://github.com/Shopify/liquid/pull/1611 started to do.

Tests for liquid extensions (e.g. custom drops, tags or filters) would depend on the library API and should probably be distinguished from language tests so they don't get converted to YAML tests.  Perhaps they shouldn't use those decoupled test helpers like `assert_template_result` or `assert_match_syntax_error` to make it easier to separate those tests when we are ready to rewrite them in YAML.

Each YAML test would provide the liquid source, expected output and parameters for a single render.  As such, we would need to separate ruby test methods that perform multiple renders into their own named test.

PR #1641, opened 2022-10-21
Use static environment assert_template_result input values 

dylanahsmith:
As part of https://github.com/Shopify/liquid/issues/1621, we want tests for the liquid language itself, things that aren't liquid implementation specific.  However, in practice we use the static environment for input even though `assert_template_result` is using the non-static environment that is shared with the counters.  As such, this is neither testing behaviour which we see in practice, nor behaviour that we should need to preserve in a reimplementation of liquid, so I've changed `assert_template_result` to pass input values in the static environment.

Similarly, I've updated the increment/decrement tests to not depend on the ability to pass in initial values for these counters.  I've also updated them to actually test the reading of the variables, which is observable behaviour.

### DIFF
--- a/test/integration/tags/increment_tag_test.rb
+++ b/test/integration/tags/increment_tag_test.rb
@@ -6,20 +6,21 @@ class IncrementTagTest < Minitest::Test
   include Liquid
 
   def test_inc
-    assert_template_result('0', '{%increment port %}', {})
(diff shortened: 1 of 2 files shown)


dylanahsmith:
@ianks & @samdoiron ping for review

samdoiron (approved):

PR #1651, opened 2022-11-01
Special case Hash#last in variable lookup

samdoiron:
Ruby's defines Hash#first via Enumerable but not Hash#last. This is a confusing behaviour, and it bubbles up to Liquid.

Solution: add a special case for command-lookups of Hash#last to return hash.values.last.

### DIFF
--- a/test/integration/variable_test.rb
+++ b/test/integration/variable_test.rb
@@ -130,4 +130,8 @@ class VariableTest < Minitest::Test
   def test_raw_value_variable
     assert_template_result('bar', '{{ [key] }}', { 'key' => 'foo', 'foo' => 'bar' })
   end
+
+  def test_hash_last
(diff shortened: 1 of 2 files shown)


test/integration/variable_test.rb
```diff
@@ -130,4 +130,8 @@ def test_dynamic_find_var
   def test_raw_value_variable
     assert_template_result('bar', '{{ [key] }}', { 'key' => 'foo', 'foo' => 'bar' })
   end
+
+  def test_hash_last
+    assert_template_result('last', '{{ hash.last }}', { 'hash' => { 'first' => 'first', 'last' => 'last' } })
```
samdoiron, line 135: Whoops, this test actually doesn't exercise the new path because the key is also `last`. Fixing now.

PR #1662, opened 2022-12-25
Bring back Registers#values_at

isimluk:
Previously, registers returned from liquid have been plain hashes and thus users of the library were able to call `#value_at` on the registers.

This ability has been removed by the introduction of Liquid::Registers class.


Czx00000 (approved):

ashmaroli:
Not sure how much this is practical, but you could just delegate missing methods to `:static`:
```ruby
def method_missing(method, *args, &block)
  if static.respond_to?(:method)
    static.send(method, *args, &block)
  else
    super
  end
end
```

PR #1665, opened 2023-01-06
[Liquid reference docs] Adding more content to divided_by filter

liamgriffin:
Addressing feedback from https://github.com/Shopify/shopify-dev/issues/25953

This PR is adding more content to the `divided_by` filter description [(on this page of shopify.dev)](https://shopify.dev/api/liquid/filters/divided_by) by detailing the behaviour of how different results are generated by different types of divisors.

Before: 

After:

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -757,7 +757,7 @@ module Liquid
     # @liquid_type filter
     # @liquid_category math
     # @liquid_summary
-    #   Divides a number by a given number. The `divided_by` filter produces a result of the same type as the divisor. This means if you divide by an integer, the result will be an integer, and if you divide by a float, the result will be a float. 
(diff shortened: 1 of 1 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -757,7 +757,7 @@ def times(input, operand)
     # @liquid_type filter
     # @liquid_category math
     # @liquid_summary
-    #   Divides a number by a given number.
+    #   Divides a number by a given number. The `divided_by` filter produces a result of the same type as the divisor. This means if you divide by an integer, the result will be an integer, and if you divide by a float, the result will be a float.
```
shainaraskas, line 760: TIL!

shainaraskas (approved):
Looks lovely! 

I'd consider actually breaking this into an example so you can demonstrate the different impacts on rounding, like the example in the [open source docs](https://shopify.github.io/liquid/filters/divided_by/#controlling-rounding). Alternatively, you can edit the current example so you can demo the rounding behavior, e.g.:

image

PR #1667, opened 2023-01-10
Instrument usage of bug with iteration of String with offset or 0 limit

dylanahsmith:
## Problem

Looks like https://github.com/Shopify/liquid/commit/01dea9467122bf84b0ad9d8249b85aa36d050fef introduced a bugs from trying to preserve ruby 1.8.7 behaviour.  One that I think seems worth trying to fix is the special case of the limit and offset not applying to a String, which seems like enough of an edge case that it may not be used in practice.

The other regression from that commit demonstrates how little usage there was of String iteration at the time, since in ruby 1.8.7 liquid/ruby would iterate over lines of a String (i.e. String#each would split on the default separator `$/`).

## Solution

Use Liquid::Usage.instrument to see if any code iterates of a String with a positive offset or 0 limit.  That way we can decide whether we can fix this buggy behaviour or need to preserve it for backwards compatibility.

### DIFF
--- a/lib/liquid/utils.rb
+++ b/lib/liquid/utils.rb
@@ -16,7 +16,11 @@ module Liquid
 
       # Maintains Ruby 1.8.7 String#each behaviour on 1.9
       if collection.is_a?(String)
-        return collection.empty? ? [] : [collection]
+        return [] if collection.empty?
+        if from > 0 || to == 0
+          Usage.increment("string_slice_bug")
+        end
+        return [collection]
       end
       return [] unless collection.respond_to?(:each)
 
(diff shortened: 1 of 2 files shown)


mattdrose (approved):

dylanahsmith:
The instrumentation shows there are a couple hundred uses of per-hour, so it looks like existing code already depends on this bug.

Unfortunately, it is hard to dig into the usage (with Liquid::Usage.instrument) to get a concrete example and see if the scope of the bug can be reduced (E.g. Is it the `from` / `to` parameter that needs to trigger the bug? Is the bug needed for both the `for` and `tablerow` tags? How many themes actually depend on this bug?).  Perhaps we should have a different method Liquid::Usage method that takes more arguments to we can more easily extract the concrete usage.

PR #1668, opened 2023-01-10
Backport ruby 3.2 support for liquid 4

### LINKED ISSUE #1664: Backporting Ruby 3.2 support to 4.x series
It seems that Jekyll still has Liquid locked to versions 4.x, and it doesn't seem like this will change soon. It would be great if Ruby 3.2 support was backported to Liquid 4.x, as that should hopefully allow current Jekyll to run on Ruby 3.2 🙏🏻

dylanahsmith:
@janko, @parkr & @markets please review and confirm this fixes #1664

I took a similar approach to ruby 2.7 by preserving `Liquid::Template#taint_mode` and `Liquid::Template#taint_mode=` but just not doing any taint checking, to avoid an API breaking change as liquid 5 does.

The rest are just development changes:
* Backport migration of CI from travis to github actions
* Cherry-pick dd70fcfec8abf9cc29be78ca1ce51c8315a6117d to fix ParseTreeVisitorTest for ruby 3 compatibility
* Bump rake in the Gemfile for ruby 3.2 compatibility
* Use the liquid-c 4-0-stable branch, since liquid-c 4.1.0 isn't compatible with liquid v4, but a fix backported from 4.1 to avoid warnings during extension compilation being turned into errors (which affected CI and probably gcc in general)


markets:
Yay! Full matrix is green ✅ again pointing to this branch: https://github.com/markets/jekyll-timeago/pull/36/commits/936bb99c08a699ea29b3b992c5f3e07e79a004cd

Thanks @dylanahsmith!

PR #1671, opened 2023-01-11
Use cache built into `ruby/setup-ruby` action

ashmaroli:
The `ruby/setup-ruby` GitHub Action has a provision for caching installed bundle. Using that cache is better than handling it locally.

### DIFF
--- a/.github/workflows/liquid.yml
+++ b/.github/workflows/liquid.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         entry:
           - { ruby: 2.7, allowed-failure: false } # minimum supported
-          - { ruby: 3.1, allowed-failure: false } # latest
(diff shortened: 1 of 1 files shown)


.github/workflows/liquid.yml
```diff
@@ -1,39 +1,36 @@
 name: Liquid
 on: [push, pull_request]
+
+env:
+  BUNDLE_PATH: "vendor/bundle"
```
dylanahsmith, line 5: Can this environment variable be removed? Since https://github.com/ruby/setup-ruby#bundle-config says

> To perform caching, this action will use bundle config --local path $PWD/vendor/bundle.
Therefore, the Bundler path should not be changed in your workflow for the cache to work (no bundle config path).

PR #1673, opened 2023-01-12
Initialize context to nil on the Drop class

jemmaissroff:
We believe this will reduce megamorphic exits in YJIT. There are currently 241 separate shapes generated with edge_name "@context" in SFR. Initializing it to nil will significantly reduce this number.

### DIFF
--- a/lib/liquid/drop.rb
+++ b/lib/liquid/drop.rb
@@ -25,6 +25,10 @@ module Liquid
   class Drop
     attr_writer :context
 
+    def initialize
+      @context = nil
+    end
+
     # Catch all for the method
     def liquid_method_missing(method)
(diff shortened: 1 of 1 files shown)


casperisfine:
cc @Shopify/liquid @Shopify/liquid-perf this is a super small change to play better with Ruby 3.2 and YJIT. I'm taking the liberty to merge this.

isaacbowen:
> 241 separate shapes generated with edge_name "@context" in SFR

@jemmaissroff What's SFR?

isaacbowen:
:D On the flipside, it's a ton of fun getting the occasional glimpse into the dynamics y'all are working with. Thanks for being willing to share. :)

PR #1675, opened 2023-01-16
Avoid a couple of duplicate assertions

dylanahsmith:
I noticed there a couple of duplicate assertions, so changed them to provide what seemed like it might have been the original intended additional test coverage.


Maaarcocr (approved):

dylanahsmith:
I just forgot to merge this.  Thanks for the reminder.

PR #1677, opened 2023-01-17
Revert "Instrument usage of bug with iteration of String with offset or 0 limit"

dylanahsmith:
Reverts Shopify/liquid#1667

> The instrumentation shows there are a couple hundred uses of per-hour, so it looks like existing code already depends on this bug.

May as well revert this.  We can always revisit this if we want to improve usage instrumentation to dig into the usage of bugs like this more.


dylanahsmith:
@mattdrose ping for review

ianks (approved):

PR #1686, opened 2023-02-13
Fix CI (as it installs RuboCop 1.45.1 and fails)

karreiro:
The CI no longer uses RuboCop 1.44.1 (instead it uses 1.45.1). So, it's failing with this error now:

```
Inspecting 136 files
.....................................................................W..................................................................

Offenses:

performance/shopify/database.rb:54:3: W: Lint/Debugger: Remove debugger entry point p(Database.tables['collections']['frontpage'].keys).
  p(Database.tables['collections']['frontpage'].keys)
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

136 files inspected, 1 offense detected
```

Also, I can open a following PR to remove `rubocop` from the `Gemfile` and update `rubocop-shopify` to `2.12.0' (still, it will require broader changes).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,6 +18,7 @@ group :benchmark, :test do
 end
 
 group :test do
+  gem 'rubocop', '~> 1.44.0'
   gem 'rubocop-shopify', '~> 2.7.0', require: false
   gem 'rubocop-performance', require: false
 
--- a/performance/shopify/database.rb
+++ b/performance/shopify/database.rb
@@ -49,8 +49,3 @@ module Database
     end
   end
 end
-
-if __FILE__ == $PROGRAM_NAME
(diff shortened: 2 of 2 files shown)


dylanahsmith (approved):

dylanahsmith:
It makes sense to be more specific with the rubocop version to avoid having CI silently break by inheriting these non-backwards compatible changes.  However, I think we could also remove that offending code, since it does seem to be for debugging.

karreiro:
Thank you for the review, @dylanahsmith! I've removed the debugging snippet on this PR and merged it.

Also, following this work, I've opened https://github.com/Shopify/liquid/pull/1687 updating `rubocop-shopify`.

PR #1692, opened 2023-02-27
render error message with actual template path

ggmichaelgo:
### Problem

When Liquid is used with a `FileSystem` that allows `render` and `include` tags to render a file just by its filename, the error message does not show its actual file path.

setup:
```
# index.liquid
{% render 'foo' with errors %}

# /some/path/snippets/foo.liquid
{{ foo.standard_error }}
```
output:
```
Liquid error (foo line 1): standard error
```

This becomes an issue when Liquid developers have duplicate filenames in multiple folders, which creates confusion where the error is actually coming from.

### Solution

In `render` and `include` tag, it will set the `context.template` with `template.name`.
The Liquid users can use `TemplateFactory` to create a `Template` with a name:
```ruby
class TemplateFactory
  def for(template_name)
    template = Liquid::Template.new
    template.name = "snippets/" + template_name
    template
  end
end

context = Liquid::Context.build(
  registers: { template_factory: StubTemplateFactory.new }
)

template = Liquid::Template.parse("....", line_numbers: true)
template.render(context)

# renders error message in a format of `Liquid error ({folder}/{filename} line #): error message`
Liquid error (snippets/foo line 1): standard error
```

### DIFF
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -72,9 +72,9 @@ module Liquid
       old_partial       = context.partial
 
       begin
-        context.template_name = partial.name if partial.name
-
+        context.template_name = partial.name
         context.partial = true
+
         context.stack do
           @attributes.each do |key, value|
             context[key] = context.evaluate(value)
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -70,9 +70,15 @@ module Liquid
 
       old_template_name = context.template_name
       old_partial       = context.partial
+
       begin
-        context.template_name = template_name
-        context.partial       = true
+        context.template_name = if Template.file_system.respond_to?(:actual_template_name)
+          Template.file_system.actual_template_name(template_name).delete_suffix('.liquid')
+        else
+          template_name
+        end
+
+        context.partial = true
         context.stack do
           @attributes.each do |key, value|
             context[key] = context.evaluate(value)
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -72,11 +72,7 @@ module Liquid
       old_partial       = context.partial
 
       begin
-        context.template_name = if Template.file_system.respond_to?(:actual_template_name)
-          Template.file_system.actual_template_name(template_name).delete_suffix('.liquid')
-        else
-          template_name
-        end
+        context.template_name = partial.name if partial.name
 
         context.partial = true
         context.stack do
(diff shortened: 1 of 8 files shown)


dylanahsmith:
> When Liquid is used with a `FileSystem` that allows `render` and `include` tags to render a file just by its filename, the error message does not show its actual file path.

This is by design, since we wouldn't want to leak the full file path.  Liquid is designed to allow users to design the page without given those users full system access, meaning that we do want to limit the path to the subpath that the user has access to.

> This becomes an issue when Liquid developers have duplicate filenames in multiple folders, which creates confusion where the error is actually coming from.

Liquid doesn't have the concept of multiple folders, but in practice the partials can end up in a different directory than the template.  Once a layout template is introduced, which may be in another folder, then this can easily introduce confusion.  As such, it makes sense to have the template name to include the subpath from the root of the directory that the user stores liquid code in.  E.g. if they stored their code in a github repo, it could give the path from the root of that repo, as opposed to the subpath for the partials/snippets.

lib/liquid/tags/include.rb
```diff
@@ -70,9 +70,15 @@ def render_to_output_buffer(context, output)
 
       old_template_name = context.template_name
       old_partial       = context.partial
+
       begin
-        context.template_name = template_name
-        context.partial       = true
+        context.template_name = if Template.file_system.respond_to?(:actual_template_name)
+          Template.file_system.actual_template_name(template_name).delete_suffix('.liquid')
```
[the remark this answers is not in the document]
dylanahsmith: The other reason having a Template#name property would be useful is that it could be used as part of given a name to the initial template.  I think we could provide that as a parse option, then Template#render can set `context.template_name = template.name if template.name`.  I think that would make it easy to set the top-level template name so that it can be included in error messages.

lib/liquid/tags/include.rb
```diff
@@ -70,9 +70,11 @@ def render_to_output_buffer(context, output)
 
       old_template_name = context.template_name
       old_partial       = context.partial
+
       begin
-        context.template_name = template_name
-        context.partial       = true
+        context.template_name = partial.name if partial.name
```
dylanahsmith, line 75: I think this can be set unconditionally.  The reason why I was saying to make it conditional in Template#render (if we support a template name parse argument) is to preserve the behaviour of any application that sets Context#template_name on the passed in Context.

```suggestion
        context.template_name = partial.name
```

lib/liquid/tags/render.rb
```diff
@@ -76,7 +76,9 @@ def render_tag(context, output)
 
       render_partial_func = ->(var, forloop) {
         inner_context               = context.new_isolated_subcontext
-        inner_context.template_name = template_name
+
+        inner_context.template_name = partial.name if partial.name
```
dylanahsmith, line 80:
```suggestion
        inner_context.template_name = partial.name
```

dylanahsmith (approved):

PR #1717, opened 2023-05-19
[trim] Fix trimming at end of raw block when specified in the opening tag

gmalette:
The `raw` tag doesn't use `BlockBody` to parse its body. As a result, it currently doesn't respect nor reset the parsing context's `trim_whitespace` value.

This PR makes it reset it at the end of the block, respecting either non-trimming (`endraw %}`) or trimming (`endraw -%}`).

This change is in line with Liquid::C's implementation.

It __does not__ respect inner whitespace control, and `{% raw -%}   inner   {%- endraw %}` will preserve the spaces. This is also in line with Liquid::C.

### DIFF
--- a/lib/liquid/tags/raw.rb
+++ b/lib/liquid/tags/raw.rb
@@ -26,6 +26,7 @@ module Liquid
       @body = +''
       while (token = tokens.shift)
         if token =~ FullTokenPossiblyInvalid && block_delimiter == Regexp.last_match(2)
+          parse_context.trim_whitespace = (token[-3] == WhitespaceControl)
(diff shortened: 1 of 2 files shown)


lib/liquid/tags/raw.rb
```diff
@@ -26,6 +26,7 @@ def parse(tokens)
       @body = +''
       while (token = tokens.shift)
         if token =~ FullTokenPossiblyInvalid && block_delimiter == Regexp.last_match(2)
+          parse_context.trim_whitespace = (token[-3] == WhitespaceControl)
```
samdoiron, line 29: Do we always reset this in the next tag?

samdoiron (approved):

PR #1722, opened 2023-06-15
Add `sum` filter to standard filters

adamklingbaum:
A new filter to sum the values of an array. Either an ordinary sum for an array of numeric values, or as the sum of property values for an array of objects.

```{{ array | sum }}```

or

```{{ array | sum: {property} }}```

### Example usage

```liquid 
{{ card_balances | sum }} 
```

or

```liquid
{{ articles | sum: 'comments_count' }} 
```

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -869,6 +869,31 @@ module Liquid
       false_check || (input.respond_to?(:empty?) && input.empty?) ? default_value : input
     end
 
+    # @liquid_public_docs
+    # @liquid_type filter
+    # @liquid_category array
+    # @liquid_summary
+    #   Returns the sum of all elements in an array.
+    # @liquid_syntax array | sum
+    # @liquid_return [number]
+    def sum(input, property = nil)
+      ary = InputIterator.new(input, context)
+
+      if ary.empty?
+        0
+      elsif property.nil?
+        ary.sum do |item|
+          Utils.to_number(item)
+        end
+      else
+        ary.sum do |item|
+          item.respond_to?(:[]) ? Utils.to_number(item[property]) : 0
+        rescue TypeError
+          raise_property_error(property)
+        end
+      end
+    end
+
     private
 
     attr_reader :context
--- a/test/integration/standard_filter_test.rb
+++ b/test/integration/standard_filter_test.rb
@@ -928,6 +928,60 @@ class StandardFiltersTest < Minitest::Test
     assert_equal([{ "foo" => true }, { "foo" => "for sure" }], @filters.where(input, "foo"))
   end
 
+  def test_sum_with_all_numbers
+    input = [1, 2]
+
+    assert_equal(3, @filters.sum(input))
+    assert_raises(Liquid::ArgumentError, "cannot select the property 'quantity'") do
+      @filters.sum(input, "quantity")
+    end
+  end
+
+  def test_sum_with_numeric_strings
+    input = [1, 2, "3", "4"]
+
+    assert_equal(10, @filters.sum(input))
+    assert_raises(Liquid::ArgumentError, "cannot select the property 'quantity'") do
+      @filters.sum(input, "quantity")
+    end
+  end
+
+  def test_sum_with_nested_arrays
+    input = [1, [2, [3, 4]]]
+
+    assert_equal(10, @filters.sum(input))
+    assert_raises(Liquid::ArgumentError, "cannot select the property 'quantity'") do
+      @filters.sum(input, "quantity")
+    end
+  end
+
+  def test_sum_with_indexable_map_values
+    input = [{ "quantity" => 1 }, { "quantity" => 2, "weight" => 3 }, { "weight" => 4 }]
+
+    assert_equal(0, @filters.sum(input))
+    assert_equal(3, @filters.sum(input, "quantity"))
+    assert_equal(7, @filters.sum(input, "weight"))
+    assert_equal(0, @filters.sum(input, "subtotal"))
+  end
+
+  def test_sum_with_indexable_non_map_values
+    input = [1, [2], "foo", { "quantity" => 3 }]
+
+    assert_equal(3, @filters.sum(input))
+    assert_raises(Liquid::ArgumentError, "cannot select the property 'quantity'") do
+      @filters.sum(input, "quantity")
+    end
+  end
+
+  def test_sum_with_unindexable_values
+    input = [1, true, nil, { "quantity" => 2 }]
+
+    assert_equal(1, @filters.sum(input))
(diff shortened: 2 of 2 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -869,6 +869,31 @@ def default(input, default_value = '', options = {})
...
+    # @liquid_syntax array | sum
+    # @liquid_return [number]
+    def sum(input, property = nil)
+      ary = InputIterator.new(input, context)
+
+      return 0 if ary.empty?
+
+      if property.nil?
+        # If we're doing an ordinary sum, all values must be numeric
+        raise(Liquid::ArgumentError, "cannot sum non-numeric values") unless ary.all? { |el| el.is_a?(Numeric) }
```
samdoiron, line 886: This should probably use `Utils.to_number` for consistency with the `plus` filter.

adamklingbaum: Is your suggestion `...unless ary.all? { |el| Utils.to_number(el).is_a?(Numeric) }`

In other words, not just checking if the element is a Number, but if it can be coerced into to a number? (e.g., numeric strings like "123")

adamklingbaum: Looking at `to_number`, it seems that for anything that is not a `Float`, `Numeric`, or `String` and does not respond to `:to_number`, it returns `0`.

Would we want `{{ [1, 2, 3, "a"] | sum }}` to return `6` or raise a `Liquid::ArgumentError`?

shainaraskas (approved):

test/integration/filter_test.rb
```diff
@@ -165,6 +165,36 @@ def test_liquid_argument_error
     assert_match(/\ALiquid error: wrong number of arguments /, exc.message)
     assert_equal(exc.message, Template.parse(source).render)
   end
+
+  def test_sum
+    src_w_prop = "{{ ary | sum: 'qty' }}"
+    src_wo_prop = "{{ ary | sum }}"
```
charlespwd, line 171: Bunch of nits (non-major):

- I think this would be better with multiple tests

  ```rb
  def test_sum; end
  def test_sum_with_props; end
  def test_sum_non_numeric_values_error; end
  def test_sum_missing_props; end
  ```

  That way, we can inline the source string instead of naming it and it would make the tests more dumb and easier to read. e.g.

  ```rb
  assert_template_result("1000", "{{ numbers | sum }}", { "numbers" => [400, 600] })
  ```


- For consistensy with other tests, the variables holding the strings should be called `source` I think.

- `ary` -> `numbers`, `ary`->`items`
- `qty` -> `quantity`

lib/liquid/standardfilters.rb
```diff
@@ -869,6 +869,31 @@ def default(input, default_value = '', options = {})
       false_check || (input.respond_to?(:empty?) && input.empty?) ? default_value : input
     end
 
+    # @liquid_public_docs
+    # @liquid_type filter
+    # @liquid_category array
+    # @liquid_summary
+    #   Returns the sum of all elements in an array.
+    # @liquid_syntax array | sum
+    # @liquid_return [number]
```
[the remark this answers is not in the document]
charlespwd: Hmm looks like it's coming from somewhere else. There's a diff between what's shown on shopify.dev and what's here. @shainaraskas, what's up with that?

test/integration/filter_test.rb
adamklingbaum: Thanks @charlespwd. I was building the tests in the wrong class. Now that I'm looking in the right place, I see the pattern that you are referencing and I'll revise accordingly.

lib/liquid/standardfilters.rb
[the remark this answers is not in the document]
shainaraskas: Filter and tag parameter docs [are stored in shopify/storefront-renderer](https://vault.shopify.io/page/Liquid-docs-source-files~1vWb.md#liquid). This is a limitation of our example model, and the fact that parameters are only documented as examples.

e.g. [this sort by property example](https://shopify.dev/docs/api/liquid/filters/sort#sort-sort-by-an-array-item-property) is stored [here](https://github.com/Shopify/storefront-renderer/blob/master/app/liquid/examples/gem/filters/sort/parameters/sort-property.liquid)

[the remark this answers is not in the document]
shainaraskas: you can use `@liquid_optional_param` to document parameters for intellisense, though.

test/integration/standard_filter_test.rb
```diff
@@ -928,6 +928,47 @@ def test_where_no_target_value
...
+    end
+  end
+
+  def test_sum_with_property_on_unindexable_values
+    input = [true, nil, :foo]
+    assert_equal(0, @filters.sum(input, "quantity"))
+
+    mixed_input = [true, nil, :foo, { "quantity" => 3 }]
+    assert_equal(3, @filters.sum(mixed_input, "quantity"))
+  end
```
samdoiron, line 970: Implementation looks good to me! Before we merge, could we add some tests for the `#to_number` case as well as the `TypeError` case? For the later, an array of arrays should work, eg

```ruby 
assert_equal(0, @filters.sum([[1]], "quantity")) # => 0
```

adamklingbaum: `test_sum_with_property_on_indexable_non_map_values` was intended to test the `TypeError` case
https://github.com/Shopify/liquid/blob/81fdaf81580b81e86a93256d983358185fd9f0c0/test/integration/standard_filter_test.rb#L956-L962

`test_sum_with_non_numeric_values` was intended to test the `to_number` case. 
https://github.com/Shopify/liquid/blob/81fdaf81580b81e86a93256d983358185fd9f0c0/test/integration/standard_filter_test.rb#L941-L944

But I can make these more complete – I can think of some additional edge cases to test.

adamklingbaum: Pushed up an updated test suite that I believe covers the relevant edge cases.

test/integration/standard_filter_test.rb
```diff
@@ -928,6 +928,60 @@ def test_where_no_target_value
...
+    assert_raises(Liquid::ArgumentError, "cannot select the property 'quantity'") do
+      @filters.sum(input, "quantity")
+    end
+  end
+
+  def test_sum_with_nested_arrays
+    input = [1, [2, [3, 4]]]
+
+    assert_equal(10, @filters.sum(input))
+    assert_raises(Liquid::ArgumentError, "cannot select the property 'quantity'") do
```
charlespwd, line 953: It me or this error message kind of doesn't make sense with the test title?

charlespwd (approved):
LGTM! :)

PR #1749, opened 2023-10-16
Add deep search for filter taking in properties

andershagbard:
This PR allows for backward compatibility deep search :woah:

Like so:
```liquid
{{ input | where: 'some.nested.property', 'foo', deep: true }}
{{ input | where: 'some_nested_property', 'foo', deep: '_' }} # Custom separator
```

Added deep search for the following filters. This should be all Liquid filters which takes in a property argument:
- [x] map
- [x] sort
- [x] sort_natural
- [x] sum
- [x] uniq
- [x] where
- [x] compact


andershagbard:
@dylanahsmith I am not sure if you still work on Liquid, but I would love for someone in the Liquid team to take a look at this PR.

andershagbard:
@karreiro @benjaminsehl ping

andershagbard:
It's to safeguard against the JSON field type, which can contain dots in it's properties. It's an edge case I guess.

```json
{
  "foo.bar": true
}
```

Ref: https://github.com/Shopify/liquid/pull/1436#issuecomment-916936710

andershagbard:
Closed by https://github.com/Shopify/liquid/pull/1869

PR #1750, opened 2023-10-20
Fix for custom tag rendering

IevaGraz:
After https://github.com/Shopify/liquid/pull/1091 the behaviour when a render method of custom tags returns nil has changed. Before something like 
``` 
 Liquid::Template.parse('foo {% custom_url%20SOME-CUSTOM-PARAM %}').render()
```
that depends on

```
 class CustomUrl < Liquid::Tag
     ...
    def render(_context)
      return unless something

      #build url
    end
```

Would return just `foo `.
After the change it returns `foo Liquid error: internal` 
Because [this line here ](https://github.com/Shopify/liquid/blob/e3dcc75ab533eb177cabf3df16ed27c1a21fc33c/lib/liquid/tag.rb#L57) throws `TypeError:    no implicit conversion of nil into String`

### DIFF
--- a/lib/liquid/tag.rb
+++ b/lib/liquid/tag.rb
@@ -54,7 +54,8 @@ module Liquid
     # of the `render_to_output_buffer` method will become the default and the `render`
     # method will be removed.
     def render_to_output_buffer(context, output)
-      output << render(context)
+      render_result = render(context)
+      output << render_result if render_result
       output
     end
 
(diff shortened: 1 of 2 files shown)


ianks (approved):

ianks:
Just checked this with our app's CI and it looks good. Thanks for the fix!

PR #1763, opened 2023-11-30
fix parsing comment tag delimiter

ggmichaelgo:
https://github.com/Shopify/liquid/pull/1755 introduced a bug to `comment` tag parsing.

The `comment` tag uses this regex to match the tag delimiter:
```regex
 /\A(.*)(?-mix:\{\%)-?\s*(endcomment)\s*(.*)?-?(?-mix:\%\})\z/m
```

After matching the `endcomment`, the regex is also matching the extra string with `\s*(.*)?`
Since it doesn't require a whitespace after matching with the tag delimiter, it allows delimiters like `{% endcommentABC %}` to be matched.

### DIFF
--- a/lib/liquid/tags/comment.rb
+++ b/lib/liquid/tags/comment.rb
@@ -15,7 +15,7 @@ module Liquid
   #   {% endcomment %}
   # @liquid_syntax_keyword content The content of the comment.
   class Comment < Block
-    TAG_DELIMITER = /\A(.*)#{TagStart}#{WhitespaceControl}?\s*(endcomment)\s*(.*)?#{WhitespaceControl}?#{TagEnd}\z/om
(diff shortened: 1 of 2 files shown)


lib/liquid/tags/comment.rb
```diff
@@ -15,7 +15,7 @@ module Liquid
   #   {% endcomment %}
   # @liquid_syntax_keyword content The content of the comment.
   class Comment < Block
-    TAG_DELIMITER = /\A(.*)#{TagStart}#{WhitespaceControl}?\s*(endcomment)\s*(.*)?#{WhitespaceControl}?#{TagEnd}\z/om
+    TAG_DELIMITER = /\A(.*)#{TagStart}#{WhitespaceControl}?\s*(endcomment)\s*(\s.*)?#{WhitespaceControl}?#{TagEnd}\z/om
```
peterzhu2118, line 18: I think this achieves the same result.

```suggestion
    TAG_DELIMITER = /\A(.*)#{TagStart}#{WhitespaceControl}?\s*(endcomment)\s+(.*)?#{WhitespaceControl}?#{TagEnd}\z/om
```

PR #1764, opened 2023-12-01
implement whitespace control to comment tag

ggmichaelgo:
https://github.com/Shopify/liquid/pull/1755 introduced a whitespace control bug to `comment` tag parsing.

The `comment` tag is now updated to update the `parse_context.trim_whitespace` when it recognises a whitespace control character like the [raw tag](https://github.com/Shopify/liquid/blob/cbb422e5d39b52671cacf86ca684713d565b4bed/lib/liquid/tags/raw.rb#L27-L30).

### DIFF
--- a/lib/liquid/tags/comment.rb
+++ b/lib/liquid/tags/comment.rb
@@ -73,7 +73,10 @@ module Liquid
           when "endcomment"
             comment_tag_depth -= 1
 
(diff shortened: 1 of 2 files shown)


peterzhu2118 (approved):
Sorry for the late review. This LGTM!

PR #1769, opened 2023-12-12
fix parsing comment tag with extra string

ggmichaelgo:

https://github.com/Shopify/liquid/pull/1755 introduced a bug in `comment` tag that prevents the tag from having extra string inside.

With the bug, this valid template fails to be parsed
```liquid
{%comment}{% assign a = 1 %}
  {% if true %}
{%endcomment}{% endif %}
```

### DIFF
--- a/lib/liquid/tags/comment.rb
+++ b/lib/liquid/tags/comment.rb
@@ -15,8 +15,6 @@ module Liquid
   #   {% endcomment %}
   # @liquid_syntax_keyword content The content of the comment.
   class Comment < Block
-    TAG_DELIMITER = /\A(.*)#{TagStart}#{WhitespaceControl}?\s*(endcomment)\s*(\s.*)?#{WhitespaceControl}?#{TagEnd}\z/om
-
     def render_to_output_buffer(_context, output)
       output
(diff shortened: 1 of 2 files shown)


peterzhu2118:
I'm confused about the example template. Is `{%comment}` valid opening  for a comment? Is `{%endcomment}` a valid closing? What's the `{% endif %}` supposed to close?

ggmichaelgo:
> I'm confused about the example template. Is `{%comment}` valid opening for a comment? Is `{%endcomment}` a valid closing? What's the `{% endif %}` supposed to close?

That is correct.
With Liquid 5.4.0, you can see how the `comment` tag used to get parsed.
```ruby
require 'liquid'

template = Liquid::Template.parse("{%comment}{% assign a = 1 %}{%endcomment}{% endif %}")
pp template.root.nodelist[0]
```

peterzhu2118 (approved):
I think this PR makes sense. However, I'm not a fan that this is valid Liquid syntax because it's confusing and hard to understand. Do we have a plan to deprecate or drop support for weird syntax like this?

PR #1770, opened 2023-12-14
don't allow invalid syntax inside comment tag

ggmichaelgo:
https://github.com/Shopify/liquid/pull/1755 introduced a new feature that allows invalid code to be inside a `comment` tag.

For Liquid version `5.4.0`, this is a valid template, and not with our current `main` branch:
```liquid
{% comment %}
  {% assign a = 1
  {% endcomment %}
{% endcomment %}
```
The `assign` tag is not closed immediately, `{% assign a = 1  {% endcomment %}` is parsed as the `assign` tag.

Since the https://github.com/Shopify/liquid/pull/1755 changed the `comment` tag parser to use the `BlockBody::FullTokenPossiblyInvalid` regex pattern, it parses `{% assign a = 1  {% endcomment %}` to be the `comment` tag delimiter, and it raises a syntax error by the extra `{% endcomment %}` token.

### How are you solving this issue?

In order to remove this breaking change, this PR updates the `comment` to parse tokens like `BlockBody` and not like `raw` tag.
However, the `comment` tag will still allow unclosed tags such as this example:
```liquid
{% comment %}
  {% if true %}
  {% if  %}
{% endcomment %}
```

### DIFF
--- a/test/unit/tags/comment_tag_unit_test.rb
+++ b/test/unit/tags/comment_tag_unit_test.rb
@@ -39,7 +39,7 @@ class CommentTagUnitTest < Minitest::Test
     LIQUID
   end
 
-  def test_block_body_must_be_valid
+  def test_open_tags_in_comment
     assert_template_result('', <<~LIQUID.chomp)
       {% comment %}
         {% assign a = 123 {% comment %}
--- a/test/unit/tags/comment_tag_unit_test.rb
+++ b/test/unit/tags/comment_tag_unit_test.rb
@@ -31,25 +31,45 @@ class CommentTagUnitTest < Minitest::Test
     LIQUID
   end
 
-  def test_allows_incomplete_tags_inside_a_comment
-    assert_template_result("", <<~LIQUID.chomp)
+  def test_allows_unclosed_tags
+    assert_template_result('', <<~LIQUID.chomp)
       {% comment %}
-        {% assign foo = "1"
+        {% if true %}
+      {% endcomment %}
+    LIQUID
+  end
+
+  def test_block_body_must_be_valid
+    assert_template_result('', <<~LIQUID.chomp)
+      {% comment %}
+        {% assign a = 123 {% comment %}
       {% endcomment %}
     LIQUID
 
-    assert_template_result("", <<~LIQUID.chomp)
-      {% comment %}
+    assert_raises(Liquid::SyntaxError) do
+      assert_template_result("", <<~LIQUID.chomp)
         {% comment %}
-          {% invalid
+          {% assign foo = "1"
         {% endcomment %}
-      {% endcomment %}
-    LIQUID
+      LIQUID
+    end
 
-    assert_template_result("", <<~LIQUID.chomp)
-      {% comment %}
-      {% {{ {%- endcomment %}
-    LIQUID
+    assert_raises(Liquid::SyntaxError) do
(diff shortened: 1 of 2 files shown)


test/unit/tags/comment_tag_unit_test.rb
```diff
@@ -31,25 +31,45 @@ def test_does_not_parse_nodes_inside_a_comment
...
+    assert_template_result('', <<~LIQUID.chomp)
       {% comment %}
-        {% assign foo = "1"
+        {% if true %}
       {% endcomment %}
     LIQUID
+  end
 
-    assert_template_result("", <<~LIQUID.chomp)
+  def test_block_body_must_be_valid
```
peterzhu2118, line 42: The new name for this test is a bit confusing. I'm not sure what you mean by "block body must be valid".

test/test_helper.rb
```diff
@@ -3,6 +3,8 @@
 
 ENV["MT_NO_EXPECTATIONS"] = "1"
 require 'minitest/autorun'
+require 'minitest/focus'
+require 'byebug'
```
peterzhu2118, line 7: Did you mean to commit this?

Gemfile
```diff
@@ -21,6 +21,8 @@ group :test do
   gem 'rubocop', '~> 1.44.0'
   gem 'rubocop-shopify', '~> 2.12.0', require: false
   gem 'rubocop-performance', require: false
+  gem "minitest-focus"
+  gem 'byebug'
```
peterzhu2118, line 25: Ditto.

test/test_helper.rb
ggmichaelgo: Oops... my bad...

Gemfile
ggmichaelgo: 🤦

test/unit/tags/comment_tag_unit_test.rb
ggmichaelgo: right... this unit test is testing the comment tag's parsing behaviour with open tags like `{% assign a=1`
I have renamed the unit test to `test_open_tags_in_comment`

test/unit/tags/comment_tag_unit_test.rb
```diff
@@ -31,25 +31,45 @@ def test_does_not_parse_nodes_inside_a_comment
...
+        {% endcomment %}
+      LIQUID
+    end
+
+    assert_raises(Liquid::SyntaxError) do
+      assert_template_result("", <<~LIQUID.chomp)
+        {% comment %}
+        {% {{ {%- endcomment %}
+      LIQUID
+    end
```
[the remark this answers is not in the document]
ggmichaelgo: I agree, however, we can't introduce a change that will break existing Liquid templates.

For an example,
```liquid
{% comment %}
  {% assign a = 123
  {% endcomment %}
{% endcomment %}
```
This Liquid code is a valid Liquid code with version `5.4.0` because the `assign` tag is captured as `{% assign a = 123 {% endcomment %}`. (Liquid allows extra string inside a tag)

So, if we parse `comment` tag like the `raw` tag, we would capture, we would ignore the `{% assign a = 123` and capture `{% endcomment %}` from line 3 as the comment tag's delimiter. Since the template has an extra `{% endcomment %}` tag delimiter, the template is no longer valid.

peterzhu2118 (approved):

PR #1775, opened 2024-01-10
allow non-string template source

ggmichaelgo:

https://github.com/Shopify/liquid/pull/1774 introduced a bug that doesn't allow non-string values such as `nil` or `Integer` as a template source.

Expected behaviour:
```ruby
require 'liquid'

Liquid::Template.parse(nil).render # => ""
Liquid::Template.parse(1).render # => "1"
```

### How are you solving this?

This PR updates the `Template#parse` to only check the template's validity if the template is a `String`

### DIFF
--- a/lib/liquid/template.rb
+++ b/lib/liquid/template.rb
@@ -107,8 +107,9 @@ module Liquid
     # Returns self for easy chaining
     def parse(source, options = {})
       parse_context = configure_options(options)
+      source = source.to_s.to_str
 
-      if source.is_a?(String) && !source.valid_encoding?
+      unless source.valid_encoding?
         raise SyntaxError, parse_context.locale.t("errors.syntax.invalid_template_encoding")
       end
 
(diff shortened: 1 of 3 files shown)


lib/liquid/template.rb
```diff
@@ -108,7 +108,7 @@ def initialize
     def parse(source, options = {})
       parse_context = configure_options(options)
 
-      unless source.valid_encoding?
+      if source.is_a?(String) && !source.valid_encoding?
```
samdoiron, line 111: What if the `to_s` value is a string with an invalid encoding?

ggmichaelgo: Good catch... That is possible... 
```ruby
class ByteString
  def to_s
    "\xC0"
  end
end

Liquid::Template.parse(ByteString.new) # this should raise an error
```

I have updated the PR to convert the `source` to a `String` object from the `Template#parse` function instead of `Tokenize#initialize`.

gmalette:
Wait what? How/when does this happen?

PR #1776, opened 2024-01-12
add new TemplateEncodingError

ggmichaelgo:

Instead of raising the `SyntaxError` when the template has an invalid UTF-8 sequence, raise a specific `TemplateEncodingError`.


zachmccormick:
Does this need to be updated in the liquid-c gem? I am getting failures for https://github.com/Shopify/liquid-c/blob/main/test/unit/variable_test.rb#L289

PR #1781, opened 2024-01-24
update variable's context before invoking its to_liquid

ggmichaelgo:

Liquid-C PR: https://github.com/Shopify/liquid-c/pull/212 (Liquid-C PR needs to be merged first)

On variable lookups, `Context` is invoking the drop's `to_liquid` function, and update the values `context`  to itself.

This becomes problematic if the drop's `to_liquid` has some contextual checks:

```ruby
require 'liquid'

class ProductsDrop < Liquid::Drop
  def initialize(products)
    @products = products
  end

  def size
    @products.size
  end

  def to_liquid
    if @context["forloop"]
      @products.first(@context["forloop"].length)
    else
      @products
    end
  end
end

context = {
  "products" => ProductsDrop.new(["A", "B", "C", "D", "E"]),
}

template = Liquid::Template.parse(<<~LIQUID)
  {% for i in (1..3) %}
    for-loop: {{ products | size }} # returns 3
  {% endfor %}

  {{ products }} # it still returns 3 instead of 5
LIQUID

puts template.render(context)
```

We could solve issue by updating the `Context#find_variable` to update the drop's `context` before invoking its `to_liquid` method.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,6 +23,6 @@ group :test do
   gem 'rubocop-performance', require: false
 
   platform :mri, :truffleruby do
-    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'master'
+    gem 'liquid-c', github: 'Shopify/liquid-c', ref: 'main'
   end
 end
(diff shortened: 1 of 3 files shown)


ggmichaelgo:
ping @Shopify/guardians-of-the-liquid

samdoiron (approved):

PR #1787, opened 2024-02-15
Add support in regular expressions for UTF-8 whitespace detection

zachmccormick:
We ran across a nasty bug at Braze where a customer was supplying the UTF-8 non-breaking space character in a Liquid template they were providing to us, and it took a very long time to debug why it was not parsing correctly. It turns out that the user-supplied Liquid string had some UTF-8 non-breaking spaces in it, which the current regular expressions do not count as whitespace (`\s` only includes ASCII whitespace, while `[[:space:]]` includes ASCII _and_ UTF-8 whitespace characters).

I replaced `\s` everywhere, but I added a single test case that red-greens against the existing code. Getting full coverage of every possibility seemed excessive, although I'm open to implementing more thorough tests if it's needed before merging.

### DIFF
--- a/lib/liquid.rb
+++ b/lib/liquid.rb
@@ -36,8 +36,8 @@ module Liquid
   VariableEnd                 = /\}\}/
   VariableIncompleteEnd       = /\}\}?/
   QuotedString                = /"[^"]*"|'[^']*'/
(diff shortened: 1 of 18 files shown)


zachmccormick:
I have signed the CLA!

Issue #1715, opened 2023-05-17
Release a new gem version


isaacbowen:
5.4.0 was the last version cut, and that [was back in July of 2022](https://github.com/Shopify/liquid/commit/6f8722a).

My specific motivation in filing this is that `{%- endraw -%}` didn't start working until https://github.com/Shopify/liquid/commit/74e505f, in February, and I'd like to be able to use that with an official stable release. :D

drnic:
Now that it's nearing the end of September 2023 perhaps a new v5.5 release? Thanks in advance

isaacbowen:
closed by #1791 ☺️ 🧘

PR #1791, opened 2024-03-21
Bump to v5.5.0

ianks:
Locking in a version to publish here, since it's been a long time.


isaacbowen:
THANK YOU

*closes #1715

PR #1801, opened 2024-05-11
Add an ontional limit argument to split filter

### LINKED ISSUE #1800: Add limit to split filter, like split method in Ruby
In Ruby language, the `split` method accepts an optional limit argument which is handy for the cases in which the user want to split the string into max number of chunks. It would be handy to have this feature in Liquid too.

I would like to implement it if there is no objection.

ahangarha:
This PR adds an optional limit argument to the `split` filter in Liquid to make it work similar to the Ruby `split` method.

Closes #1800

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -284,11 +284,13 @@ module Liquid
     # @liquid_type filter
     # @liquid_category string
     # @liquid_summary
-    #   Splits a string into an array of substrings based on a given separator.
+    #   Splits a string into an array of substrings based on a given separator and an optional limit number
     # @liquid_syntax string | split: string
     # @liquid_return [array[string]]
-    def split(input, pattern)
-      input.to_s.split(pattern.to_s)
+    def split(input, pattern, limit = nil)
+      limit = limit.respond_to?(:to_i) ? limit.to_i : 0
+
+      input.to_s.split(pattern.to_s, limit.to_i)
     end
 
     # @liquid_public_docs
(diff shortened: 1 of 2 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -284,11 +284,13 @@ def truncatewords(input, words = 15, truncate_string = "...")
...
     # @liquid_category string
     # @liquid_summary
-    #   Splits a string into an array of substrings based on a given separator.
+    #   Splits a string into an array of substrings based on a given separator and an optional limit number
     # @liquid_syntax string | split: string
     # @liquid_return [array[string]]
-    def split(input, pattern)
-      input.to_s.split(pattern.to_s)
+    def split(input, pattern, limit = nil)
+      limit = limit.respond_to?(:to_i) ? limit.to_i : 0
```
ianks, line 291: I think we should use `Liquid::Utils.to_integer` here.

ahangarha: Oh! I didn't know. I will update the PR ASAP.

ahangarha: I made the requested change. Also I updated the method docs.

ianks: After thinking about it, I think you're right. `Utils.to_integer` is better here. Let's add a test that it raises a `Liquid::ArgumentError` when a non-int is passed as well.

ahangarha: Great. I made the required changes.

PR #1848, opened 2024-11-04
clean up all warnings by using new Environment

ggmichaelgo:


espen:
Looks like `Liquid::Template.error_mode` is still referenced two places. I guess these should be changed?

https://github.com/Shopify/liquid/blob/main/performance/profile.rb#L6
https://github.com/Shopify/liquid/blob/main/performance/memory_profile.rb#L56

espen:
Can I just change `Liquid::Template.register_filter` to `Liquid::Environment.default` or do I need to do something more complex?

PR #1869, opened 2024-12-12
Add `find`, `find_index`, `has`, and `reject` filters to arrays

karreiro:
This PR implements the [RFC#1864](https://github.com/Shopify/liquid/discussions/1864) and introduces support for the `find`, `find_index`, `has`, and `reject` filters into the Liquid Arrays API.

The new filters support multi-level access to drops, so they can access properties like this:
```liquid
{% assign product = products | find: 'price.unit', 'BRL' %}
```

For the sake of consistency, multi-level access has been introduced for all related filters in the Array API: `map`, `sort`, `sort_natural`, `sum`, `uniq`, `where`, and `compact`.

--

Special thanks to @andershagbard who started this work on https://github.com/Shopify/liquid/pull/1573 and https://github.com/Shopify/liquid/pull/1749. I'm including those commits as part of this PR; while the implementation approach is slightly different, important work happened there, and it's valuable to have that context in the Liquid commit history.

Also, thank you to everyone who contributed with thoughts and suggestions in the RFC.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -367,7 +367,10 @@ module Liquid
     #   Sorts the items in an array in case-sensitive alphabetical, or numerical, order.
     # @liquid_syntax array | sort
     # @liquid_return [array[untyped]]
-    def sort(input, property = nil)
+    # @liquid_optional_param deep [boolean | string] Whether to use dot notation to perform a deep search. A string can be passed to change separator.
+    def sort(input, property = nil, options = {})
+      options = {} unless options.is_a?(Hash)
+      deep = deep_search_properties(property, options)
       ary = InputIterator.new(input, context)
 
       return [] if ary.empty?
@@ -378,7 +381,13 @@ module Liquid
         end
       elsif ary.all? { |el| el.respond_to?(:[]) }
         begin
-          ary.sort { |a, b| nil_safe_compare(a[property], b[property]) }
+          ary.sort do |a, b|
+            if deep[:enable]
+              nil_safe_compare(a.dig(*deep[:properties]), b.dig(*deep[:properties]))
+            else
+              nil_safe_compare(a[property], b[property])
+            end
+          end
         rescue TypeError
           raise_property_error(property)
         end
@@ -396,7 +405,10 @@ module Liquid
     #   > string, so sorting on numerical values can lead to unexpected results.
     # @liquid_syntax array | sort_natural
     # @liquid_return [array[untyped]]
-    def sort_natural(input, property = nil)
+    # @liquid_optional_param deep [boolean | string] Whether to use dot notation to perform a deep search. A string can be passed to change separator.
+    def sort_natural(input, property = nil, options = {})
+      options = {} unless options.is_a?(Hash)
+      deep = deep_search_properties(property, options)
       ary = InputIterator.new(input, context)
 
       return [] if ary.empty?
@@ -407,7 +419,13 @@ module Liquid
         end
       elsif ary.all? { |el| el.respond_to?(:[]) }
         begin
-          ary.sort { |a, b| nil_safe_casecmp(a[property], b[property]) }
+          ary.sort do |a, b|
+            if deep[:enable]
+              nil_safe_casecmp(a.dig(*deep[:properties]), b.dig(*deep[:properties]))
+            else
+              nil_safe_casecmp(a[property], b[property])
+            end
+          end
         rescue TypeError
           raise_property_error(property)
         end
@@ -423,7 +441,10 @@ module Liquid
     #   This requires you to provide both the property name and the associated value.
     # @liquid_syntax array | where: string, string
     # @liquid_return [array[untyped]]
-    def where(input, property, target_value = nil)
+    # @liquid_optional_param deep [boolean | string] Whether to use dot notation to perform a deep search. A string can be passed to change separator.
+    def where(input, property, target_value = nil, options = {})
+      options = {} unless options.is_a?(Hash)
+      deep = deep_search_properties(property, options)
       ary = InputIterator.new(input, context)
 
       if ary.empty?
@@ -439,7 +460,8 @@ module Liquid
         end
       else
         ary.select do |item|
-          item[property] == target_value
+          item_value = deep[:enable] ? item.dig(*deep[:properties]) : item[property]
+          item_value == target_value
         rescue TypeError
           raise_property_error(property)
         rescue NoMethodError
@@ -457,6 +479,7 @@ module Liquid
     #   Removes any duplicate items in an array.
     # @liquid_syntax array | uniq
     # @liquid_return [array[untyped]]
+<<<<<<< HEAD
 =======
     # Reject the elements of an array to those with a certain property value.
     # By default the target is any falsy value.
@@ -485,6 +508,12 @@ module Liquid
     # provide optional property with which to determine uniqueness
 >>>>>>> bfdcfcea (Add reject filter)
     def uniq(input, property = nil)
+=======
+    # @liquid_optional_param deep [boolean | string] Whether to use dot notation to perform a deep search. A string can be passed to change separator.
+    def uniq(input, property = nil, options = {})
+      options = {} unless options.is_a?(Hash)
+      deep = deep_search_properties(property, options)
+>>>>>>> 58d7d1a8 (Add deep search for suitable filters)
       ary = InputIterator.new(input, context)
 
       if property.nil?
@@ -493,7 +522,7 @@ module Liquid
         []
       else
         ary.uniq do |item|
-          item[property]
+          deep[:enable] ? item.dig(*deep[:properties]) : item[property]
         rescue TypeError
           raise_property_error(property)
         rescue NoMethodError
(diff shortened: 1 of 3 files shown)


lib/liquid/standardfilters.rb
```diff
@@ -918,6 +948,50 @@ def sum(input, property = nil)
...
+        end
+      rescue TypeError
+        raise_property_error(property)
+      rescue NoMethodError
+        return nil unless item.respond_to?(:[])
+        raise
+      end
+    end
+
+    def fetch_property(drop, property_or_keys)
```
andershagbard, line 970:
```suggestion
    def get_property(drop, property_or_keys)
```

I would probably name this method `get_`. My understanding of the wording, is that get is internal, and fetch is external.

[Other languages](https://laravel.com/docs/11.x/helpers) also use `get` for this kind of thing.

lib/liquid/standardfilters.rb
```diff
@@ -424,29 +424,59 @@ def sort_natural(input, property = nil)
...
+      filter_array(input, property, target_value) { |ary, &block| ary.find(&block) }
+    end
+
+    # @liquid_public_docs
+    # @liquid_type filter
+    # @liquid_category array
+    # @liquid_summary
+    #   Returns the index of the first item in an array with a specific property value.
+    # @liquid_description
+    #   This requires you to provide both the property name and the associated value.
```
jg-rp, line 475: Without providing an associated value, the following test cases currently pass. Is this intended behaviour? The situation is similar for other filters that use `filter_array()`.

```ruby
def test_find_index_without_target_value
    products = [
      { "title" => "Pro goggles",    "price" => 1299 },
      { "title" => "Thermal gloves", "price" => 1499 },
      { "title" => "Alpine jacket",  "price" => 3999 },
      { "title" => "Mountain boots", "price" => nil },
      { "title" => "Safety helmet",  "price" => 1999 }
    ]

    template = <<~LIQUID
      {%- assign index = products | find_index: 'price' -%}
      {{- index -}}
    LIQUID
    expected_output = "0"

    assert_template_result(expected_output, template, { "products" => products })
  end

  def test_find_index_without_target_value_missing_property
    products = [
      { "title" => "Pro goggles" },
      { "title" => "Thermal gloves", "price" => 1499 },
      { "title" => "Alpine jacket",  "price" => 3999 },
      { "title" => "Mountain boots", "price" => nil },
      { "title" => "Safety helmet",  "price" => 1999 }
    ]

    template = <<~LIQUID
      {%- assign index = products | find_index: 'price' -%}
      {{- index -}}
    LIQUID
    expected_output = "1"

    assert_template_result(expected_output, template, { "products" => products })
  end
```

karreiro: Thank you for bringing this up, @jg-rp!

Yes, this is the intended behavior to maintain consistency with the `where` filter. When no value is provided, the existing `where` filter operates as shown below:

```ruby
def test_find_index_without_target_value1
  products = [
    { "title" => "Pro goggles",    "price" => 1299 },
    { "title" => "Thermal gloves", "price" => 1499 },
    { "title" => "Alpine jacket",  "price" => 3999 },
    { "title" => "Mountain boots", "price" => nil },
    { "title" => "Safety helmet",  "price" => 1999 }
  ]

  template = <<~LIQUID
    {{- products | where: 'price' | map: 'title' | join: ', ' -}}
  LIQUID
  expected_output = "Pro goggles, Thermal gloves, Alpine jacket, Safety helmet"

  assert_template_result(expected_output, template, { "products" => products })
end
```

As we can notice in the example, when no value is specified, the `where` filter selects items where the given attribute is truthy. Therefore, the new filters are designed to follow the same convention for consistency.

I hope this helps clarify the behavior. Please feel free to reopen the discussion if you have any further questions or thoughts.

Maaarcocr (approved):

lib/liquid/standardfilters.rb
```diff
@@ -918,6 +948,50 @@ def sum(input, property = nil)
...
+      # ]
+      # ```
+      value = drop[property_or_keys]
+
+      return value if !value.nil? || !property_or_keys.is_a?(String)
+
+      keys = property_or_keys.split('.')
+      keys.reduce(drop) do |drop, key|
+        drop.respond_to?(:[]) ? drop[key] : drop
+      end
```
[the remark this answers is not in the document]
charlespwd: This is what happens when all you have left at home is decaf 🙈! TY

bakura10:
Awesome news <3. That's a very welcome addition to Liquid which will help us to clean up a lot of our code. Do you know when this will be available to use in Shopify?

lib/liquid/standardfilters.rb
```diff
@@ -918,6 +948,50 @@ def sum(input, property = nil)
...
+      #   { 'name' => 'Item 2', 'price' => { 'value' => 39900 } }
+      # ]
+      # ```
+      value = drop[property_or_keys]
+
+      return value if !value.nil? || !property_or_keys.is_a?(String)
+
+      keys = property_or_keys.split('.')
+      keys.reduce(drop) do |drop, key|
+        drop.respond_to?(:[]) ? drop[key] : drop
```
ianks, line 991: This seems weird to me... does this mean that:

```
product.dafsdfasdfasdf #=> ProductDrop
```

I would have expected it to be like: 

```
product.dafsdfasdfasdf #=> nil
```

lib/liquid/standardfilters.rb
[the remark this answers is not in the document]
jg-rp: I have reservations about parsing strings at render time in general. At the very least it will play havoc with static analysis.

lib/liquid/standardfilters.rb
karreiro: That would be weird indeed. But I believe it's already working as expected (and `product.dafsdfasdfasdf` already returns `nil`), as this test passes:

```ruby
def test_prop_test
  products = [
    { "title" => "Pro goggles" },
    { "title" => "Thermal gloves", "price" => 1499 },
    { "title" => "Alpine jacket",  "price" => 3999 },
    { "title" => "Mountain boots", "price" => nil },
    { "title" => "Safety helmet",  "price" => 1999 },
  ]

  template = <<~LIQUID
    {%- assign example1 = products | find_index: 'dafsdfasdfasdf' -%}
    {{- example1 -}}
    {%- assign example2 = products | find_index: 'dafsdfasdfasdf.dafsdfasdfasdf' -%}
    {{- example2 -}}
  LIQUID
  expected_output = ""

  assert_template_result(expected_output, template, { "products" => products })
end
```

The reason we're achieving the expected behavior is that:
- For `example1`, we're resolving that at line 985
- For `example2`, we're resolving that because in the first iteration `drop` is product and returns nil for the first `dafsdfasdfasdf` call, then in the second call `drop` is already `nil`

So, I believe we're already achieving that behavior, but please let me know if I'm missing something :)

karreiro:
:wave: Hey everyone,

We’ve conducted additional assessments of the backward-compatibility of this change and confirmed this is not 100% backward-compatible as stated [here](https://github.com/Shopify/liquid/pull/1869/files#diff-1bf3b7ed31dc04a9f20a15a4bb842285148f259b7d995a2ef13d82ff471de677R972-R984).

We’re keeping the new filters, but the nested properties will be re-assessed and implemented by https://github.com/Shopify/liquid/issues/1906.
  
I have [valuable concerns](https://github.com/Shopify/liquid/pull/1899#issuecomment-2598286898) raised by the community and [powerful alternatives](https://github.com/jg-rp/python-liquid2/issues/15) that deserve deeper consideration. 

Nested properties need to be supported, but they won’t be part of 5.7, as it’s more involved than this PR states.

Thanks to everyone following this PR.

rickydazla:
Love this update! Usage question: how would I use check if product.tags contains a certain value?

`product.tags | has: '???', 'MyTag` 

Using any of `title`, `name`, `tag`, or `tags` in the above doesn't seem to work...

andershagbard:
> Love this update! Usage question: how would I use check if product.tags contains a certain value?
> 
> `product.tags | has: '???', 'MyTag'`
> 
> Using any of `title`, `name`, `tag`, or `tags` in the above doesn't seem to work...

https://shopify.dev/docs/api/liquid/basics#contains

```liquid
if product.tags contains 'MyTag'
```

bakura10:
@karreiro , after testing this feature, I was a bit surprised that find_index and has does not allow to be used without properties. Would it be possible to support things like that:

```
{% assign has_product = block.settings.product_list | has: product %}
{% assign product_index = block.settings.product_list | find_index: product %}
```

andershagbard:
> @karreiro , after testing this feature, I was a bit surprised that find_index and has does not allow to be used without properties. Would it be possible to support things like that:
> 
> ```
> {% assign has_product = block.settings.product_list | has: product %}
> {% assign product_index = block.settings.product_list | find_index: product %}
> ```

I think that the `product_list` input is a bit weird. It doesn't seem like an actual array when accessing it's properties.

But, it would be nice if these filters could handle objects as well, like `contains` do.

```liquid
{%- liquid
assign variant = product.variants.first

# false
product.variants | has: variant

# blank
product.variants | find_index: variant

# true
if product.variants contains variant
```

PR #1870, opened 2024-12-13
don't add comment node to the nodelist

ggmichaelgo:
This is a micro optimization. We don't need to add the comment node to the nodelist, and we can shave off a few microseconds by skipping this step 😅


macournoyer:
I think this means comments won't be available in the AST anymore right? That might break tools using AST walking to inspect comments.

I only know of https://github.com/Shopify/theme-check, but it's deprecated. So might be good.

Do you think that might be an issue?

ggmichaelgo:
> I think this means comments won't be available in the AST anymore right? That might break tools using AST walking to inspect comments.
> 
> I only know of https://github.com/Shopify/theme-check, but it's deprecated. So might be good.
> 
> Do you think that might be an issue?

oh good point... it should be configurable to omit these nodes 👍

isaacbowen:
Hello from an AST-walking codebase! 👋 Thanks for thinking of us! 🥰

PR #1874, opened 2024-12-17
Bump version patch

bahar-p:
- Fast follow on https://github.com/Shopify/liquid/pull/1873 after digging further and getting more clarity on the why. Brought back the double string conversion, `to_s.to_str`. See the [comment](https://github.com/Shopify/liquid/pull/1874#discussion_r1888954627) as to why. 

- Bump the version. Next minor release will be 5.6.0, bumping the patch version for semver compliance. I also added the latest fix in the Changelog for the release candidate. I'm sure there has been more if you have context feel free to add the update there.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,12 @@
 # Liquid Change Log
 
+## 5.6.0 (unreleased)
+
+### Fixes
+
+* Fix Tokenizer to handle null source value (#1873) [Bahar Pourazar]
+
+
 ## 5.5.0 2024-03-21
 
 Please reference the GitHub release for more information.
--- a/lib/liquid/version.rb
+++ b/lib/liquid/version.rb
@@ -2,5 +2,5 @@
 # frozen_string_literal: true
 
 module Liquid
-  VERSION = "5.6.0.rc2"
+  VERSION = "5.6.0.rc3"
 end


lib/liquid/tokenizer.rb
```diff
@@ -5,7 +5,7 @@ class Tokenizer
     attr_reader :line_number, :for_liquid_tag
 
     def initialize(source, line_numbers = false, line_number: nil, for_liquid_tag: false)
-      @source         = source.to_s
+      @source         = source.to_s.to_str
```
bahar-p, line 8: For posterity: 

This is necessary to handle cases where `to_s` returns a subclass of String but we explicitly want to turn it into an instance of String. As stated in this PR, `ActiveSupport::SafeBuffer` is an subclass of string, calling `ActiveSupport::SafeBuffer#to_s` returns self (an instance of safe_buffer) but we actually want an instance of string. Calling `to_str` ensures we're always dealing with an instance of String

bahar-p: `to_str` is only defined on String: https://docs.ruby-lang.org/en/3.3/String.html#method-i-to_str

ianks (approved):

PR #1878, opened 2024-12-21
Breaking changes in v5.6.0

espen:
Added info about breaking change from https://github.com/Shopify/liquid/pull/1847


espen:
I have signed the CLA!

espen:
Should probably also mention changes from https://github.com/Shopify/liquid/pull/1848

PR #1879, opened 2024-12-24
Update README.md with default environment

benstein:
Smaller projects, especially when upgrading from earlier versions, may not need different environment scopes, so just adding to the docs.


benstein:
I have signed the CLA!

ianks (approved):

PR #1899, opened 2025-01-17
Use `VariableLookup` instead of dot notation in standard array filters

karreiro:
Following @ianks' suggestion (https://github.com/Shopify/liquid/pull/1869/files#r1919321969), this PR updates standard array filters to use `Liquid::VariableLookup` instead of the dot notation convention.

I initially considered using `Liquid::VariableLookup`, but recognized that it opens the door to broader use cases like accessing arrays, which might not be ideal as mentioned here (https://github.com/Shopify/liquid/pull/1869#discussion_r1907113691).

However, in the interest of a consistent implementation, we see value in evaluating this approach, which is the goal of this PR :)

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -1007,10 +1007,10 @@ module Liquid
 
       return value if !value.nil? || !property_or_keys.is_a?(String)
 
-      keys = property_or_keys.split('.')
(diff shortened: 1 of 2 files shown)


aswamy (approved):
Did some tophatting where i added `variants` array to the list of products in our sandbox, and it looks good 👍 

image

PR #1900, opened 2025-01-17
Fix `bundle exec rake example`

karreiro:
I frequently use the `bundle exec rake example` command to experiment with Liquid features, but we're missing `webrick` in the `Gemfile` to make it work as expected :)

Before:
image

After:
image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,11 +14,14 @@ group :benchmark, :test do
   gem 'memory_profiler'
   gem 'terminal-table'
   gem "lru_redux"
-  gem "webrick"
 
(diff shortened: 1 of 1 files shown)


EvilGenius13 (approved):

Gemfile
```diff
@@ -14,6 +14,7 @@ group :benchmark, :test do
   gem 'memory_profiler'
   gem 'terminal-table'
   gem "lru_redux"
+  gem "webrick"
```
ggmichaelgo, line 17: Do we need the `webrick` gem to be  in the `benchmark/test` group?

karreiro: Actually, I've moved `webrick` to the `development` group. I believe it fits better there, as neither `test` nor `benchmark` are so representative of its purpose.

PR #1902, opened 2025-01-23
Fix the `find` filter to return `nil` when filtering empty arrays

karreiro:
This PR is a patch that addresses an issue in the `find` filter introduced by https://github.com/Shopify/liquid/pull/1869.

When filtering an empty array, the `find` filter should return `nil`, just as it returns the found element when it locates the desired element or `nil` when the element cannot be found. However, it was incorrectly returning an empty array.


karreiro:
Thanks a lot for the review, @frandiox! I've applied the fixes for `find_index` and `has` as well :)

EvilGenius13 (approved):

PR #1905, opened 2025-01-30
Raise SyntaxError on invalid UTF8 strings in lexer/tokenizer

catlee:


History.md
```diff
@@ -2,6 +2,10 @@
 
 ## 5.8.0 (unreleased)
 
+## 5.7.2 2025-01-30
```
karreiro, line 5: As Liquid 5.7.2 is already released, I believe we may bump this to 5.7.3 if we need this fix released when we merge this PR. Also, I believe we might want to bump `version.rb` following [this guide](https://github.com/Shopify/liquid/blob/main/CONTRIBUTING.md#releasing).

Issue #1906, opened 2025-01-31
Introduce support for nested properties on Array filters


karreiro:
The https://github.com/Shopify/liquid/pull/1869 PR introduced support for new filters and nested properties together.

While that is a valuable feature, it might be risky for some runtimes, introduces [challenges for some tools and other runtime implementations](https://github.com/Shopify/liquid/pull/1899#issuecomment-2598286898), and could be replaced b [powerful alternatives](https://github.com/jg-rp/python-liquid2/issues/15).

This issue aims to assess support for that feature, without necessarily following the approach adopted by https://github.com/Shopify/liquid/pull/1869, and to consider a long-term path that preserves Liquid’s stability.

PR #1907, opened 2025-01-31
Fix array filters to not support nested properties

karreiro:
This PR removes support for nested properties.

We have conducted additional assessments regarding the backward compatibility aspect of this change and discovered it is not 100% backward compatible with all runtimes and temples. We are keeping the new filters, but nested properties will be re-assessed and implemented by https://github.com/Shopify/liquid/issues/1906.

We have [valuable concerns](https://github.com/Shopify/liquid/pull/1899#issuecomment-2598286898) raised by the community, along with [powerful alternatives](https://github.com/jg-rp/python-liquid2/issues/15) that warrant deeper consideration.

Nested properties must still be supported, but they will not be part of 5.7.x, as the required work is more significant than initially stated.

### DIFF
--- a/History.md
+++ b/History.md
@@ -2,6 +2,10 @@
 
 ## 5.8.0 (unreleased)
 
+## 5.7.2 2025-01-31
+
+* Fix array filters to not support nested properties
+
 ## 5.7.1 2025-01-24
 
 * Fix the `find` and `find_index`filters to return `nil` when filtering empty arrays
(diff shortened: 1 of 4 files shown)


ggmichaelgo (approved):

spookyvert:
hello sir @karreiro , 
https://github.com/Shopify/liquid/issues/1915

is this issue a side-effect of these changes?

PR #1921, opened 2025-02-27
Omit padding in base64_url_safe_encode filter

### LINKED ISSUE #1862: Base64 URL Safe Implementation
Base64 filters were added in Liquid 5.1 together with URL safe variants. There seems to be two variants out in the wild, and the question is can/should Liquid support both? base64_url_safe_decode supports both. base64_url_safe_encode does not.

1. Ruby's implementation swaps `+` and `/` only
2. Microsoft's implementations (Microsoft.IdentityModel.Tokens, System.Buffers.Text of .NET9) also remove the `=` padding. 

[RFC 4648](https://www.rfc-editor.org/rfc/rfc4648#section-5)
> The pad character "=" is typically percent-encoded when used in an URI, but if the data length is known implicitly, this can be avoided by skipping the padding

[Mozilla](https://developer.mozilla.org/en-US/docs/Glossary/Base64) defines it as 
> A common variant is "Base64 URL safe", which **_omits the padding_** and replaces +/

Related to [dotliquid#542](https://github.com/dotliquid/dotliquid/issues/542)

sebastienros:
Fixes #1862

`base64_url_safe_encode` should not render `=` characters. This change is compatible with already encoded strings, meaning `base64_url_safe_decode` will work on string encoded with the previous version that emitted padding characters.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -177,7 +177,7 @@ module Liquid
     # @liquid_syntax string | base64_url_safe_encode
     # @liquid_return [string]
     def base64_url_safe_encode(input)
-      Base64.urlsafe_encode64(Utils.to_s(input))
+      Base64.urlsafe_encode64(Utils.to_s(input), padding: false)
     end
 
     # @liquid_public_docs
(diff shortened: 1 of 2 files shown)


sebastienros:
It's a new year, maybe I can get a review, I believe it's an easy one.

Issue #1264, opened 2020-06-23
Ternary operator feature support


tmmgrafikr:
I know this feature has been requested before in this thread: https://github.com/Shopify/liquid/issues/236

**And based on 85 liking the idea, versus 2 not liking the idea says something.**

But that issue was closed back in 2016, and as the docs says:

```
Things we will merge
- Features that are likely to be useful to the majority of Liquid users
```
I personally still see a great advantage in having the opportunity to use ternary operators, the code is way more slick and friendly to look at.

I saw the ```default``` filter was suggested, but doesn't always come in handy, when using logic.

**Without ternary**
```
{% assign has_selected_variant = false %}
{% if product.selected_variant != nil %}
	{% assign has_selected_variant = true %}
{% endif %}
```

**With ternary**
```
{% assign has_selected_variant = product.selected_variant != nil ? true : false %}
```

Curzey:
This or #1271 would be so clean! 🚀

paulhuisman:
Please implement this. Saving heaps of lines of unnecessary code.

mirceapiturca:
This would be really handy and add readability to the code. I would love to see Liquid getting closer to JavaScript syntax (Ruby here also).

Curzey:
@dcalixto

Curzey:
@dylanahsmith 😶

dylanahsmith:
> I saw the `default` filter was suggested, but doesn't always come in handy, when using logic.

It seems. like the difference between what you want and the default filter is the ability to specify the value to use when it is truthy.

The `condition ? value_if_true : value_if_false` C-style syntax seems to those that aren't familiar with it.  My bias would be to something closer to python style conditional expression syntax (`value_if_true if condition else value_if_false`), which seems more readable and immediately understandable.  However, preceding with a condition would actually make it easier to turn into a filter by replacing `?` and `:` with the filter name and a keyword argument name.

Starting with a filter would make it much easier to introduce without worrying about backwards compatibility concerns.  It wouldn't preclude the introduction of special syntax for it.  It would actually make it easier to gauge adoption of the feature, that would help with justifying special syntax for making it more concise.

tmmgrafikr:
@dylanahsmith Would love to see an example of suggested filter option. Either way it would be an awesome feature, I think 
most liquid users would appreciate.

You thinking of something like this, or what is the idea of replacing ?: with filter names?
`assign ternary = somelogic | truthy: 'something' | falsy: 'somethingelse'`

[...]

PR #1922, opened 2025-03-04
Support for Boolean Expressions in Liquid

albchu:
## Overview
This PR introduces a syntax improvement to Liquid templates:

**Boolean Expression Support**: Enhanced boolean handling with proper operator precedence

This change makes templates more readable and concise by allowing developers to write expressions in a more natural way.

### Boolean Expression Enhancements
- Add support for logical operators (`and`, `or`)
- Implement proper operator precedence (e.g., `false and false or true` evaluates to `true`)
- Allow direct boolean expressions in assignment statements

#### Before:
```liquid
{%- liquid
  assign lazy_load = false
  if media_position > 1
    assign lazy_load = true
  endif
%}
```

#### After:
```liquid
{% assign lazy_load = media_position > 1 %}
```

## Benefits
- **More Readable Code**: Expressions are written in a more natural and familiar syntax
- **Less Verbose**: Fewer characters and more straightforward expressions
- **Easier Maintenance**: Simpler syntax reduces cognitive load for developers
- **Familiar Syntax**: Follows conventions from other programming languages

## Testing
Added a new test suite:
- `boolean_unit_test.rb`: Tests boolean operators, precedence, and assignment

All tests should pass with the implementation.

### DIFF
--- a/test/unit/boolean_unit_test.rb
+++ b/test/unit/boolean_unit_test.rb
@@ -398,14 +398,9 @@ class BooleanUnitTest < Minitest::Test
     HTML
 
     # This bugged output only happens in lax mode.
-    prev_error_mode = Liquid::Environment.default.error_mode
-    Liquid::Environment.default.error_mode = :lax
-    actual_lax_output = Liquid::Template.parse(template).render(context)
-    Liquid::Environment.default.error_mode = prev_error_mode
+    assert_with_lax_parsing(template, expected_lax_output, context)
 
-    actual_strict_output = Liquid::Template.parse(template).render(context)
-
-    assert_equal(expected_lax_output.delete("\n"), actual_lax_output.delete("\n"))
+    # Default test parsing mode (strict) works as properly expected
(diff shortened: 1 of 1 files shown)


t-kelly:
yessssssss

albchu:
> ooo, is a goal here (explicit or implicit) to deepen the functionality of `if` and `unless`? via parity for comparisons across the board?

Hey @isaacbowen, sort of! The current focus is to enable more concise boolean statements in output markup. So previously all falsey values would render as nothing. We'd like to support syntax where if `var_1 = true` and `var_2 = false`, we can evaluate `{{ var_1 and var_2 }}` to `"false"`.

madsenmm:
This looks like a simplified version of the [Ternary operator feature request](https://github.com/Shopify/liquid/issues/1264)
Would it maybe be possible to look into that, as it covers more usecases?

efekurnaz:
yes, yes, yes! and thumbsup for @madsenmm this could be the next step for this.

PR #1929, opened 2025-03-13
Stringify properties before filtering arrays

ianks:
Previously, filtering code allowed would blindly pass the `property` when calling `#[]` on the target input. While this usually works, it falls apart with non-string keys and can lead to some strange behaviors:

```ruby
template = Liquid::Template.parse("{{ input | where: property, 0 }}")
template.render({ "input" => [5, 10, 15, 20], "property" => 1 }) # => "520"
template.render({ "input" => [5, 10, 15, 20], "property" => 0 }) # => "1020"
```

Let's just always coerce properties to strings, first.

### DIFF
--- a/lib/liquid/standardfilters.rb
+++ b/lib/liquid/standardfilters.rb
@@ -979,8 +979,10 @@ module Liquid
     attr_reader :context
 
     def filter_array(input, property, target_value, default_value = [], &block)
-      ary = InputIterator.new(input, context)
+      property = Liquid::Utils.to_s(property)
+      return default_value if property.empty?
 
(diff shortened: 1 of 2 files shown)


karreiro (commented):
Thank you for this PR, @ianks!

We have a minor issue in this scenario:
```
{% assign fruits = 'banana apple tomato' | split: ' ' %}
{{ fruits | where: '' }}
```

With that fixed, I believe we're going to be ready to merge :)

ianks:
Verified this change against shadowed traffic, and this change did not impact any existing Liquid.

PR #1943, opened 2025-04-02
Revert `Utils.to_s` on all array filters

Maaarcocr:


isaacbowen:
Why?

isaacbowen:
Context on [my "Why?" question](https://github.com/Shopify/liquid/pull/1943#issuecomment-2775444690): I care about staying in sync with the unfolding strategy/philosophy y'all are working with on this project. If you're willing to share context for even patch-level stuff like this, I'm paying close attention, and it'd make my own Liquid-integrated product work more effective.

Maaarcocr:
> Context on [my "Why?" question](https://github.com/Shopify/liquid/pull/1943#issuecomment-2775444690): I care about staying in sync with the unfolding strategy/philosophy y'all are working with on this project. If you're willing to share context for even patch-level stuff like this, I'm paying close attention, and it'd make my own Liquid-integrated product work more effective.

we may revert this because it would cause some shops that use liquid for their theming to break today

PR #1944, opened 2025-04-03
Fix regression when using empty/nil properties with array filters

ianks:
Fixed a regression where empty properties like `{{ array | where: '' }}` would return an empty array. Now both empty strings and nil properties act as no-ops, simply returning the original array. This makes the library more forgiving when properties are missing or blank, reducing unexpected errors in your templates.


ianks:
Verified this against shadowed traffic and things look fine, this should be good to go

PR #1945, opened 2025-04-07
Ensure liquid values when comparing using the spaceship operator

dpetrick:
Not doing this can cause argument errors on types that should be comparable (like `IntegerDrop`s).


ianks:
Can you add a test for this?

PR #1954, opened 2025-05-01
Expose tag body in the Doc tag

jamesmengo:


lib/liquid/tags/doc.rb
```diff
@@ -31,21 +31,26 @@ module Liquid
   class Doc < Block
     NO_UNEXPECTED_ARGS = /\A\s*\z/
 
+    attr_reader :body
+
     def initialize(tag_name, markup, parse_context)
```
graygilmore, line 36: Should we set `@body` to an empty string in `initialize`? It feels weird to have an attr_reader on something that doesn't exist until `parse` gets run.

lib/liquid/tags/doc.rb
```diff
@@ -31,21 +31,26 @@ module Liquid
...
         tag_name = token =~ BlockBody::FullTokenPossiblyInvalid && Regexp.last_match(2)
 
         raise_nested_doc_error if tag_name == @tag_name
 
         if tag_name == block_delimiter
           parse_context.trim_whitespace = (token[-3] == WhitespaceControl)
+          @body << Regexp.last_match(1) if Regexp.last_match(1) != ""
           return
         end
+        @body << token unless token.empty?
```
graygilmore, line 53: For my own learning here: what are these two pieces of code actually shoveling into `@body`?

jamesmengo: Great question! I'm not too familiar with how the tokenizer itself decides what to chunk up, however when we parse Liquid it is parsed into an array of tokens

For example, `{% doc %} Documentation content\n  @param {string} foo - test\n {% enddoc %}` pass the following tokens to `parse`
1) `Documentation content
  @param {string} foo - test`
2) `{% enddoc %}`


To answer your question:
`@body << token unless token.empty?` writes tokens to body, and is doing the bulk of the work

`@body << Regexp.last_match(1) if Regexp.last_match(1) != ""` handles an edge case. When the token being processed looks like `{{ incomplete{% enddoc %}`, we strip the final content before the `{% enddoc %}`. I've added a test for this case as it seems to only be triggered by liquid characters

lib/liquid/tags/doc.rb
```diff
@@ -59,7 +63,7 @@ def blank?
     end
 
     def nodelist
-      []
+      [@body]
```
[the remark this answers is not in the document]
jamesmengo: I was on the fence about this since I vaguely remember a chat between @karreiro and I about whether `blank?` should just return `true` all of the time, however I think that with the move to nodelist you are right!

I will create an issue to follow up on this with him when he returns, but this should be easy to change to let's just fix forward for now

EvilGenius13 (approved):

PR #1975, opened 2025-07-30
Add hermetic template recording and replay

tobi:
## Summary

Adds hermetic recording and deterministic replay for successful Liquid renders.

```ruby
Liquid::TemplateRecorder.record("render.json") do
  Liquid::Template.parse(source).render!(assigns)
end

Liquid::TemplateRecorder.replay_from("render.json", mode: :verify).render
```

### What is captured

- root template and every parsed partial, with SHA-256 integrity checks
- partial file contents while preserving the existing one-argument FileSystem API
- ordinary Hash/Array assigns without wrapping or changing render semantics
- only properties actually read from Liquid Drops (Drop ivars are never scraped)
- exact filter calls/results, engine strictness, and final output

### Storage

- `.json`: an atomic session file containing every render in the block
- `.jsonl`: append-only, one compact and independently replayable render per line
  - file locking for concurrent writers
  - successful renders are durable independently
  - a truncated final line is recoverable
  - replay any record with `index:` (the last record is the default)

JSONL is deliberately the canonical production-sampling format rather than one large pretty JSON object. Compression is kept orthogonal: rotated JSONL can be compressed externally, and a future `.jsonl.zst` writer can use one independent zstd frame per record without changing the schema. A single long-lived compressed stream would make append/recovery/selection worse.

### Replay modes

- `compute`: run available filters normally
- `strict`: return exact recorded filter results and reject a changed call sequence
- `verify`: compute normally and reject output differences

## Design changes from the previous version

This is a ground-up simplification after reviewing the old implementation. It removes the global `Template.parse` monkey patch, proxy Hash/Array values, guessed Drop ivar extraction, lossy filter summaries, broken custom FileSystem keyword change, duplicated render wrappers, benchmark-specific CLI, and unrelated performance changes. Recording state is thread-local, nested sessions are rejected, JSON writes are atomic, and failures never delete a pre-existing destination.

## Test plan

- `bundle exec rake test`: all lax/strict/strict2 unit and integration runs pass
- fixed the stale `liquid-spec` branch pin (the upstream PR is merged) and two current-main strict2/RuboCop test regressions so CI can install and run
- focused recorder suite covers Drop fidelity/privacy, nested Drops, includes and legacy file systems, class and instance parsing, multiple renders/templates, JSONL append/recovery/indexing, exact strict filter replay, checksum validation, failed writes, supported render argument forms, nesting, and concurrent sessions
- RuboCop: no offenses on changed Ruby files

See `docs/template_recorder.md` for the API and format tradeoffs.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,6 @@ group :test do
 end
 
 group :spec do
-  # Using feature branch until https://github.com/Shopify/liquid-spec/pull/144 is merged
-  gem 'liquid-spec', github: 'Shopify/liquid-spec', branch: 'self-drop-env-lookup-specs'
+  gem 'liquid-spec', github: 'Shopify/liquid-spec'
   gem 'activesupport', require: false
 end
--- a/docs/template_recorder.md
+++ b/docs/template_recorder.md
@@ -1,0 +1,72 @@
+# Recording and replaying renders
+
+`Liquid::TemplateRecorder` captures successful template renders so they can be
+replayed without the application's file system or Drop implementations.
+Recording does not wrap or replace assigns, so the recorded render has the same
+semantics as a normal render.
+
+```ruby
+Liquid::TemplateRecorder.record("render.json") do
+  template = Liquid::Template.parse(source)
+  template.render!(assigns)
+end
+
+replayer = Liquid::TemplateRecorder.replay_from("render.json", mode: :verify)
+replayer.render # raises if the output changed
+```
+
+A recording contains the root template, every parsed partial, partial contents,
+plain Hash/Array values resolved by the template, properties actually read from `Liquid::Drop` objects,
+filter-call diagnostics, engine options, and the rendered output. Drop instance
+variables are never inspected. An unsupported Ruby object raises
+`Liquid::TemplateRecorder::SerializationError` rather than silently producing a
+recording that cannot be replayed.
+
+## Storage formats
+
+A `.json` destination is written atomically after the recording block succeeds.
+It contains a session with every render performed by the block.
+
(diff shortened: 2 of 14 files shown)


isaacbowen:
:))))))))))))))))

ianks:
This is really cool. I spent some time integrating this into our storefront to test it out, and here are some things that came up:

**1. Multi-template architectures**
Our storefront uses JSON templates that render multiple section templates in a single request. The current recorder design captures a single template source, but in practice we often have dozens of section templates rendering together. It might be helpful to support recording multiple templates in a single recording session, perhaps with a `templates` array that tracks each parsed template along with its name/path.

**2. Template parsing flow variations**
We noticed the recorder hooks into `Liquid::Template.parse` (class method), but in our codebase we create template instances first and then call `parse` on them:
```ruby
template = Liquid::Template.new
template.parse(source, options)
```
The instance method interception works, but the RecordingTemplate wrapper approach might need some adjustments for this pattern.

**3. Section rendering context**
For applications using section-based architectures, it would be valuable if the recording could capture which sections were rendered and in what order. This would make it easier to trace execution flow in complex page layouts.

**4. File path resolution**
The file system recording works well. One thing we noticed is that the paths in our file system reads are often relative to different roots, so having some metadata about the base path context would help with replay accuracy.

**5. Template source capture timing**
When templates are parsed through adapter layers or caching mechanisms, the source might not be available at the `Liquid::Template.parse` call site. Consider capturing source at multiple points in the template lifecycle to handle these cases.

The JSON schema is well structured and the recorder's approach to tracking variable access and filter calls is solid. Would be happy to discuss extending this to handle multi-template scenarios if you're interested.

PR #1991, opened 2025-09-30
Introduce `:rigid` parsing mode

karreiro:
This PR introduces the `:rigid` parsing mode because the existing `:strict` parser can still be too forgiving in certain situations:

```
<table>
  {% tablerow i in (1..10) limit: foo=>bar %}
    {{ i }}
  {% endtablerow %}
</table>
```


lib/liquid/environment.rb
```diff
@@ -34,7 +34,7 @@ class << self
       # @param file_system The default file system that is used
       #  to load templates from.
       # @param error_mode [Symbol] The default error mode for all templates
-      #  (either :strict, :warn, or :lax).
+      #  (either :strict, :warn, :lax, or :rigid).
```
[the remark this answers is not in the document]
karreiro: Good call! I've reordered modes there :)

History.md
```diff
@@ -1,5 +1,8 @@
 # Liquid Change Log
 
+## 5.8.8
```
charlespwd: Should at least be minor, no?

charlespwd (approved):
Thanks for cleaning all this up :D LGTM other than I think this is at least a minor version change :D

test/integration/tags/render_tag_test.rb
```diff
@@ -105,6 +105,32 @@ def test_dynamically_choosen_templates_are_not_allowed
...
+  end
+
+  def test_optional_commas
+    partials = { 'snippet' => 'hello {{ arg1 }} {{ arg2 }}' }
+    assert_template_result('hello value1 value2', '{% render "snippet", arg1: "value1", arg2: "value2" %}', partials: partials)
+    assert_template_result('hello value1 value2', '{% render "snippet"  arg1: "value1", arg2: "value2" %}', partials: partials)
+    assert_template_result('hello value1 value2', '{% render "snippet"  arg1: "value1"  arg2: "value2" %}', partials: partials)
+  end
+
   def test_include_tag_caches_second_read_of_same_partial
```
aswamy:
```suggestion
  def test_render_tag_caches_second_read_of_same_partial
```

test/test_helper.rb
```diff
@@ -82,10 +82,12 @@ def with_global_filter(*globals, &blk)
       Environment.dangerously_override(environment, &blk)
     end
 
-    def with_error_mode(mode)
+    def with_error_mode(*modes)
```
aswamy:
```suggestion
    def with_error_modes(*modes)
```

test/integration/variable_test.rb
```diff
@@ -209,4 +209,69 @@ def test_variable_lookup_should_not_hang_with_invalid_syntax
...
     end
   end
+
+  def test_filter_with_single_trailing_comma
+    template = '{{ "hello" | append: "world", }}'
+
+    with_error_mode(:strict) do
+      error = assert_raises(Liquid::SyntaxError) { Template.parse(template) }
+      assert_match(/is not a valid expression/, error.message)
+    end
```
aswamy: Situations like this is pretty much why i say `strict` should be "deprecated" rather than just say that it validates _some_ tags. It is just inaccurate here.

bin/render
EvilGenius13: Is this staying in the PR?

test/integration/tags/cycle_tag_test.rb
```diff
@@ -36,13 +26,157 @@ def test_cycle_with_variables_inside_for_loop
...
+
+  def test_unnamed_cycle_dependent_counter_when_used_with_literal_values
+    template = <<~LIQUID
+      {%- cycle "1", "2" -%}
+      {%- cycle "1", "2" -%}
+      {%- cycle "1", "2" -%}
+    LIQUID
+
+    assert_template_result("121", template)
+  end
```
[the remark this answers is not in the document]
charlespwd: I didn't delete the test, it's still there but has a different name (right above this one)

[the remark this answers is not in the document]
charlespwd: But yeah. I had a similar reaction. There are arguments for both being valid, but we're stuck with this because of the following:
-  if we change `{% cycle "1", "2" %}{% cycle "1", "2" %}` to behave differently, we'd break themes
- if we change the following, we'd break themes 
```liquid
{% for i in (1..5) %}
  <div class="container--{% cycle even_variant, odd_variant %}">
    <div class="element--{% cycle even_variant, odd_variant %}">
      ...
    </div>
  </div>
{% endfor %}
```

There are arguments for both to behave the way they do and I'd rather not try to rewrite this with a group because this is already rigid valid and we won't be able to tell the intention just by looking at the code.

bin/render
aswamy: I like it for debugging purposes :P

karreiro: Yes, that's intentional :) Just like we have the `example/server`, I'm keeping this as a lightweight development helper for future use cases.

Rakefile
```diff
@@ -33,7 +33,7 @@ task :rubocop do
   end
 end
 
-desc('runs test suite with both strict and lax parsers')
+desc('runs test suite with all parsers (lax, strict, and rigid)')
```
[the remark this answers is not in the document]
karreiro: Good call! Applied :)

lib/liquid/condition.rb
```diff
@@ -48,8 +48,8 @@ def self.operators
       @@operators
     end
 
-    def self.parse_expression(parse_context, markup)
-      @@method_literals[markup] || parse_context.parse_expression(markup)
+    def self.parse_expression(parse_context, markup, safe: false)
+      @@method_literals[markup] || parse_context.parse_expression(markup, safe: safe)
```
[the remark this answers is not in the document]
karreiro: I’ve wrestled with the same question myself.

The `gemspec` does require Ruby >= 3.0.0, so using `safe:` is definitely an option—and it’s my personal preference, too. However, for the sake of consistency with the rest of the existing codebase, I decided to stick with `safe: safe` in this and similar cases.

test/test_helper.rb
karreiro: Great call! I've applied this globally 🙇

README.md
```diff
@@ -99,14 +99,14 @@ Setting the error mode of Liquid lets you specify how strictly you want your tem
...
+Liquid also comes with a parser that can be used when editing templates to give better error messages
 when templates are invalid. You can enable this new parser like this:
 
 ```ruby
-Liquid::Environment.default.error_mode = :strict
-Liquid::Environment.default.error_mode = :strict # Raises a SyntaxError when invalid syntax is used
-Liquid::Environment.default.error_mode = :warn # Adds strict errors to template.errors but continues as normal
-Liquid::Environment.default.error_mode = :lax # The default mode, accepts almost anything.
+Liquid::Environment.default.error_mode = :rigid  # Raises a SyntaxError when invalid syntax is used in all tags
+Liquid::Environment.default.error_mode = :strict # Raises a SyntaxError when invalid syntax is used in some tags
```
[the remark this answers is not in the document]
aswamy: That's a fair reason, i don't want to blow up the scope. Let's proceed as is. Thanks for all the work in this PR ❤️

lib/liquid/tags/include.rb
```diff
@@ -101,6 +84,61 @@ def render_to_output_buffer(context, output)
...
+      p.consume(:end_of_string)
+    end
+
+    def strict_parse(markup)
+      lax_parse(markup)
+    end
+
+    def lax_parse(markup)
+      if markup =~ SYNTAX
+
```
ggmichaelgo: [nit] Shouldn't rubocop have caught this empty line 🤔 
```suggestion
```

lib/liquid/tags/include.rb
```diff
@@ -101,6 +84,61 @@ def render_to_output_buffer(context, output)
...
+      if with_or_for
+        @variable_name_expr = safe_parse_expression(p)
+      end
+
+      alias_name = nil
+      if p.id?("as")
+        alias_name = p.consume(:id)
+      end
+
+      @alias_name = alias_name
```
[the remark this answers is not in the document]
karreiro: Excellent call! I've applied this removing the local `alias_name` variable (as we only need the `@alias_name`)

lib/liquid/tags/cycle.rb
```diff
@@ -65,19 +54,82 @@ def render_to_output_buffer(context, output)
...
+        break if p.look(:end_of_string)
+
+        @variables << maybe_dup_lookup(safe_parse_expression(p))
+      end
+
+      p.consume(:end_of_string)
+
+      unless @is_named
+        @name = @variables.to_s
+        @is_named = !@name.match?(/\w+:0x\h{8}/)
```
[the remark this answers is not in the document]
karreiro: Excellent call here too! I've extracted the constant :)

lib/liquid/tags/include.rb
karreiro: Personally, I think rubocop shouldn't remove this line. But, I've removed it ;)

test/unit/condition_unit_test.rb
```diff
@@ -166,6 +166,37 @@ def test_default_context_is_deprecated
...
+    environment = Environment.build(error_mode: :strict)
+    parse_context = ParseContext.new(environment: environment)
+    result = Condition.parse_expression(parse_context, 'product.title')
+
+    assert_instance_of(VariableLookup, result)
+    assert_equal('product', result.name)
+    assert_equal(['title'], result.lookups)
+  end
+
+  def test_parse_expression_in_rigid_mode_raises_internal_error
```
graygilmore: Not 100% sold on the value of these tests since `Condition` doesn't actually implement this behavior itself. I'm not opposed to leaving it in but not sure if it's actually testing something.

test/integration/tags/table_row_test.rb
```diff
@@ -255,4 +258,212 @@ def test_table_row_does_not_leak_interrupts
...
+    with_error_modes(:lax, :strict) do
+      expected = <<~OUTPUT
+        <tr class="row1">
+        </tr>
+      OUTPUT
+      assert_template_result(expected, template)
+    end
+
+    with_error_modes(:rigid) do
+      # Rigid mode validates expression syntax and rejects invalid expressions
```
graygilmore: nit
```suggestion
```

graygilmore (approved):
Awesome work both of you!

PR #2001, opened 2025-10-20
Introduce Inline Snippets tag

dejmedus:
This PR implements [RFC#1916](https://github.com/Shopify/liquid/discussions/1916) and introduces support for the new inline snippets `{% snippet %}` tag, which allows us to define reusable template components directly within Liquid files

Traditionally, snippets are stored as separate `.liquid` files in a `snippets/` directory and rendered using the `{% render %}` tag. The `{% snippet %}` tag allows us to define and render inline snippets in the same file
  
``` liquid
  {% snippet input %}
    {% doc %}
      @param {string} type 
      @param {string} value
    {% enddoc %}

    <input type="{{ type }}" value="{{ value }}" />
  {% endsnippet %}

  {% render input, type: "text", value: "Hi!" %}
```

### DIFF
--- a/lib/liquid/tags/render.rb
+++ b/lib/liquid/tags/render.rb
@@ -58,7 +58,7 @@ module Liquid
         template_name = partial.name
         context_variable_name = @alias_name || template_name.split('/').last
       else
-        raise Liquid::ArgumentError, parse_context.locale.t("errors.file_system.includes")
+        raise ::ArgumentError
       end
 
       render_partial_func = ->(var, forloop) {
--- a/lib/liquid/locales/en.yml
+++ b/lib/liquid/locales/en.yml
@@ -34,5 +34,3 @@
       render: "Argument error in tag 'render' - Dynamically chosen templates are not allowed"
     disabled:
       tag: "usage is not allowed in this context"
-    file_system:
-      includes: "This liquid context does not allow includes"
--- a/test/integration/tags/snippet_test.rb
+++ b/test/integration/tags/snippet_test.rb
@@ -494,7 +494,7 @@ class SnippetTest < Minitest::Test
 
 
 
-        Liquid error (index line 2): This liquid context does not allow includes
+        Liquid error (index line 2): internal
       TEXT
       template.name = "index"
 
@@ -518,7 +518,7 @@ class SnippetTest < Minitest::Test
 
 
 
-        Liquid error (index line 2): This liquid context does not allow includes will raise an error.
+        Liquid error (index line 2): internal will raise an error.
 
         Bla bla test.
 
@@ -1017,7 +1017,7 @@ class SnippetTest < Minitest::Test
 
 
 
-        Liquid error (index line 2): This liquid context does not allow includes
+        Liquid error (index line 2): internal
       TEXT
       template.name = "index"
 
@@ -1041,7 +1041,7 @@ class SnippetTest < Minitest::Test
 
 
 
-        Liquid error (index line 2): This liquid context does not allow includes will raise an error.
+        Liquid error (index line 2): internal will raise an error.
 
         Bla bla test.
 


lib/liquid/tags/snippet.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Liquid
+  # @liquid_type tag
+  # @liquid_category theme
```
karreiro: I believe we could use `variable` to be consistent with the `capture` tag (also thinking it created a `SnippetDrop`)

```suggestion
  # @liquid_category variable
```

lib/liquid/tags/snippet.rb
```diff
@@ -0,0 +1,39 @@
...
+      if p.look(:id)
+        @to = p.consume(:id)
+      else
+        raise SyntaxError, options[:locale].t("errors.syntax.snippet")
+      end
+    end
+
+    def render_to_output_buffer(context, output)
+      snippet_drop = SnippetDrop.new(@body)
+      context.scopes.last[@to] = snippet_drop
```
karreiro: I believe we could have something like a [`with_snippet`](https://github.com/Shopify/liquid/blob/1c1e7119065b0fa915149b9e2bdc303490a0dfff/lib/liquid/tags/capture.rb#L35) method here. This way, we make sure the assignments for inline snippets will respect the [resource limits](https://github.com/Shopify/liquid/blob/1c1e7119065b0fa915149b9e2bdc303490a0dfff/lib/liquid/resource_limits.rb#L11) :)

lib/liquid/tags/render.rb
```diff
@@ -47,28 +47,48 @@ def render_to_output_buffer(context, output)
...
 
-      partial = PartialCache.load(
-        template_name,
-        context: context,
-        parse_context: parse_context,
-      )
+      if is_inline
+        template_name = template_name.name
+        snippet_drop = context[template_name]
+        raise ::ArgumentError unless snippet_drop.is_a?(Liquid::SnippetDrop)
```
karreiro: Great stuff with this error -- really appreciate the care 🚀

lib/liquid/tags/render.rb
```diff
@@ -138,8 +169,14 @@ def lax_parse(markup)
       @is_for_loop = (with_or_for == FOR)
 
       @attributes = {}
-      markup.scan(TagAttributes) do |key, value|
-        @attributes[key] = parse_expression(value)
+      markup.scan(/(#{ContextInheritance})|#{TagAttributes.source}/) do |context_marker, key, value|
```
karreiro: I believe we might be a bit more strict in this regex to prevent the scenario below:

```ruby
require 'liquid'

source = <<~EOS
  {% assign world = ', world!' %}

  {% snippet hello %}
    Hello{{ world }}
  {% endsnippet %}

  {% render hello, "..." %}
EOS

template = Liquid::Template.parse(source, error_mode: :lax)

puts template.render
# => Hello, world!
```

karreiro (commented):
Thank you for this PR, @dejmedus! Excellent stuff! I've left some minor comments, but it's really cool to see inline snippets working :) 🚀

lib/liquid/tags/render.rb
```diff
@@ -107,18 +127,29 @@ def rigid_parse(markup)
...
 
       @attributes = {}
-      while p.look(:id)
-        key = p.consume
-        p.consume(:colon)
-        @attributes[key] = safe_parse_expression(p)
+      while p.look(:dotdotdot) || p.look(:id)
+        if p.consume?(:dotdotdot)
+          @attributes.delete("...")
+          @attributes["..."] = true
```
charlespwd: I think we have to be thinking that we might expand this to `...product`. If we do, it would probably mean that we'd implement this differently. 

We need some kind of system for "spreading" props of something. Even if right now we're only spreading the props of the global context.

lib/liquid/tags/render.rb
```diff
@@ -107,18 +127,29 @@ def rigid_parse(markup)
...
     end
 
     def rigid_template_name(p)
-      p.consume(:string)
+      if p.look(:string)
+        p.consume(:string)
+      # inline snippets use variable identifiers
+      elsif p.look(:id)
+        p.consume(:id)
+      end
```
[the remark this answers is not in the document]
charlespwd: I think we might even want something like `consumeOneOf(:string, :id)` because the error that end_of_string throws is a bit misleading (expecting end_of_string but found "whatever")... actually we tried to expect string or id but neither passed kind of situation.

lib/liquid/tags/render.rb
dejmedus: I moved the regex directly into the `markup.scan` and updated it to make sure it can only be preceded by whitespace or a `,` :)

<details><summary>Fixed scenario</summary>
correct-spread-regex
</details>

lib/liquid/tags/render.rb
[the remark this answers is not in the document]
dejmedus: Just to double check: it sounds like we want to make sure that render is provided a string or an id and if not we'd like a descriptive error? 

> actually we tried to expect string or id but neither passed kind of situation

What about something like:

`raise SyntaxError, "Expected a string or identifier, found <whatever>"`

lib/liquid/tags/render.rb
karreiro: Personally, I think something like this would be good enough for now (then we may expand the error space in the future)

```ruby
...
@attributes = {}
while p.look(:dotdotdot) || p.look(:id)
  if p.consume?(:dotdotdot)
    raise SyntaxError if !p.look(:comma) && !p.look(:end_of_string) # TODO: error message
    @attributes.delete("...")
    @attributes["..."] = true
  else
    key = p.consume
    p.consume(:colon)
    @attributes.delete(key)
    @attributes[key] = safe_parse_expression(p)
  end
  p.consume?(:comma) # optional comma
end

p.consume(:end_of_string)
...
```

What do you think @dejmedus @charlespwd ?

charlespwd: Alright so I think all of my comments were unwarranted. Got confused because I saw a hash, but since we're doing `@attributes.each` and since hashes maintain insertion order, we're good with the code as is.

Am a bit worried about the performance of copying all the attributes to make the scoping work, but it's ok.

dejmedus:
>  instead of checking for SnippetDrop, if the ruby object that is passed responds to a certain method 

Thank you for the detailed explanation @Maaarcocr! I added a [`to_partial` method to SnippetDrop](https://github.com/Shopify/liquid/pull/2001/files#diff-be58ce77ce7b3ba3eb862058259cfe4d7a88e604f48cd9e7bcb7281c25c8f20cR52) so render can differentiate by `respond_to?(:to_partial)`. Does this sound like what we are aiming for?

lib/liquid/tags/snippet.rb
```diff
@@ -0,0 +1,45 @@
...
+  # @liquid_name snippet
+  # @liquid_summary
+  #   Creates a new inline snippet.
+  # @liquid_description
+  #   You can create inline snippets to make your Liquid code more modular.
+  # @liquid_syntax
+  #   {% snippet input %}
+  #     value
+  #   {% endsnippet %}
+
```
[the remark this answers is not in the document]
dejmedus: Good call! Fixed

karreiro (approved):
Thank you, @dejmedus! Excellent stuff 🚀

lib/liquid/tags/render.rb
```diff
@@ -55,6 +55,9 @@ def render_tag(context, output)
       elsif @template_name_expr.is_a?(String)
         partial = PartialCache.load(template, context: context, parse_context: parse_context)
         template_name = partial.name
+      elsif template.nil?
+        template_name = @template_name_expr.respond_to?(:name) ? @template_name_expr.name : @template_name_expr
+        raise FileSystemError, "No such template '#{template_name}'"
```
Maaarcocr: this should already be taken care of from the line below

PaulNewton:
@bakura10 it's reverted, and is in RFC https://github.com/Shopify/liquid/discussions/1916
I can use this like crazy too but it's current approach is flawed and needs more scrutiny and better feedback; and non-ambiguous naming or it's just gonna confuse merchants.

Probably be faster if the slack was still around so more shopify-partner even knew this was a thing they were finally chasing so more people were chiming in, or pushing on their shopify-connections.

> 
> [v5.11.0](https://github.com/Shopify/liquid/releases/tag/v5.11.0) [Latest](https://github.com/Shopify/liquid/releases/latest)
> What's Changed
> ...
> [Revert the Inline Snippets tag (https://github.com/Shopify/liquid/pull/2001), treat its inclusion in the latest Liquid release as a bug, and allow for feedback on RFC#1916 to better support Liquid developers by @karreiro in https://github.com/Shopify/liquid/pull/2012](https://github.com/Shopify/liquid/releases#:~:text=Revert%20the%20Inline%20Snippets%20tag%20(%232001)%2C%20treat%20its%20inclusion%20in%20the%20latest%20Liquid%20release%20as%20a%20bug%2C%20and%20allow%20for%20feedback%20on%20RFC%231916%20to%20better%20support%20Liquid%20developers%20by%20%40karreiro%20in%20%232012)

PR #2016, opened 2025-11-28
Preserve literal semantics in strict2 case/when

dejmedus:
Previously, `strict2` used `safe_parse_expression` to parse `when` expressions causing `blank` and `empty` to be treated as string literals (Expression::LITERALS maps 'empty' => ''), rather than method literals

This caused unexpected behaviour:

```liquid
{%- case products -%}
  {%- when empty -%}
    previously: doesn't render (products == '' is false)
    now: renders (products.empty? is true)
{%- endcase -%}
```

This PR instead calls `Condition.parse_expression` with `safe: true`, which will correctly handle `blank`/`empty`

<details><summary>Example</summary>

example
</details>

### DIFF
--- a/test/unit/tags/case_tag_unit_test.rb
+++ b/test/unit/tags/case_tag_unit_test.rb
@@ -82,6 +82,26 @@ class CaseTagUnitTest < Minitest::Test
     end
   end
 
+  def test_case_when_empty
+    template = <<~LIQUID
+      {%- case x -%}
+        {%- when 2 or empty -%}
+          2 or empty
+        {%- else -%}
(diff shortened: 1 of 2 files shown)


test/unit/tags/case_tag_unit_test.rb
```diff
@@ -82,6 +82,26 @@ def test_case_when_with_or
     end
   end
 
+  def test_case_when_empty
```
dejmedus: I'm only adding a test for `empty` as `blank` requires activesupport in order to render as we expect, however both have parity with `lax`

PR #2059, opened 2026-03-18
Add HybridTag base class and BlockBody reparenting

charlespwd:
## Summary

Adds foundational support for "hybrid tags" — tags that can function in both self-closing form (`{% section 'name' %}`) and block form (`{% section 'name' %}...{% endsection %}`). 

### Changes

- **`Liquid::HybridTag`** — New base class extending `Liquid::Block` that supports two rendering modes:
  - Self-closing: tag has no body, renders via `render_self_closing_to_output_buffer`
  - Block form: tag captures sibling nodes via end-tag-triggered reparenting, renders via `render_block_form_to_output_buffer`
  - Subclasses register which end tag triggers reparenting (e.g., `endsection`)

- **`BlockBody#parse` reparenting** — When the parser encounters an end tag for a registered hybrid tag (e.g., `{% endsection %}`), it looks back through `@nodelist`, finds the matching hybrid tag, splices intervening nodes into its body, and flips it to block form. No tokenizer changes needed.

- **Parse-time nesting prevention** — Hybrid tags of the same type cannot be nested (raises `SyntaxError`). Sequential hybrid tags and different types are allowed.

- **`@blank` recomputation** — After reparenting, the parent `BlockBody` recomputes its `@blank` flag based on remaining nodes.

### Design

Uses **end-tag-triggered reparenting** rather than tokenizer lookahead. When `BlockBody#parse` encounters `{% endsection %}`:
1. Walks `@nodelist` backwards to find the matching `section` tag
2. Splices all nodes between the hybrid tag and end tag
3. Calls `reparent_as_block(spliced_nodes)` on the hybrid tag
4. Recomputes `@blank` for the parent body

This approach requires zero tokenizer modifications.

### Why nesting is prohibited for hybrid tags

Hybrid tags support an optional end tag — the same tag can appear in self-closing form (`{% render 'a' %}`) or block form (`{% render 'a' %}...{% endrender %}`). This creates a fundamental **parser ambiguity** when nesting is allowed, because the parser cannot determine which opening tag an end tag belongs to.

Consider this template:

```liquid
{% render 'a' %}
{% render 'b' %}
{% endrender %}
{% render 'c' %}
{% endrender %}
```

Without indentation (which Liquid does not use for parsing), this is ambiguous. It could mean any of:

1. **`a` contains `b` and `c`** — `a` contains `b` (not self closing) and `c` (self closing)
2. **All three are sequential** — `a` is self-closing, `b` is a block, `c` is a block
3. **`a` contains `b`, then `c` is sequential** — `a` contains `b` (selfclosing),`c` is block

More concretely:

```liquid
{% render 'header' %}
{% render 'main' %}{% endrender %}
{% render 'footer' %}{% endrender %}
```

vs.

```liquid
{% render 'header' %}
  {% render 'nav' %}
{% endrender %}
{% render 'footer' %}{% endrender %}
```

These are visually different but structurally identical to the parser — whitespace/indentation is not significant in Liquid.

**The root cause**: when you can *omit* `end$tag`, the parser cannot tell whether an opening tag is self-closing or is the start of a block that hasn't been closed yet. With mandatory end tags (like `{% block %}...{% endblock %}`), there is no ambiguity — every opening tag must have a matching close.

This is why `Liquid::Block` subclasses (like the `block` tag) *can* nest — their end tags are mandatory. HybridTag's optional end tag makes nesting fundamentally ambiguous, so we prevent it at parse time.

## Test plan

- [x] Unit tests for HybridTag self-closing form (no end tag → stays self-closing)
- [x] Unit tests for HybridTag block form (end tag → reparents siblings)
- [x] `@blank` recomputation after reparenting
- [x] Nesting prevention (same-type hybrid tags raise SyntaxError)
- [x] Sequential hybrid tags (allowed)
- [x] Mixed self-closing + block in same template
- [x] Full existing Liquid test suite passes — zero regressions


lib/liquid/hybrid_tag.rb
```diff
@@ -0,0 +1,44 @@
...
+    def block_form?
+      !!@body
+    end
+
+    def nodelist
+      @body ? @body.nodelist : Const::EMPTY_ARRAY
+    end
+
+    def blank?
+      raise NotImplementedError, "#{self.class} must implement blank?"
```
mmorissette, line 23: In the description you mention

> `@blank` recomputation — After reparenting, the parent BlockBody recomputes its @blank flag based on remaining nodes.

Should this be done here instead of letting subclasses manage blank?

lib/liquid/hybrid_tag.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Liquid
+  class HybridTag < Block
+    def reparent_as_block(children, parse_context)
```
mmorissette, line 5: `parse_context`  is never used here

lib/liquid/document.rb
```diff
@@ -37,6 +37,10 @@ def unknown_tag(tag, _markup, _tokenizer)
       end
     end
 
+    def blank?
```
mmorissette, line 40: Why is this needed?

lib/liquid/block_body.rb
```diff
@@ -269,5 +281,46 @@ def raise_missing_tag_terminator(token, parse_context)
...
+          "errors.syntax.hybrid_tag_no_match",
+          end_tag: end_tag_name,
+          tag: hybrid_tag_name,
+        )
+      end
+
+      children = @nodelist.slice!((hybrid_index + 1)..)
+      hybrid_tag = @nodelist[hybrid_index]
+
+      hybrid_tag.reparent_as_block(children, parse_context)
```
mmorissette, line 321: Do we need to update @blank here after reparenting so that whitespace control works correctly for the parent body? 

```liquid
{%- if something -%}
  {% hybrid %}content{% endhybrid %}
{% endif %}
```

Without updating it, whitespace control might not work properly.

PR #2060, opened 2026-03-19
Reject bare-bracket syntax in strict2 and introduce `self` keyword

aswamy:
## Summary

Adds bare-bracket rejection to the strict2 parser and introduces a `self` keyword for explicit variable lookups.

**Parser changes:**
- `Parser#expression` now raises a `SyntaxError` when it encounters bare-bracket access (`['product']`) in strict2/rigid mode
- `ParseContext#new_parser` passes `reject_bare_brackets: true` when `error_mode` is `:strict2` or `:rigid`

**`self` keyword:**
- New `Expression::SELF` constant (`'self'`)
- New `SelfDrop` class — a drop that provides variable-only access to the context's scope chain (local > file > global), without exposing filters, interrupts, or other context internals
- `Context#find_variable` returns a `SelfDrop` when the key is `self` and `self` hasn't been explicitly assigned as a local variable
- If `self` is explicitly assigned (e.g. `{% assign self = 'value' %}`), the local value takes precedence
- `Context#variable_defined?` — new method that checks key existence across scopes/environments using `Hash#key?`, so nil-valued variables are correctly treated as defined (used by `SelfDrop#key?`)
- Annotated with YARD `@liquid_public_docs` tags for public documentation generation

**`Variable#==`:**
- Added equality method comparing `name` and `filters`, used by the rewriter to detect AST equivalence when deciding whether to flag nodes for rewrite

## Tophat

```bash
cd /path/to/liquid
bundle install
bin/render  # or: bundle exec irb -r liquid
```

**Bare-bracket rejection in strict2:**
```ruby
# Rejected
Liquid::Template.parse("{{ ['product'] }}", error_mode: :strict2)
# => Liquid::SyntaxError: Bare bracket access is not allowed in strict2 mode. Use self['...'] instead

# Accepted
Liquid::Template.parse("{{ product }}", error_mode: :strict2)
Liquid::Template.parse("{{ self['product'] }}", error_mode: :strict2)
Liquid::Template.parse("{{ product['title'] }}", error_mode: :strict2)
```

**`self` resolves through the scope chain:**
```ruby
# self['var'] walks local > file > global
t = Liquid::Template.parse("{{ self['product'] }}")
t.render('product' => 'shoes')
# => "shoes"

# Local assigns take precedence
t = Liquid::Template.parse("{% assign product = 'local' %}{{ self['product'] }}")
t.render('product' => 'global')
# => "local"

# Assigning self itself works
t = Liquid::Template.parse("{% assign self = 'hello' %}{{ self }}")
t.render
# => "hello"

# nil-valued variables are correctly handled under strict_variables
t = Liquid::Template.parse("{{ self['x'] }}")
t.render!({ 'x' => nil }, strict_variables: true)
# => "" (no UndefinedVariable error)
```

**Lax mode is unaffected:**
```ruby
Liquid::Template.parse("{{ ['product'] }}", error_mode: :lax)
# => OK, no error
```


lib/liquid/self_drop.rb
```diff
@@ -0,0 +1,38 @@
...
+    rescue UndefinedVariable
+      nil
+    end
+
+    def key?(key)
+      @context.variable_defined?(key)
+    end
+
+    def to_liquid
+      self
```
graygilmore, line 35: Image

lib/liquid/context.rb
graygilmore, line 1: Would love to see some additional tests for the new logic.

graygilmore (approved):
The drop looks good to me but I'd prefer to have somebody with a little more Liquid context give another approval.

PR #2063, opened 2026-03-20
Fix tokenizer to skip closing braces inside quoted strings

knu:
The byte-scanning tokenizer does not account for quoted strings when scanning for `}}` to terminate a variable token.  A `}` inside a quoted filter argument prematurely ends the token, causing a SyntaxError.

For example, the following valid template fails to parse:

    {{ message | replace: "{name}", customer_name }}

This patch makes `next_variable_token` skip over single- and double-quoted strings so that their contents do not interfere with `}}` detection.


knu:
I have signed the CLA!

PR #2064, opened 2026-03-23
Add for_loop? to Include tag for AST-based keyword detection

aswamy:
## Summary

Stores `@is_for_loop` during parsing in the `Include` tag, matching the existing pattern in the `Render` tag. This allows consumers (like the liquid-rewriter) to determine the `with`/`for` keyword from the parsed AST instead of re-parsing raw markup with regex.

**Changes:**
- Added `FOR = 'for'` constant
- Added `for_loop?` public method
- `strict2_parse`: captures the `with`/`for` token from `p.id?` and stores `@is_for_loop`
- `lax_parse`: updated `SYNTAX` regex to capture the keyword in its own group (group 3), stores `@is_for_loop`
- Regex capture groups shifted: variable name is now group 4, alias name is now group 6

No runtime behavior change — `Include#render_to_output_buffer` still uses `variable.is_a?(Array)` for iteration, not `@is_for_loop`.

Depends on: [Shopify/liquid-rewriter-private#214](https://github.com/Shopify/liquid-rewriter-private/pull/214)

## Tophat

```bash
bundle install
bundle exec irb -r liquid
```

```ruby
# for keyword detected
t = Liquid::Template.parse("{% include 'product' for products %}")
t.root.nodelist.first.for_loop?  # => true

# with keyword detected
t = Liquid::Template.parse("{% include 'product' with product %}")
t.root.nodelist.first.for_loop?  # => false

# no keyword
t = Liquid::Template.parse("{% include 'header' %}")
t.root.nodelist.first.for_loop?  # => false
```

### DIFF
--- a/lib/liquid/tags/include.rb
+++ b/lib/liquid/tags/include.rb
@@ -20,7 +20,8 @@ module Liquid
   class Include < Tag
     prepend Tag::Disableable
 
-    SYNTAX = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSegment}+))?/o
+    FOR = 'for'
+    SYNTAX = /(#{QuotedFragment}+)(\s+(with|#{FOR})\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSegment}+))?/o
     Syntax = SYNTAX
 
     attr_reader :template_name_expr, :variable_name_expr, :attributes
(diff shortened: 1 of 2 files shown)


test/integration/tags/include_tag_test.rb
```diff
@@ -439,4 +439,50 @@ def test_include_attribute_with_invalid_expression
       assert_match(/Unexpected character =/, error.message)
     end
   end
+  # -- for_loop? AST-based keyword detection --
```
charlespwd: Looks good but let's remove that comment. IDK why claude loves these so much. I struggle with em too.

charlespwd (approved):

lib/liquid/tags/include.rb
```diff
@@ -20,8 +20,9 @@ module Liquid
   class Include < Tag
     prepend Tag::Disableable
 
-    SYNTAX = /(#{QuotedFragment}+)(\s+(?:with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSegment}+))?/o
+    SYNTAX = /(#{QuotedFragment}+)(\s+(with|for)\s+(#{QuotedFragment}+))?(\s+(?:as)\s+(#{VariableSegment}+))?/o
```
graygilmore: nit: I kind of like how `render` uses `FOR` inside of `SYNTAX`: https://github.com/Shopify/liquid/blob/main/lib/liquid/tags/render.rb#L29-L30

graygilmore (approved):

EvilGenius13 (approved):

graygilmore (approved):

PR #2065, opened 2026-03-25
Add strict2_parse to assign and capture tags

aswamy:
## Summary

Adds `strict2_parse` to the `assign` and `capture` tags so invalid variable names are rejected in strict2 mode.

Both tags previously used only regex (`VariableSignature`) to validate variable names across all parse modes. This allowed invalid identifiers like `(a(b(c)`, `[x.y]`, and `a.b` to be silently accepted.

**assign:**
- `strict2_parse` uses the regex to split on `=`, then validates the LHS with `Parser#consume(:id)` + `consume(:end_of_string)`
- RHS is delegated to `Variable.new` as before

**capture:**
- `strict2_parse` uses `Parser#consume(:id)` + `consume(:end_of_string)` to validate the variable name

Both tags now include `ParserSwitching` and dispatch through `strict_parse_with_error_mode_fallback`. Lax mode is unchanged.

Depends on: [Shopify/liquid-rewriter-private#215](https://github.com/Shopify/liquid-rewriter-private/pull/215) (rewriter changes)

## Tophat

```bash
bundle install
bundle exec irb -r liquid
```

```ruby
# Invalid names rejected in strict2
Liquid::Template.parse("{% assign (a(b(c) = 1234 %}", error_mode: :strict2)
# => Liquid::SyntaxError

Liquid::Template.parse("{% capture [x.y %}hello{% endcapture %}", error_mode: :strict2)
# => Liquid::SyntaxError

# Valid names still work
Liquid::Template.parse("{% assign my-var = 'hello' %}", error_mode: :strict2)
# => OK

# Lax mode unchanged
Liquid::Template.parse("{% assign (a(b(c) = 1234 %}", error_mode: :lax)
# => OK
```

### DIFF
--- a/lib/liquid/tags/assign.rb
+++ b/lib/liquid/tags/assign.rb
@@ -18,6 +18,8 @@ module Liquid
   # @liquid_syntax_keyword variable_name The name of the variable being created.
   # @liquid_syntax_keyword value The value you want to assign to the variable.
   class Assign < Tag
+    include ParserSwitching
+
     Syntax = /(#{VariableSignature}+)\s*=\s*(.*)\s*/om
 
     # @api private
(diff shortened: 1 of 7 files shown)


graygilmore (approved):
This all seems reasonable!

lib/liquid/tags/assign.rb
```diff
@@ -18,6 +18,8 @@ module Liquid
   # @liquid_syntax_keyword variable_name The name of the variable being created.
   # @liquid_syntax_keyword value The value you want to assign to the variable.
   class Assign < Tag
+    include ParserSwitching
```
EvilGenius13, line 21: I don't think you need the include here because `Tag` already has it.

lib/liquid/tags/capture.rb
```diff
@@ -20,17 +20,35 @@ module Liquid
   # @liquid_syntax_keyword variable The name of the variable being created.
   # @liquid_syntax_keyword value The value you want to assign to the variable.
   class Capture < Block
+    include ParserSwitching
```
EvilGenius13, line 23: Same here

EvilGenius13 (approved):

PR #2067, opened 2026-03-26
Add strict2_parse to increment and decrement tags

aswamy:
> [!NOTE]
> Stacked on [Shopify/liquid#2065](https://github.com/Shopify/liquid/pull/2065) (assign/capture strict2_parse)

## Summary

Adds `strict2_parse` to `Increment` and `Decrement` tags. Both tags previously accepted any string as a variable name via `markup.strip` — no Parser validation was performed in any mode.

**Changes:**
- Both tags now `include ParserSwitching` and use `parse_with_selected_parser`
- `lax_parse`: preserves the original `markup.strip` behavior
- `strict_parse`: delegates to `lax_parse`
- `strict2_parse`: validates the variable name via `p.consume(:id)` + `p.consume(:end_of_string)`

This means `{% increment foo bar %}` and `{% decrement 11aa %}` are now rejected in strict2 mode.

## Tophat

```bash
bundle install
bundle exec irb -r liquid
```

```ruby
# Rejected in strict2
Liquid::Template.parse('{% increment foo bar %}', error_mode: :strict2)
# => Liquid::SyntaxError

Liquid::Template.parse('{% decrement 11aa %}', error_mode: :strict2)
# => Liquid::SyntaxError

# Accepted
Liquid::Template.parse('{% increment my-var %}', error_mode: :strict2)
# => OK

# Lax mode unchanged
Liquid::Template.parse('{% increment foo bar %}', error_mode: :lax)
# => OK (preserves backward compatibility)
```

### DIFF
--- a/lib/liquid/tags/increment.rb
+++ b/lib/liquid/tags/increment.rb
@@ -23,11 +23,27 @@ module Liquid
   #   {% increment variable_name %}
   # @liquid_syntax_keyword variable_name The name of the variable being incremented.
   class Increment < Tag
+    include ParserSwitching
+
     attr_reader :variable_name
 
     def initialize(tag_name, markup, options)
       super
+      parse_with_selected_parser(markup)
+    end
+
+    def lax_parse(markup)
       @variable_name = markup.strip
+    end
+
+    def strict_parse(markup)
+      lax_parse(markup)
+    end
+
+    def strict2_parse(markup)
+      p = @parse_context.new_parser(markup.strip)
+      @variable_name = p.consume(:id)
+      p.consume(:end_of_string)
     end
 
     def render_to_output_buffer(context, output)
--- a/lib/liquid/tags/decrement.rb
+++ b/lib/liquid/tags/decrement.rb
@@ -23,11 +23,27 @@ module Liquid
   #   {% decrement variable_name %}
   # @liquid_syntax_keyword variable_name The name of the variable being decremented.
   class Decrement < Tag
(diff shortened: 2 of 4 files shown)


lib/liquid/tags/increment.rb
```diff
@@ -23,13 +23,32 @@ module Liquid
...
+    end
+
+    def strict2_parse(markup)
+      stripped = markup.strip
+      return @variable_name = stripped if stripped.empty?
+
+      p = @parse_context.new_parser(stripped)
+      @variable_name = p.consume(:id)
+      p.consume(:end_of_string)
+    end
```
aswamy, line 47: i thought of merging the code between increment and decrement since they share so much, but didn't see it with other similar tags like `case` and `if` so i just kept it verbose

EvilGenius13: Seems fair to leave them separate. I do have a question.

The guard `return @variable_name = stripped if stripped.empty?` - I think we want to remove it because in other tags like capture if we don't find a variable name we raise a syntax error through `p.consume(:id)`

aswamy: Hmm yeah i think that's fair. I looked at how lax mode handles `{% increment %}` (without a variable name) and apparently it's allowed. Internally, it looks like it's creating an empty-string variable that can be accessed via `self[""]`. Ill rewrite this into a valid variable name. Good catch!

graygilmore (approved):

PR #2069, opened 2026-04-05
Minor cleanups to #2056

cpakman:
I stumbled on #2056 and noticed a few places readability could be improved, as well as one bug. I wanted to give a go at it for fun to see what I could come up with. I set a guardrail of keeping perf within a small margin of the original PR while trying to simplify the content to be easier to reason about and maintain. This rebases the original PR onto latest main and adds 5 additional commits to do some cleanup

*  Fixes an infinite loop in tokenize_fast where a template ending with a stray { would spin forever — the stray-brace scan loop could exit without advancing pos. Regression tests included.
* Adds a ByteTables module (four 256-entry boolean lookup arrays) to replace scattered inline byte-range comparisons throughout the codebase, and consolidates all byte constants behind Cursor::. 
* Breaks the try_fast_parse monolith in variable.rb into smaller methods. 
* Various dead code removal, idiom cleanup, and comments on the less obvious fast-parse logic.

* Performance vs autoresearch baseline: combined performance +3%, allocations +1.3%.


isaacbowen:
🥰 thank you

PR #2073, opened 2026-04-13
Add tests for reversed with limit/offset combinations

AnirudhDabas:
The for tag supports combining `reversed` with `limit` and `offset`, but
there are no tests exercising these combinations together. This adds three
cases that verify the two-phase parse behaviour -- `reversed` is matched
first, then limit/offset attributes are scanned separately.

Relates to #2056.

All 38 existing tests continue to pass.


AnirudhDabas:
I have signed the CLA!

PR #2080, opened 2026-04-30
fix(render): preserve scope chain when 'self:' bound on render tag

EvilGenius13:
## Summary

- `{% render 'snippet', self: obj %}` shadowed the `SelfDrop` because the render tag wrote `self:` directly into `@scopes[0]` under key `'self'`. Subsequent `self[var]` lookups inside the snippet did strict key access on `obj` instead of walking the scope chain, resolving to nil for any key not in `obj`.
- Fix: layered binding. `SelfDrop` now holds an optional `bound_self`; `[]` and `key?` consult it first and fall through to the existing scope-chain walk on miss. `Render#render_tag` special-cases attribute key `Expression::SELF` and routes the value to `inner_context.self_drop.bound_self =` instead of the generic scope write.
- Behavior preserved for existing `self:` users — when the key exists in the bound object, the bound value still wins.

## Why

PR #2060 introduced the `SelfDrop` and the `self[var]` syntax as the strict-mode replacement for bare-bracket lookups. Its tests exercise `self[var]` only at the top level via `template.render(hash)`. There was no coverage for the `{% render 'snippet', self: obj %}` interaction — and the render tag's existing generic attribute mechanism (which writes any attribute key into `@scopes[0]`) silently shadows the new SelfDrop when the attribute key happens to be `'self'`.

This PR adds that coverage and fixes the shadowing.

## Implementation

- `lib/liquid/self_drop.rb` — added `bound_self` accessor; `[]` and `key?` consult bound first via duck typing (matches the pattern in `lib/liquid/variable_lookup.rb:59-61`), fall through to `@context.find_variable(key)` on miss.
- `lib/liquid/context.rb` — added a memoized `self_drop` reader so the `SelfDrop` returned by `find_variable` is the same instance the render tag mutates.
- `lib/liquid/tags/render.rb` — attribute loop special-cases `Expression::SELF` (already used at `context.rb:215`) and routes to `inner_context.self_drop.bound_self =`.

Lookup order is bound-first, scope-chain-on-miss. Justified because it preserves all currently-passing `self:` users' hits and only changes previously-nil misses to "now resolves via scope-chain fallthrough."

Each `new_isolated_subcontext` creates its own `Context`, its own lazily-created `SelfDrop`, and its own `bound_self` — so nested renders, each passing their own `self:`, stay isolated.

## Test plan

- [x] New `test/integration/self_drop_render_test.rb` (8 tests, all passing):
  - Baseline: rewritten `self[name]` resolves correctly when no `self:` is bound (scope-chain walk works as before)
  - Bound-self shape: `self[name]` with `self:` bound resolves the snippet's local assigns when the key is missing from the bound object
  - Forloop-constructed key variant of the bound-self shape
  - Bare-bracket prohibition still raises the expected error in `:strict2`
  - Two-deep nested render with `self:` bound at both levels — each snippet sees only its own `bound_self`
  - Two-deep nested render with inner snippet unbound — leak-detection probes confirm outer `self:` does NOT leak in
  - Everything-chain composite: two nested renders + top-level `self[var]` + regular `{{ var }}` interpolations
  - Bound-key-hit pin: when the lookup key exists in the bound object, the bound value wins (precedence guarantee)
- [x] No regressions in \`test/integration/variable_test.rb\`, \`test/integration/tags/render_tag_test.rb\`, \`test/integration/drop_test.rb\`, \`test/integration/context_test.rb\`
- [ ] Mirror change in \`Shopify/liquid-vm\` — follow-up PR (test scaffolding already exists)

## Notes

- One pre-existing test failure in \`test/integration/tags/render_tag_test.rb\` (\`uninitialized constant RenderTagTest::TestEnumerable\`) and one env-dependent flake in \`test/integration/context_test.rb\` (\`stackprof\` LoadError). Both verified pre-existing on \`main\`; not introduced by this PR.
- \`bundle exec rake integration_test\` is broken in the local Nix Ruby env (native extension build failures); tests were run individually via \`ruby -Itest -Ilib\`.

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -200,6 +200,10 @@ module Liquid
       object.respond_to?(:evaluate) ? object.evaluate(self) : object
     end
 
+    def self_drop
+      @self_drop ||= SelfDrop.new(self)
+    end
+
     # Fetches an object starting at the local scope and then moving up the hierachy
     def find_variable(key, raise_on_not_found: true)
       # This was changed from find() to find_index() because this is a very hot
@@ -208,7 +212,7 @@ module Liquid
 
       # `self` resolves to a SelfDrop (enabling `self['var']` lookups),
       # but only when it hasn't been explicitly assigned as a local variable.
-      return SelfDrop.new(self) if key == Expression::SELF && !index
+      return self_drop if key == Expression::SELF && !index
 
       variable = if index
         lookup_and_evaluate(@scopes[index], key, raise_on_not_found: raise_on_not_found)
--- a/lib/liquid/self_drop.rb
+++ b/lib/liquid/self_drop.rb
@@ -16,23 +16,42 @@ module Liquid
   #   then the local value takes precedence over the `self` object.
   # @liquid_access global
   class SelfDrop < Drop
+    attr_accessor :bound_self
+
     def initialize(context)
       super()
       @context = context
+      @bound_self = nil
     end
 
     def [](key)
-      @context.find_variable(key)
+      if @bound_self && bound_has?(key)
+        bound_lookup(key)
+      else
+        @context.find_variable(key)
+      end
     rescue UndefinedVariable
       nil
     end
 
     def key?(key)
-      @context.variable_defined?(key)
+      (@bound_self && bound_has?(key)) || @context.variable_defined?(key)
     end
 
     def to_liquid
       self
+    end
+
+    private
+
+    def bound_has?(key)
+      @bound_self.respond_to?(:key?) && @bound_self.key?(key)
+    end
+
+    def bound_lookup(key)
+      return unless @bound_self.respond_to?(:[])
+
+      @bound_self[key]
     end
   end
 end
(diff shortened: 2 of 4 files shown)


lib/liquid/context.rb
aswamy, line 1: Could we move the `self` stateful logic to the context rather than the drop? I think we should keep the self drop code itself as light as possible. I know this makes the context bigger, but i dont think we can avoid that here.

cc @karreiro 

```
    attr_accessor ....., :bound_self

    def self_lookup(key)
      if @bound_self && bound_self_has?(key)
        @bound_self[key]
      else
        find_variable(key)
      end
    rescue UndefinedVariable
      nil
    end

    def self_defined?(key)
      (@bound_self && bound_self_has?(key)) || variable_defined?(key)
    end
```

karreiro (commented):
Thank you for this PR, @EvilGenius13!

The motivation here is in the right place, as we're making `self` behavior more resilient while preserving values set by users.

However, given the mental model we adopted in https://github.com/Shopify/liquid/pull/2082 of handling self as a regular drop, I don't think we want to proceed with this one.

Thanks again for the PR!

PR #2081, opened 2026-05-06
Move `self` handling to variable lookup

jg-rp:
This PR shows an alternative implementation of `self` by removing `SelfDrop` and moving handling of `self` from `Context` to `VariableLookup`.

This approach changes `self` semantics to fix the issue identified in #2080. Instead of conditionally replacing `self` with a drop, we now fall back to top-level variable resolution if `self[key]` does not exist.

A comforting side effect is that `self` can no longer be aliased with `{% assign this = self %}`, or be passed as an argument `{% render 'snippet', this: self %}` (this does not cause too much of a problem because `SelfDrop.context` is always replaced with the current context).

(Edit: My assumptions about the intended scoping rules for `self` might not be quite right.)


navayla777-dotcom (commented):
Imitation doctor platform development assignment perhaps?

PR #2082, opened 2026-05-06
Prevent `SelfDrop` context mutation across render boundaries

karreiro:
This PR fixes `SelfDrop` scoping so a `self` assigned in one template preserves its original scope when passed to a rendered template.

```liquid
# template1.liquid
{% assign var = 42 %}
{% assign s = self %}
{% render 'template2', other_self: s %}

# template2.liquid
{% assign var = 43 %}
{{ other_self.var }}  => 43 (wrong — should be 42)
```

The fix renames `@context` to `@self_context` in `SelfDrop`, so the drop has a specific reference to its originating context instead of reusing the inherited `context` that gets overwritten at the `context.rb` level on line 220.

### DIFF
--- a/lib/liquid/self_drop.rb
+++ b/lib/liquid/self_drop.rb
@@ -16,19 +16,19 @@ module Liquid
   #   then the local value takes precedence over the `self` object.
   # @liquid_access global
   class SelfDrop < Drop
-    def initialize(context)
+    def initialize(self_context)
       super()
-      @context = context
+      @self_context = self_context
     end
 
     def [](key)
-      @context.find_variable(key)
+      @self_context.find_variable(key)
     rescue UndefinedVariable
       nil
     end
 
     def key?(key)
-      @context.variable_defined?(key)
+      @self_context.variable_defined?(key)
     end
 
     def to_liquid
--- a/lib/liquid/self_drop.rb
+++ b/lib/liquid/self_drop.rb
@@ -31,10 +31,10 @@ module Liquid
       @self_context.variable_defined?(key)
     end
 
-    undef context
-
     def to_liquid
       self
     end
+
+    undef context=
   end
 end
--- a/lib/liquid/self_drop.rb
+++ b/lib/liquid/self_drop.rb
@@ -31,7 +31,7 @@ module Liquid
       @self_context.variable_defined?(key)
     end
 
-    def context=(_); end
+    undef context
 
     def to_liquid
       self
--- a/lib/liquid/self_drop.rb
+++ b/lib/liquid/self_drop.rb
@@ -31,6 +31,8 @@ module Liquid
       @self_context.variable_defined?(key)
     end
 
+    def context=(_); end
+
     def to_liquid
       self
     end
--- a/test/integration/self_drop_context_test.rb
+++ b/test/integration/self_drop_context_test.rb
@@ -68,4 +68,24 @@ class SelfDropContextTest < Minitest::Test
 
     assert_template_result('42', source)
   end
+
+  def test_self_drop_context_writer_is_a_noop
+    context = Context.new
+    drop = SelfDrop.new(context)
+    assert(drop.respond_to?(:context=))
+    drop.context = Context.new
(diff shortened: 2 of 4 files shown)


test/integration/self_drop_context_test.rb
aswamy, line 1: Rename to `test/integration/self_drop_context_test.rb`

aswamy (approved):
Not sure if liquid-spec also allows render tags, but ill update it to capture your test. Thank you for this ❤️ 

0https://github.com/Shopify/liquid-spec/pull/124

test/integration/self_drop_context_test.rb
karreiro: Good call! Fixed :)

ianks (changes requested):
We likely want to `undef context` as well, which serves as a signal that the drop doesn’t support contextualization

karreiro:
Thank you for the review, @ianks!

I did consider `undef_method :context=`, but that felt [a bit unexpected](https://en.wikipedia.org/wiki/Liskov_substitution_principle), also because some places (e.g., `template.rb:155`) call `Drop#context=` without any guard. I went with `def context=(_); end` instead to keep the interface intact. Please let me know your thoughts!

jg-rp:
Hi 👋 

This looks like a subtle change in policy for the `{% render %}` tag. A deliberate mechanism for template authors to bring all of the parent template's scope into a partial template, albeit namespaced under an arbitrary variable.

I understand if you've decided this is a positive new feature. It's just that with this change partial templates are less isolated. They are potentially more tightly coupled to their parent templates, which I thought was one of the reasons for introducing `{% render %}` and depreciating `{% include %}`.

And, of course, having arbitrary aliases of `self` does make static analysis much harder.

karreiro:
Thank you for the review, @jg-rp! That's a fair concern indeed.

I had the same instinct when I first saw this, but talking with @ianks clarified this in my mind:
- `{% include %}` did was leak the parent's entire scope implicitly, making static analysis painful and coupling invisible
- `self` works differently, as it's a readonly drop being explicitly passed, so there's no expensive implicit scope shared, just cheap lookups

The positive feature we're introducing with `self` is being able to rewrite expressions like `{{ ['a'] }}` to `{{ self['a'] }}`.

The isolation boundary is still there. It just has a handle to a drop that was deliberately handed to it.

I hope this clarifies the concern and motivation. Thanks again for the review!

lib/liquid/self_drop.rb
```diff
@@ -16,21 +16,23 @@ module Liquid
...
     rescue UndefinedVariable
       nil
     end
 
     def key?(key)
-      @context.variable_defined?(key)
+      @self_context.variable_defined?(key)
     end
 
+    def context=(_); end
```
ianks:
```suggestion
    undef context
```

this should be done so upstream implementations know the value is not contextualizable

ianks (changes requested):

karreiro:
Thank you for the review, @ianks!

I think you meant `undef :context=` -- `Drop` only defines `attr_writer :context` (the setter), there's no `context` reader method, so `undef context` would raise a `NameError`.

That said, I trust your instinct here. But I want to flag the complexity this introduces. Because `context=` is defined on `Drop` but undefined on `SelfDrop`, every call site for `Liquid::Drop` now needs a guard:

```ruby
case arg
when Liquid::Drop
  drop.context = c if drop.respond_to?(:context=)
...
```

Most call sites on Liquid already had this, but `template.rb:155` didn't -- that's fixed in this PR too. It's a subtle new contract that other Liquid runtimes will need to implement.

If the goal is signaling that a drop is not contextualizable, a method that communicates intent directly might be cleaner:

```ruby
# drop.rb
def contextualizable?
  true
end
```

Then `SelfDrop` overrides it to return `false`. No `respond_to?` guards, explicit intent, and upstream implementations get a clear signal about what's contextualizable and what isn't.

Happy to keep the `undef` approach if there's a reason I'm not seeing, just wanted to surface the trade-off.

ianks (approved):

PR #2091, opened 2026-06-02
Fix `SelfDrop` equality

karreiro:
`Liquid::SelfDrop` currently behaves like a wrapper object for equality, even though every meaningful operation delegates back to the captured context. That means two separate `self` lookups from the same context compare as different values, leaking implementation detail into both Ruby callers and template behavior:

```ruby
ctx = Liquid::Context.new({})
a = ctx.find_variable("self")
b = ctx.find_variable("self")

[a == b, a.equal?(b), a == a]
# Before: [false, false, true]
# After:  [true,  false, true]
```

This change makes `SelfDrop` equality compare the captured context identity, and aligns `eql?`/`hash` with the same rule so Ruby hash-key semantics stay coherent.

### DIFF
--- a/lib/liquid/context.rb
+++ b/lib/liquid/context.rb
@@ -208,7 +208,7 @@ module Liquid
 
       # `self` resolves to a SelfDrop (enabling `self['var']` lookups),
       # but only when it hasn't been explicitly assigned as a local variable.
-      return SelfDrop.new(self) if key == Expression::SELF && !index
+      return @self_drop ||= SelfDrop.new(self) if key == Expression::SELF && !index
 
       variable = if index
         lookup_and_evaluate(@scopes[index], key, raise_on_not_found: raise_on_not_found)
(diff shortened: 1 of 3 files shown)


aswamy (approved):
Looks good 👍 I also checked out the test on liquid-spec and seem consistent.

EvilGenius13 (approved):

Maaarcocr (approved):

jg-rp:
Have you considered lazily creating, storing and reusing at most one `SelfDrop` instance per context, instead of creating a new `SelfDrop` for each call to `find_variable("self")`? Then you could rely on identity equality.

karreiro:
Thank you, @jg-rp!

I've applied the suggestion to reuse the same instance instead of creating a new `SelfDrop` per context, that's a great idea.

It would also enable us to revert the `==` family of changes, but I'm leaning toward keeping both since `lib/liquid/self_drop.rb` should be a shallow wrapper around the context, enabling `self` evaluation. Thanks again for the suggestion.

PR #2094, opened 2026-06-11
Let environment `self` shadow `SelfDrop`

ianks:
Previously, bare `self` returned `SelfDrop` as soon as local scopes missed, which meant an environment or counter variable named `self` couldn't shadow it. This PR fixes that by checking environments before falling back to `SelfDrop`, while still suppressing strict-variable errors for the fallback path.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,7 @@ group :test do
 end
 
 group :spec do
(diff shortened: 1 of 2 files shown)


karreiro (approved):
Thank you, @ianks!

PR #2104, opened 2026-07-05
Fix custom tag render result stringification

### LINKED ISSUE #1893: Missing `.to_s` when rendering a node (dropped possibily by accident)
Hey y'all,

I'm upgrading from Liquid v4 to v5 and encountered a subtle yet **impactful breaking change** that caused us a significant number of test failures. It seems that during a refactor, the call to `.to_s` for a node output was unintentionally (?) dropped.

If we travel back in time and look at [black_body.rb from commit c2ef24](https://github.com/Shopify/liquid/blob/c2ef247be5b742945e6cf1a0023ce84770190751/lib/liquid/block_body.rb#L104), we can appreciate how it'd still explicitly call `.to_s` on the output of `node#render` unless the result was an array.
Image

However, [while refactoring to improve rendering](https://github.com/Shopify/liquid/pull/1091/files?w=1#diff-ec6f3ab191efb7fdc68101dd9df369f98395016152698407d1a695d10fcb08a5R42-R45) back in #1091 the new `render_to_output_buffer` dropped this, maybe by accident (?). Instead, it started "appending" the result of `render` directly to the output buffer without stringifying it.

Image

This causes this funny case where if your output returns a number, the `<<` operation ends up adding a byte instead of the stringified value.
i.e.:

Image

sjh9714:
## Summary

Fixes #1893.

Custom tags that rely on the legacy `render(context)` compatibility path were appending non-string render results directly to the output buffer. That made an integer render result use `String#<<` byte semantics, so `49` rendered as `1`. This restores the pre-refactor behavior by joining array render results and stringifying other truthy render results before appending.

## Changes

- Convert legacy custom tag render results before appending to the output buffer.
- Add regression coverage for integer and array render results.

## Testing

- Red check on a temporary `main` worktree: `docker run --rm -v /tmp/liquid-red-1893:/repo -w /repo ruby:4.0.2 ruby -Itest -Ilib test/integration/tag_test.rb -n /stringify/` failed with `Expected: "49"`, `Actual: "1"`.
- `docker run --rm -v "$PWD":/repo -w /repo ruby:4.0.2 ruby -Itest -Ilib test/integration/tag_test.rb -n /stringify/`
- `docker run --rm -v "$PWD":/repo -w /repo ruby:4.0.2 ruby -Itest -Ilib test/integration/tag_test.rb`
- `docker run --rm -v "$PWD":/repo -w /repo ruby:4.0.2 ruby -Itest -Ilib test/integration/block_test.rb`
- `docker run --rm -v "$PWD":/repo -w /repo ruby:4.0.2 ruby -c lib/liquid/tag.rb`
- `docker run --rm -v "$PWD":/repo -w /repo ruby:4.0.2 ruby -c test/integration/tag_test.rb`
- `git diff --check`

Verification gap: I also tried `ruby -Itest -Ilib -S rake base_test` in the Ruby 4.0.2 container, but the plain container lacked `lru_redux`. A disposable Bundler install was blocked because the `Shopify/liquid-spec` branch `self-drop-env-lookup-specs` could not be resolved.

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.


sjh9714:
I have signed the CLA!

